Issue #9, opened 2019-09-13
Package still sometimes hangs for a few seconds after completion


tmeasday:
I am using `chromaui/chromatic-cli#next`.

https://www.loom.com/share/672614f6b3ca4f888da35514a41ed2b7

PR #19, opened 2019-09-25
Exit hard after success, do not wait for bulky logs

ndelangen:
fixes: #9 

Do a hard exit even if process build was a success

### DIFF
--- a/bin/main.js
+++ b/bin/main.js
@@ -113,7 +113,10 @@ export async function run(argv) {
       log.error('Problems encountered:');
       errors.forEach(e => log.error(e));
     }
+
+    // Not sure what exit code to use but this can mean error.
+    process.exitCode = process.exitCode || 255;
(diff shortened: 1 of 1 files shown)


tmeasday:
@ndelangen as you can see from the CI failures, this isn't quite right. In the succeeding case, the `exitCode` will be 0, which is `||`-ed to `255`.

Looking at the current (published) code:

https://github.com/chromaui/chromatic/blob/v2.2.2/packages/storybook-chromatic/src/bin/chromatic-test.js#L247-L277

I think the correct logic would be.

1. In the `try/catch` set `process.exitCode = 255`
2. At the end, call `process.exit(process.exitCode)`.

ndelangen:
Thanks, I found a good solution I think.

tmeasday:
Ok that looks good.

PR #36, opened 2019-10-09
FIX 34 - respect the CHROMA_* env variables

ndelangen:
I reworked the constants file so it will will use either the CHROMA_* or CHROMATIC_* env variables (and still has a default for some)


ndelangen:
So you want ONLY the ?_APP_CODE to be posisble to be prefixed with CHROMA?

PR #39, opened 2019-10-09
Add LICENSE file

NMinhNguyen:
I wasn't too sure what exactly to put on line 3:

> Copyright (c) Chroma Software Inc.

but I based it on Facebook's MIT license (https://github.com/facebook/react/blob/4be45be5ff4b26892bab83b5f9022afac1a664fb/LICENSE#L3) and `Chroma Software Inc.` in the footer of https://hichroma.com/

Reason I'm proposing this change is there's a licence checker at work and I believe it expects there to be a `LICENSE` file, presumably, in addition to the `license` field in `package.json` which you already have:

https://github.com/chromaui/chromatic-cli/blob/71c4e439e2b9c1c2d70b7eac46487e224eb765be/package.json#L18


ndelangen:
👏 thank you!

PR #43, opened 2019-10-09
Fix/GitHub actions

ndelangen:
Users are trying to run chromatic-cli in GitHub actions.

Depending on the event type, sometimes this works as expected, sometimes it doesn't.

I investigated when it wrk and when it doesn't and why. It turns out the GITHUB_REF does not point to a branch when the event is of type `pull_request`, and the GITHUB_SHA refers to a commit of the last commit merged into the base branch. (This explains why chromatic was reporting the branch to be the base branch).

I created a official github action for chromatic here:
https://github.com/chromaui/action
(this is still in progress)

I created a couple of fixes to make this package work in github actions as long as the event is `push`.

Also fixes: #45


ndelangen:
Screenshot 2019-10-09 at 15 46 25

bin/git/getCommitAndBranch.js
```diff
@@ -17,8 +17,14 @@ export async function getCommitAndBranch({ inputFromCI } = {}) {
     TRAVIS_REPO_SLUG,
     TRAVIS_PULL_REQUEST_SHA,
     TRAVIS_PULL_REQUEST_BRANCH,
+    GITHUB_WORKFLOW,
```
tmeasday: Is this an environment variable we ask users to set?

ndelangen: No, github sets it.

tmeasday: Always?

tmeasday:
On branch names:

It’s pretty important because the Pr feature relies on it. Are there other env vars we can look at? What is the checked out git branch in each case?

On commit hash:

It is crucial we get this right or baselines get lost and chromatic is pretty much not useful. Is it available in a different env var? Should we call back to looking at the git history to work it out?

ndelangen:
@tmeasday I'm working on https://github.com/chromaui/action to resolve this.

bin/git/getCommitAndBranch.js
ndelangen: in a github action, yes

ndelangen:
I'm close:
Screenshot 2019-10-15 at 11 54 06

There are 2 deviations, 1 is the pull-request manual action.
Github gives a GITHUB_REF of a PR, not a branch, and the commit is of a future merge commit, not the last commit by a user.

The other is one of the 'via actions', I'm unsure which, investigating...

ndelangen:
The pull-request event seems to be handled correct in the action:
Screenshot 2019-10-15 at 12 12 43

PR #50, opened 2019-10-24
ADD storybookOptions to the input for chromatic

ndelangen:
@tmeasday this is step 1 for https://github.com/chromaui/chromatic/issues/1956

Shall we schedule a moment we can tackle step 2 together maybe?

### DIFF
--- a/bin/tester/runtimes.js
+++ b/bin/tester/runtimes.js
@@ -1,6 +1,7 @@
 /* eslint-disable no-useless-catch, no-console, no-underscore-dangle */
 import { JSDOM, VirtualConsole, ResourceLoader } from 'jsdom';
 import dedent from 'ts-dedent';
+import { stringify } from 'telejson';
 
 import { extract } from '../storybook/extract';
 import { toSpec } from './toSpec';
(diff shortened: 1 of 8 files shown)


bin/tester/runtimes.js
```diff
@@ -97,17 +99,26 @@ export default async function getRuntimeSpecs(url, { verbose = false } = {}) {
...
   } finally {
     // cleanup
     dom.window.close();
   }
 }
+
+const getOptions = ({ options }) => ({
+  hierarchyRootSeparator: options.hierarchyRootSeparator,
+  hierarchySeparator: options.hierarchySeparator,
+  storySort: options.storySort ? options.storySort.toString() : undefined,
```
tmeasday: Do you think we should pass this up to the server, or should we instead just sort the stories before passing them up?

tmeasday: (Not psyched about `eval`-ing user input in our server node process ;)

ndelangen: Yea, I wasn't sure about it either.

tmeasday: Ok, let's maybe just sort client side then.

Issue #14, opened 2019-09-18
Create a shim for fetch


tmeasday:
Looks like a combination of `unfetch`/Apollo throws in JSDOM if fetch is not defined.

![image](https://user-images.githubusercontent.com/132554/65102474-865ab680-da0e-11e9-8537-36dd37faf748.png)

PR #51, opened 2019-10-24
ADD unfetch as a fetch polyfill, as it's not in JSDom by default?

ndelangen:
issue: https://github.com/chromaui/chromatic-cli/issues/14
docs for polyfill: https://github.com/developit/unfetch
not included in jsdom yet: https://github.com/jsdom/jsdom/issues/1724

### DIFF
--- a/.github/workflows/chromatic-action.yml
+++ b/.github/workflows/chromatic-action.yml
@@ -1,0 +1,16 @@
+name: "Chromatic via action"
+on: [push, pull_request]
+
+jobs:
+  chromatic:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - run: |
+        yarn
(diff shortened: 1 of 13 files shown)


ndelangen:
Haven't tested this yet

tmeasday:
🤔 @ndelangen we don't really need the code to *work* just not error. Usually with these "shims" I would include way less code than this.

ndelangen:
So you'd rather have

```js
window.fetch = () => new Promise()
```

?

tmeasday:
I think so. Do you anticipate problems with that?

ndelangen:
Yeah if we ever allow async stories in storybook, we may need fetch in JSdom

ndelangen:
@tmeasday I changed it to a pure shim

Issue #31, opened 2019-10-06
Storybook upload does not work properly from windows


tmeasday:
![image](https://user-images.githubusercontent.com/132554/62741109-1b1bdd00-ba7d-11e9-8a35-22153ed0e22e.png)

I think what is happening here is the path we get for the assets uses the windows separator (which is correct on the client side) but we need to convert it to `/` when talking to AWS

ndelangen:
PR is open, with a fix!

PR #56, opened 2019-11-13
ADD a workflow that runs on windows

ndelangen:
https://github.com/chromaui/chromatic-cli/issues/31


ndelangen:
So I have it working on windows 🎉 

I guess the issue might be with static dirs? I'll try adding one with a few files in it.

tmeasday:
I guess so? If the upload doesn't have any subdirectories I guess the problem will not occur. Great job though

ndelangen:
FIXED IT!!! 🎉

Issue #59, opened 2019-11-16
We should use `prBranch` from `env-ci` when available.


tmeasday:
In release 3.0.0 https://github.com/pvdlg/env-ci/releases/tag/v3.0.0 `env-ci` included a breaking change that means the `branch` property points to the PR target, not source. We want the source!

PR #60, opened 2019-11-17
Use prBranch when available

ndelangen:
fixes: https://github.com/chromaui/chromatic-cli/issues/59

### DIFF
--- a/bin/git/getCommitAndBranch.js
+++ b/bin/git/getCommitAndBranch.js
@@ -6,6 +6,13 @@ import log from '../lib/log';
 
 export const debug = setupDebug('chromatic-cli:tester');
 
+const notHead = b => {
+  if (!b || b === 'HEAD') {
+    return false;
+  }
+  return b;
+};
+
 export async function getCommitAndBranch({ inputFromCI } = {}) {
   // eslint-disable-next-line prefer-const
   let { commit, committedAt, committerEmail, committerName } = await getCommit();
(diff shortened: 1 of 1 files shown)


bin/git/getCommitAndBranch.js
```diff
@@ -73,19 +73,20 @@ export async function getCommitAndBranch({ inputFromCI } = {}) {
...
+    // $GERRIT_BRANCH is for Gerrit/Jenkins: https://wiki.jenkins.io/display/JENKINS/Gerrit+Trigger
+    // $CI_BRANCH is a general setting that lots of systems use
+    branch =
+      prBranchFromEnvCi ||
+      process.env.HEAD ||
+      process.env.GERRIT_BRANCH ||
+      process.env.CI_BRANCH ||
+      process.env.GITHUB_REF ||
+      branchFromEnvCi ||
+      branch ||
```
tmeasday: @ndelangen this changes the logic when `prBranch` is undefined (which it is for many CI systems or when there is no PR), as it will now preference the environment vars. I'd prefer not to change the behaviour in such cases.

ndelangen: I thought this would be identical, can you explain to me what I need to change?

tmeasday: We might just need to be a bit more careful. 

Previously we grabbed `envCi().branch` and used it unless it was falsey or 'HEAD'.

We should probably do something like:

1. Use the git branch unless it is falsey or HEAD (existing check on line 75)
2. Use `envCi().prBranch` unless it is falsey or HEAD
3. Use `envCi().branch` unless it is falsey or HEAD
4. Use various env vars
5. Fall back to 'HEAD' or nothing

ndelangen: I think this is resolved now

Issue #62, opened 2019-11-19
Not wait for chromatic cloud todo its thing


benjaminkay93:
We currently use chromatic to build our storybook, upload it, and then wait for the build to complete like so.

`chromatic test --no-interactive --build-script-name=build`

We recently added `--exit-zero-on-changes` when we relised that the hook from chromatic back into our PR's would fail if stories had changed and then pass once we updated the story to be accepted. Thus meaning we always wanted to pass the CI step to avoi dhaving to run our Jenkins job in its entirety again.

This means the chromatic CLI now runs its build, uploads and waits for the success/failure to come back, essentially blocking our pipeline for 1.5 mins whilst chromatic does its stuff.

Is there anyway to just fire and forget the stories in our CI, so that when we run the chromatic CLI tool we just run the build and upload and not wait for a successful/failure of our stories inside chromatic itself? saving 1.5 mins of time because we have the chromatic hook to let us know if theres any failures?

tmeasday:
Hi @benjaminkay93 that sounds like a reasonable request. We've considered it before and it's certainly sensible.

Would you be interested in implementing it yourself? It would probably be a matter of creating the option in the `bin/register.js` file and passing it through to the `tester/index.js` file and exiting sometime after making this API call: https://github.com/chromaui/chromatic-cli/blob/master/bin/tester/index.js#L356-L390

Let me know if so. Otherwise we'll try and get it prioritized soon.

PR #63, opened 2019-11-20
Add --exit-once-sent-to-chromatic flag

benjaminkay93:
Added the flag, not sure how to test it locally to ensure that it works, would be a help if someone could advise, perhaps its just an npm link? 

This is raised as part of #62


tmeasday:
This is awesome @benjaminkay93!

tmeasday:
Hi @benjaminkay93, so I think our process is to check in the build `dist` directory as well (@ndelangen I think we were going to add some checks for this, should we make an issue for that?)

If you do that, you should be able to run this branch against any app by doing something like `yarn add benjaminkay93/chromatic-cli#master`

tmeasday:
OK, here's what I'm thinking about this:

1. Let's call the flag `--exit-once-uploaded`?
2. You shouldn't be able to use this flag and use the tunnel, so it should error if you set it and `--script-name` or `--do-not-start`.

benjaminkay93:
Yeah that's about right, thanks for sorting that for me :) 🚀

When will it be published?

ndelangen:
Should be released today or tomorrow!

Issue #64, opened 2019-11-22
Create deprecation warning for tunnel


tmeasday:
Let's drop the `--script-name` option from the docs/usage and also indicate to people they should stop using it.

 - [ ] Remove `--script-name` and `--no-start` from Chromatic docs / GH action? / CLI usage?
 - [ ] <strike>Add a deprecation warning if the above is used "we will remove in v4"</strike>

PR #69, opened 2019-12-03
REMOVE references to options that would use the tunnel

ndelangen:
fixes: https://github.com/chromaui/chromatic-cli/issues/64


bin/main.js
```diff
@@ -18,13 +18,10 @@ export async function parseArgv(argv) {
       
     Storybook options
       --build-script-name [name], -b  The npm script that builds your Storybook [build-storybook]
-      --script-name [name], -s  The npm script that starts your Storybook [storybook]
-      --exec <command>, -e  Alternatively, a full command to run to start your storybook
-      --do-not-start, -S  Don't attempt to start or build; use if your Storybook is already running
+      --storybook-build-dir, -d  <dirname>  Provide a directory with your built storybook; use if you've already built your storybook
       
       --storybook-port <port>, -p  What port is your Storybook running on (auto detected from -s, if set)?'
```
tmeasday: I guess we should drop the `--storybook-port` and https ones too?

ndelangen: I agree

PR #71, opened 2019-12-03
Fix/url shims should be writable

ndelangen:
needed for https://github.com/chromaui/action/pull/10


ndelangen:
@tmeasday going to release this in a patch, I really want to make progress on https://github.com/chromaui/action/pull/10 and this is blocking me.

PR #72, opened 2019-12-05
ADD extra information about build to github action

ndelangen:
errorCount
specCount
componentCount
changeCount
buildNumber

### DIFF
--- a/bin/tester/index.js
+++ b/bin/tester/index.js
@@ -329,7 +329,10 @@ export async function runTest({
   let componentCount = 0;
   let changeCount = 0;
   let buildNumber = 0;
+  let snapshotCount = 0;
   let exitUrl = '';
+  let diffs;
+  let buildStatus;
 
   const { cleanup, isolatorUrl, cachedUrl } = await prepareAppOrBuild({
     storybookVersion,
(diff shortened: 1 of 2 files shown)


bin/tester/index.js
```diff
@@ -392,11 +397,14 @@ export async function runTest({
       isolatorUrl,
     });
 
+    buildNumber = number;
```
tmeasday: I think I'd prefer:

```js
const createBuild = await ...;

({ number: buildNumber, specCount, componentCount }) = createBuild;
```

The extra variable names are just confusing IMO

bin/tester/index.js
```diff
@@ -406,19 +414,21 @@ export async function runTest({
...
+      status: buildStatus,
       autoAcceptChanges: buildAutoAcceptChanges, // if it is the first build, this may have been set
-      changeCount,
-      errorCount,
-    } = await waitForBuild(client, { buildNumber: number }, { diffs });
+    } = buildOutput;
 
-    switch (status) {
+    changeCount = buildOutput.changeCount;
+    errorCount = buildOutput.errorCount;
```
tmeasday: See above for a cleaner way to write this.

Issue #68, opened 2019-12-02
Drop `NODE_ENV=development` when building


tmeasday:
This is a historical accident. I'm not sure if it is a breaking change or not.

lifeiscontent:
@tmeasday seems in CI when chromatic is running, it doesn't include env vars. so my storybook is busted until I can figure out how to overcome this. (using next.js) in preview.js

```js
import { setConfig } from 'next/config';

setConfig({
  publicRuntimeConfig: {
    API_HOST: process.env.API_HOST,
    GOOGLE_MAPS_API_KEY: process.env.GOOGLE_MAPS_API_KEY,
  },
});
```

apparently API_HOST and GOOGLE_MAPS_API_KEY are undefined when chromatic runs in CI

tmeasday:
@lifeiscontent are the env vars used in the frontend? This seems like a storybook issue perhaps, we don't drop any env vars we just add a couple: https://github.com/chromaui/chromatic-cli/blob/master/bin/storybook/start-app.js#L60:L64

lifeiscontent:
@tmeasday thanks for the response!

We’re using next.js so in the storybook config I’m using setConfig from “next/config” and setting the env vars we need from process.env.

I’m also using a custom web pack config for storybook to step over some of the issues with next.js e.g. Babel configs

lifeiscontent:
also, fwiw, everything works as expected in development.

tmeasday:
Does it work when you run `yarn build-storybook` though @lifeiscontent?

lifeiscontent:
@tmeasday great question, let me try it!

lifeiscontent:
@tmeasday yep, it builds, should I try and run it with serve?

lifeiscontent:
@tmeasday I just ran it with serve, everything works fine.

tmeasday:
You don't get any errors to the console?

lifeiscontent:
@tmeasday None!

lifeiscontent:
@tmeasday pinged you on Intercom

PR #81, opened 2020-01-14
REMOVE nodeenv development

ndelangen:
fixes: https://github.com/chromaui/chromatic-cli/issues/68


tmeasday:
What about the `BROWSER` bit? I don't know why that's there either.

tmeasday:
The commit where it cate from is super old: https://github.com/chromaui/chromatic/commit/e1a9ba09283ff96a6c9b369e855db5c3bd41efba

ndelangen:
The NODE_ENV=TEST was to make babel parse the example storybook correctly 😖 

Removing it caused the example to fail on require.context

I was able to make the example work by removing the babel stuff for webpack's require.context, and migrating to tri-config.

ndelangen:
I merged in the better error handling, because it actually helped my find & test the problem.

I could remove it from this branch again if you really want @tmeasday or I could close the other PR, and consider it 1 cleanup PR. WDYT?

tmeasday:
Hold on a sec.

ndelangen:
I cleaned the PR up @tmeasday

.babelrc.js
```diff
@@ -4,9 +4,6 @@ module.exports = {
       presets: [
         ['@babel/preset-env', { targets: { node: 'current' } }],
       ],
-      plugins: [
-        'require-context-hook'
-      ],
```
ndelangen: The environment variable NODE_ENV=test essentially caused the static storybook to be build with this babel config.

tmeasday:
Thanks it looks reasonable. I'm just worried that you had to do this. Does it mean that Chromatic doesn't work any more with non-`main.js` storybooks?

ndelangen:
NO AFAICS it should work without / should work with the old config.js just fine @tmeasday

tmeasday:
So why did you need to do it in our test storybook? Can you explain what happened? Is the `BROWSER` var needed?

ndelangen:
@tmeasday I think build-storybook uses the babel config defined in the root. It's always done that, but because NODE_END was set to 'development' and in the root there's no babel config for that env, it wouldn't override anything.

Removing NODE_ENV='development' made it act as NODE_ENV='test' which DOES match a env specified in the root babel config, and thus it's applied.

Turns out if we add `NODE_ENV=test` here:
https://github.com/chromaui/chromatic-cli/blob/fb07c9314ff587ad2623ca79d9b9fe92cf7d0072/bin/register.js#L3

IMHO that should be `production` WDYT?

ndelangen:
We can't know exactly what impact a NODE_ENV change will have on the users config/code-output. Perhaps it's safer to release this as a major?

ndelangen:
AFACS there's no usage of process.env.BROWSER whatsoever. Removing seems to do absolutely nothing.

tmeasday:
OK, setting `NODE_ENV=test` there seems not correct either. Do you know why we did that? Was it for the sake of dotenv?

> We can't know exactly what impact a NODE_ENV change will have on the users config/code-output. Perhaps it's safer to release this as a major?

You are probably right.

ndelangen:
I'll delete this code and that makes this a major release

```
process.env.NODE_ENV = process.env.NODE_ENV || 'test'; 
```

Issue #75, opened 2019-12-10
Don't continue if the storybook throws an error


tmeasday:
## What

In order to avoid roadbumps we decided to ignore logged errors when evaluating your story list and getting a list of stories (sometimes people's code logs errors but it's not actually an issue).

We should re-evaluate this behaviour or add a flag to opt-into it.

## Why

Often errors are logged when evaluating story files (if there's a JS bug in one of the files). What this means (e.g [this build](https://www.chromaui.com/build?appId=59c59bd0183bd100364e1d57&number=7645) at [this url](https://59c59bd0183bd100364e1d57-mkwmrssyja.chromatic.com/iframe.html)) is that we end up getting a truncated list of stories (all the ones that evaluated before the error was thrown).

Chromatic does not make you accepted deleted stories (which is a tricky problem actually) and this leads to a bad time.

However, some users have stuff logged to their error console as a matter of course and can't necessarily remove it easily. So we still need a way to opt out.

## How

1. Let's change the default behaviour to cancel the build if there's something logged to the error console. 

2. Let's add a flag `--allow-console-errors` that reverts to the current behaviour

3. In the case of 1., let's say something like: "Your Storybook logged the following to the error console while loading. Errors loading your Storybook can lead to Chromatic getting your list of stories wrong and can cause a variety of problems. If you want to continue even with the errors, pass the `--allow-console-errors` flag to the `chromatic` command, however we do not recommend it."

ndelangen:
This is another potential gain from tri-config and declarative stories definition, we could isolate the runtime of each story file being added.

zol:
This just bit a customer, see here: https://app.intercom.io/a/apps/zj7sn9j1/inbox/inbox/all/conversations/25290402621

We should include this in our next package release and inform them.

tmeasday:
@ndelangen this is not working as expected on `next`. I need to set `LOG_LEVEL=verbose` to see *any* messages at all about errors now. Try

```bash
LOG_LEVEL=verbose yarn chromatic -a <code> -u https://59c59bd0183bd100364e1d57-mkwmrssyja.chromatic.com/iframe.html
```

vs

```bash
yarn chromatic -a <code> -u https://59c59bd0183bd100364e1d57-mkwmrssyja.chromatic.com/iframe.html
```

tmeasday:
Also `--allow-console-errors` doesn't appear to work.

ndelangen:
[...]

PR #82, opened 2020-01-14
ADD throw on runtime errors

ndelangen:
fixes: https://github.com/chromaui/chromatic-cli/issues/75#issuecomment-572265026

### DIFF
--- a/bin/tester/runtimes.js
+++ b/bin/tester/runtimes.js
@@ -48,15 +48,15 @@ export default async function getRuntimeSpecs(url, { verbose = false } = {}) {
   });
 
   const hasErrors = errors.length;
-  const hasVisibileErrors = hasErrors && log.level.match(/verbose/);
+  const hasVisibleErrors = hasErrors && log.level.match(/verbose/);
   const hasVisibileWarnings = warnings.length && log.level.match(/verbose/);
 
-  if (hasVisibileErrors || hasVisibileWarnings) {
+  if (hasVisibleErrors || hasVisibileWarnings) {
     log[errors.length ? 'error' : 'warn'](
       'The following problems were reported from your storybook:'
     );
 
-    if (hasVisibileErrors) {
+    if (hasVisibleErrors) {
       console.log(
         errors.reduce(
           (acc, i) => dedent`
--- a/bin/tester/runtimes.js
+++ b/bin/tester/runtimes.js
@@ -47,17 +47,16 @@ export default async function getRuntimeSpecs(url, { verbose = false } = {}) {
     }, 60000);
   });
 
-  // If the app logged something to console.error, it's probably, but not definitely an issue.
-  // See https://github.com/hichroma/chromatic/issues/757
-  if (
-    (errors.length && !log.level.match(/silent/)) ||
-    (warnings.length && log.level.match(/verbose/))
-  ) {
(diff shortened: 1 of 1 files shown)


bin/tester/runtimes.js
```diff
@@ -89,11 +89,20 @@ export default async function getRuntimeSpecs(url, { verbose = false } = {}) {
       );
     }
 
-    console.log(dedent`
-        This may lead to some stories not working right or getting detected by Chromatic
-        We suggest you fix the errors, but we will continue anyway..
-        ${separator}
-      `);
+    if (errors.length && !log.level.match(/silent/)) {
```
tmeasday: Are we going to allow them to continue if log level is silent? That seems incorrect, maybe?

bin/tester/runtimes.js
```diff
@@ -89,11 +89,20 @@ export default async function getRuntimeSpecs(url, { verbose = false } = {}) {
...
+      console.log(dedent`
+          This very likely caused some stories not working right or getting detected by Chromatic
+          Please fix the errors, we can't continue..
+          ${separator}
+        `);
+      throw new Error('Errors detected in storybook runtime');
+    } else if (warnings.length && log.level.match(/verbose/)) {
```
tmeasday:
```suggestion
    } else if (hasVisibileWarnings) {
```

Can we pull this out to a variable? I guess we are using it 3 times now.

bin/tester/runtimes.js
ndelangen: 👍

bin/tester/runtimes.js
ndelangen: sure

bin/tester/runtimes.js
```diff
@@ -47,17 +47,16 @@ export default async function getRuntimeSpecs(url, { verbose = false } = {}) {
...
   });
 
-  // If the app logged something to console.error, it's probably, but not definitely an issue.
-  // See https://github.com/hichroma/chromatic/issues/757
-  if (
-    (errors.length && !log.level.match(/silent/)) ||
-    (warnings.length && log.level.match(/verbose/))
-  ) {
+  const hasErrors = errors.length;
+  const hasVisibileErrors = hasErrors && log.level.match(/verbose/);
```
tmeasday:
```suggestion
  const hasVisibleErrors = hasErrors && log.level.match(/verbose/);
```

bin/tester/runtimes.js
```diff
@@ -89,11 +88,20 @@ export default async function getRuntimeSpecs(url, { verbose = false } = {}) {
       );
     }
 
-    console.log(dedent`
-        This may lead to some stories not working right or getting detected by Chromatic
-        We suggest you fix the errors, but we will continue anyway..
-        ${separator}
-      `);
+    if (hasErrors) {
```
tmeasday: If they have suppressed errors we probably need to say something here.

I wonder if we should just drop the silent behaviour? Is it useful to not show errors if we are going to stop?

tmeasday: OK, I took a look and silent is only used via the `DISABLE_LOGGING` env var, which we use in tests, so we don't need to worry to much about it.

ndelangen: cool

PR #85, opened 2020-01-16
Fix/better error reporting

ndelangen:
fixes: https://github.com/chromaui/chromatic-cli/issues/73

### DIFF
--- a/bin/__tests__/tester.test.js
+++ b/bin/__tests__/tester.test.js
@@ -18,7 +18,7 @@ const defaultOutput = {
   specCount: 1,
 };
 
-jest.mock('node-fetch', () => (url, { body }) => ({
+jest.mock('node-fetch', () => async (url, { body } = {}) => ({
   ok: true,
   json: async () => {
     const { query, variables } = JSON.parse(body);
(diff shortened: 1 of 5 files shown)


tmeasday:
Can we not look at the return value of the `build-storybook` command to tell if it failed?

It seems weird to try and upload an empty folder and only stop once we try and request it again later.

ndelangen:
@tmeasday We do, I think:
https://github.com/chromaui/chromatic-cli/blob/a1e26a7e340732042809055ffc36c000996c262e/bin/tester/index.js#L106-L113

But for some reason the exitCode of build-storybook is often a 0, even though something went wrong, it seems. Or our code for handling exitcode === 0 isn't right. Do you see something wrong with it?

tmeasday:
Hmm, this all seems reasonable. Does the build-storybook command ever exit with a non-zero exit code? Perhaps we should open a bug about this on storybook's end?

ndelangen:
> Hmm, this all seems reasonable. Does the build-storybook command ever exit with a non-zero exit code? Perhaps we should open a bug about this on storybook's end?

@tmeasday yes my current working theory is there's some reason storybook's build script exit's with a 0, even if something went wrong. It's possible some particular error that isn't handled, or is ignored. We should find out exactly what that scenario is and address it in 6.0.0.

PR #92, opened 2020-02-04
FIX chromaui/action/issues/14

ndelangen:
chromaui/action/issues/14


ndelangen:
@tmeasday I'm confused why it doesn't work.

The GithubAction does pass `fromCI=true`

But for good measure I've added an environment variable that should be set in Github Actions to the list of flags that should make from CI true.

If this change is ok to you, I'll release a patch of chromatic-cli & update the action.

tmeasday:
Does it work? If so, LGTM

ndelangen:
verified it works

PR #101, opened 2020-02-28
Update 'Storybook was not build' language

kylesuss:
Another docs language update based on an error I saw come through support.

PS -- am I right to target these at master? I see there is also a next branch but it is not the default branch.


kylesuss:
Nevermind -- I see @ndelangen reverted the last PR and targeted it at next. I'll close this & re-open.

PR #105, opened 2020-03-25
Fix/94 needs crossenv

### LINKED ISSUE #94: Error with adding script through cli
While working on a pr on the learnstorybook repo i was following the steps of the tutorial up until the [test section](https://www.learnstorybook.com/intro-to-storybook/react/en/test/) all is good up until the i'm executing the `npx chromatic --app-code=<app-code>` command. I let it run to completion, only when the process finishes and i'm presented with the following:
```bash
You have not added the 'chromatic' script to your 'package.json'. Would you like me to do it for you? y
chromatic info Added script 'chromatic'. You can now run it here or in CI with 'npm run chromatic' (or 'yarn chromatic')

NOTE: I wrote your app code to the 'CHROMATIC_APP_CODE' environment variable.

The app code cannot be used to read story data, it can only be used to create new builds.
If you would still prefer not to check it into source control, you can remove it from 'package.json' and set it via an environment variable instead.
```

As you can see i selected the yes (y) option to allow the cli to create a script to my project.
Now checking the `package.json` file i'm presented with the following:
```json
{
  "scripts": {
    "start": "react-scripts start",
    "build": "react-scripts build",
    "test": "react-scripts test",
    "eject": "react-scripts eject",
    "storybook": "start-storybook -p 9009 -s public --ci",
    "build-storybook": "build-storybook -s public",
    "chromatic": "CHROMATIC_APP_CODE=my-app-code chromatic"
  },
 
  "devDependencies": {
    "@storybook/addon-actions": "^5.3.12",
    "@storybook/addon-links": "^5.3.12",
    "@storybook/addon-storyshots": "^5.3.13",
    "@storybook/addons": "^5.3.12",
    "@storybook/preset-create-react-app": "^1.5.2",
    "@storybook/react": "^5.3.12",
    "react-test-renderer": "^16.12.0",
    "storybook-chromatic": "^3.5.1"
  }
}

```
I'm just leaving in the necessary information, namely the devDependencies and the scripts available.

Now as me or any other user that goes with this route as soon as he/she makes any change and tries to run the script the following will be displayed:
```bash
'CHROMATIC_APP_CODE' is not recognized as an internal or external command,
operable program or batch file.
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! taskbox-for-pr-278@0.1.0 chromatic: `CHROMATIC_APP_CODE=my-app-code chromatic`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the taskbox-for-pr-278@0.1.0 chromatic script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     \AppData\Roaming\npm-cache\_logs\2020-02-12T19_57_13_233Z-debug.log
The terminal process terminated with exit code: 1
```
This is not the first time this has popped up for me. while i was doing some work on another piece of documentation this issue presented itself, but i thought it might be a fluke, probably the package manager acting up or something on that nature.

Is this intended behaviour or something is amiss?

To add a bit more of context, i'm running on a windows 10 machine. Should anymore information be necessary i'm more than happy to push the code i'm using to a repo on GitHub.

Feel free to provide feedback

ndelangen:
FIX #94


bin/lib/package-json.js
```diff
@@ -21,5 +22,14 @@ export function addScriptToPackageJson(scriptName, scriptCommand, { appDir = pro
...
   packageJson.scripts[scriptName] = scriptCommand;
+
+  if (!packageJson.dependencies) {
+    packageJson.dependencies = {};
+  }
+  if (!packageJson.dependencies['cross-env']) {
+    packageJson.dependencies['cross-env'] = `^${execSync('npm show cross-env version')
+      .toString()
+      .trim()}`;
+  }
```
tmeasday: Wouldn't we need to run `npm/yarn` install after this step?

Is it impossible for us to depend on `cross-env` and still use it somehow in a script?

ndelangen: We're adding a script to the user's package.json. A script that uses cross-env.
So we should also add that dependency to their package.json.

I don't think we should attempt to actually install the package.

tmeasday: But I mean its not going to work when they try and run chromatic in that case.

ndelangen: ok, it wasn't so hard to add after all.

tmeasday:
Hmmm all this complication makes me think that maybe we should reconsider setting the env var after all. I was hoping we could just call cross-env from our dependencies somehow.

ndelangen:
Let's move it into the cli flag

PR #110, opened 2020-04-01
Create patch build

ghengeveld:
Fixes https://github.com/chromaui/chromatic/issues/3238


bin/git/git.js
```diff
@@ -252,3 +251,52 @@ export async function getBaselineCommits(client, { branch, ignoreLastBuildOnBran
...
+  const mergeBases = result.split(EOL).filter(Boolean);
+  if (mergeBases.length === 1) return mergeBases[0];
+
+  const branchNames = await Promise.all(
+    mergeBases.map(async sha => {
+      const name = await execGitCommand(`git name-rev --name-only --exclude="tags/*" ${sha}`);
+      return name.replace(/~[0-9]+$/, ''); // Drop the potential suffix
+    })
+  );
+  return mergeBases[branchNames.findIndex(branchName => branchName === baseBranch)];
```
tmeasday, line 293: What if none is on the `baseBranch`?

bin/tester/patchBuild.js
```diff
@@ -0,0 +1,49 @@
...
+      Failed to retrieve the merge-base. You may have specified an invalid base branch.
+      Are you sure the head branch is a descendant (i.e. fork) of the base branch?
+        (try running this command yourself: "git merge-base ${headRef} ${baseRef}")
+    `);
+  }
+
+  checkout(mergeBase);
+  installDependencies();
+
+  await runTest({ ...options }); // TODO set branch
```
tmeasday, line 45: We need to tell the index that this is a patch build so it can do the [follow up steps](https://github.com/chromaui/chromatic/issues/3238#issuecomment-605889033):

 4. We have to trigger a comparison when this patch build is created (normally a base build won't trigger one).
 5. The patch build should be appended to the parent builds of the build on the head branch most recently created after the committedAt date for the patch build.

Note that:

 A. Neither of these steps are possible if the head build doesn't exist yet
 B. Even if the head build exists, we still need to know the PR number in order to find the head build & PR (just the branch alone isn't enough as you can have more than one PR per branch -- or is it possible to tell from the combination of `headRef` & `baseRef`?)

Given A), I think we need to either: 

 i. bail if the head build doesn't exist yet (otherwise there's little point in all this) OR
 ii. create the head build first. OR
 iii. create the head build second *but* tell it the "merge base to use" somehow.

Given B) (unless I am wrong about telling the PR from the combination), I think we need to *pass the PR number* with this command, and look up the `baseRef`/`headRef` on the index.


I would suggest we maybe start with Ai. for now (which will require looking up the head build on the index) and in our instructions tell the user to run a build on the head branch *first*?

WDYT?

PR #112, opened 2020-04-06
CHANGE the output when building storybook fails

ndelangen:
This is the new situation when there's an error from storybook:
Screenshot 2020-04-06 at 19 27 56

This is when everything is fine:
Screenshot 2020-04-06 at 19 30 13


ndelangen:
The red markings are custom from my terminal (I've configured it to always highlight the word "error"

bin/tester/index.js
```diff
@@ -87,38 +87,60 @@ async function prepareAppOrBuild({
...
       // Wait for the process to exit
       await new Promise((res, rej) => {
-        child.on('error', rej);
+        const lines = [];
+        const rl = readline.createInterface({ input: child.stderr });
+
         child.on('close', code => {
           if (code !== 0) {
-            rej(new Error(`${buildScriptName} script exited with code ${code}`));
+            console.log('');
```
tmeasday, line 113: FYI I think `console.log()` has the same effect

ndelangen: This will put the error message at the bottom.

tmeasday: I'm just saying I think `console.log() === console.log('')` 🤷‍♂

ndelangen: ah, i see 👍

Issue #115, opened 2020-04-12
Cannot call static methods on `Intl` mocks


tmeasday:
This line of code (which works in a browser) breaks:

```js
Intl.DateTimeFormat.supportedLocalesOf(['ban'])
```

Related to https://github.com/chromaui/chromatic-cli/pull/53 and https://github.com/chromaui/chromatic-cli/issues/48

ndelangen:
strange:
https://github.com/chromaui/chromatic-cli/blob/19751d87d950a2aecefb522e57c9a13c8c34fe54/bin/lib/jsdom-shims.js#L261-L264

code is there, I'm not sure why it wouldn't work.

ndelangen:
If I just add 
```
console.log(Intl.DateTimeFormat.supportedLocalesOf(['ban']));
```

to a story is works in chromatic.

ndelangen:
I've also tried using `react-intl` in a few various ways. I've not been able to reproduce yet..

ndelangen:
I've been searching for where "intlConstructor" might be from, and this is what i found:

https://github.com/formatjs/intl-locales-supported/commit/f2faf42b1cbdc1ebf6c958a0069a333d861f0ade#diff-168726dbe96b3ce427e7fedce31bb0bcR23-R38

that repo is deprecated however, it's related to react-intl:
https://github.com/formatjs/react-intl

(It's in the same github org.)

ndelangen:
https://www.staging-chromatic.com/component?appId=5e92fbc4f60aba00228b8a0a&name=Intl&buildNumber=2&mode=snapshot

seems to work fine?

ndelangen:
Is it IE11 specific perhaps?

I am running this in IE11 and it fails:
https://5d3a9b0c6b939d00207f7729-mqrplctozt.staging-chromatic.com/?path=/story/basics-a--simple

ndelangen:
It seems like react-intl might have some code in it that's IE11 incompatible:
<img width="403" alt="Screenshot 2020-04-12 at 13 56 11" src="https://user-images.githubusercontent.com/3070389/79068215-98ab8300-7cc5-11ea-99c8-62ca6bba720e.png">

ndelangen:
But I don't really how that might cause: 
> `Cannot call static methods on `Intl` mocks`

tmeasday:
@ndelangen you need it to be outside of a story so it is evaluated in JSDOM

tmeasday:
[...]

PR #118, opened 2020-04-12
HOTFIX #115 on master

ndelangen:
fixes: #115 

overhaul the shims, so it's only applied when needed, & overhaul all parts into their own function.

This is a hotfix on master


tmeasday:
I don't really see how this fixes #115. 

Probably easier to use the reproduction contained in [this comment](https://github.com/chromaui/chromatic-cli/pull/120#issue-402447526) to test things out.

PR #120, opened 2020-04-13
Fix Intl mocks version 2

tmeasday:
Hotfix for #115.

Do the minimum to make our `alwaysFn()` work with static methods.

To verify that `alwaysFn` works, try the following:

```js
const alwaysFn = /*  as defined */;

class A {
  static method() {}
}

class B {}

const Bprime = alwaysFn(B);

Bprime.method();
```


ndelangen:
verified it works

PR #123, opened 2020-04-15
Show "Continue setup at ..." link after running the very first build

ghengeveld:
This is meant to prevent users from ending up on the builds page while they're supposed to be following the onboarding flow.

Screenshot 2020-04-15 om 14 39 22

This is only shown for build 1.

Also I removed the `.` after the link because my terminal was treating it as part of the URL when clicked.

### DIFF
--- a/bin/tester/index.js
+++ b/bin/tester/index.js
@@ -373,6 +373,7 @@ export async function runTest({
   let uiReview;
   let wasLimited;
   let billingUrl;
+  let setupUrl;
   let exceededThreshold;
   let paymentRequired;
 
@@ -430,6 +431,7 @@ export async function runTest({
         wasLimited,
         app: {
           account: { billingUrl, exceededThreshold, paymentRequired },
+          setupUrl,
         },
       },
     } = await client.runQuery(TesterCreateBuildMutation, {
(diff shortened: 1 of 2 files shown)


bin/tester/index.js
```diff
@@ -483,17 +483,19 @@ export async function runTest({
...
     }
 
-    if (!publishOnly) {
-      log.info(dedent`
-      Started Build ${buildNumber} (${pluralize(componentCount, 'component')}, ${pluralize(
-        specCount,
-        'story'
-      )}, ${pluralize(snapshotCount, 'snapshot')}).
+    const onlineHint =
+      buildNumber === 1 ? `Continue setup at ${setupUrl}` : `View it online at ${exitUrl}`;
```
tmeasday, line 487: I wonder if we should try to be smarter here? On the backend we have some extra logic to deal with the not completely uncommon case of the user having a few builds fail as they get things setup: https://github.com/chromaui/chromatic/blob/release/services/index/model/Build.js#L727-L733

Perhaps we should check if `build.autoAcceptChanges=true` but we didn't ask for it to be?

ghengeveld: Excellent suggestion, love it! Done.

tmeasday:
Good stuff @ghengeveld thankyou

PR #130, opened 2020-04-24
Update support emails

kylesuss:


bin/main.js
```diff
@@ -98,7 +98,7 @@ export async function run(argv) {
       dedent`
         ** Chromatic build failed. **
         Please note the session id: '${options.sessionId}'
-        contact support@hichroma.com -or- open a support ticket at https://chromaticqa.com
+        contact support@chromatic.com -or- open a support ticket at https://chromatic.com
```
tmeasday, line 101:
```suggestion
        contact support@chromatic.com -or- open a support ticket at https://www.chromatic.com
```

https://chromatic.com (or qa.com) actually will not work in the browser!

kylesuss: Good call.

PR #140, opened 2020-05-08
New UI based on Listr

ghengeveld:


ndelangen:
Looking good!

ghengeveld:
> Looks great. Love the stories

A lot more incoming! I made stories for everything now. Will push today.

tmeasday:
@ghengeveld this is pretty hard to review comprehensively at this point, I think we are probably best to just publish a beta and do some extensive QA.

PR #148, opened 2020-06-09
Generate JUnit XML report file when passing --report flag

ghengeveld:
Fixes chromaui/chromatic#3617

See also chromaui/chromatic#3685

### DIFF
--- a/bin/tasks/snapshot.js
+++ b/bin/tasks/snapshot.js
@@ -1,9 +1,13 @@
 /* eslint-disable no-param-reassign */
+import reportBuilder from 'junit-report-builder';
+import path from 'path';
+
 import { createTask, transitionTo } from '../lib/tasks';
 import { delay, matchesBranch } from '../lib/utils';
 import buildHasChanges from '../ui/messages/errors/buildHasChanges';
 import buildHasErrors from '../ui/messages/errors/buildHasErrors';
 import speedUpCI from '../ui/messages/info/speedUpCI';
+import wroteReport from '../ui/messages/info/wroteReport';
 import {
   buildComplete,
   buildError,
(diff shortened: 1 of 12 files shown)


bin/tasks/snapshot.js
```diff
@@ -29,6 +33,82 @@ const TesterBuildQuery = `
...
+      .testCase()
+      .className(spec.component.name)
+      .name(`${spec.name}${suffix}`);
+
+    switch (status) {
+      case 'SNAPSHOT_ERROR':
+        testCase.error('Server error while taking snapshot, please try again', status);
+        break;
+      case 'SNAPSHOT_CAPTURE_ERROR':
+        testCase.error('Snapshot is broken due to a JavaScript error in your Storybook', status);
```
tmeasday, line 96: This isn’t necessarily a js error. Eg could be image too large etc

bin/tasks/snapshot.js
```diff
@@ -29,6 +33,82 @@ const TesterBuildQuery = `
...
+        break;
+      case 'SNAPSHOT_DENIED':
+        testCase.failure('Snapshot was denied by a user', status);
+        break;
+      case 'SNAPSHOT_NO_CAPTURE':
+        testCase.skipped();
+        break;
+      default:
```
tmeasday, line 104: What about pending? These are the snapshots the user likely cares about

ghengeveld: Why yeah, pending should be a failure too of course.

ghengeveld: By the way the switch is only to set the error/failure/skipped flag, the pending snapshots were part of the report before, but marked 'successful'.

PR #150, opened 2020-06-11
Fix yarn cmd in README

oliverlloyd:
The yarn remove command doesn't allow the `--dev` param and will throw an error if the documented code is run

`yarn remove --dev ...` > `yarn remove ...`


tmeasday:
Thanks @oliverlloyd -- appreciate it!

PR #165, opened 2020-07-21
Use actual output dir and warn about it

### LINKED ISSUE #162: Detect or even handle when build-storybook script includes `-o`
We quite commonly get this issue when teams' `build-storybook` script includes a `-o`:

> ✖ Failed to verify your Storybook\nJavaScript failed to load at ...

The URL posted 404s as the directory we uploaded was empty.

We should handle this better by:

1. detecting the directory is empty, and telling them the likely reason (good) 👉 #163
2. detecting the script has a flag, and warning them if it's not going to yield what they expect
3. actually just grabbing their SB from the specified output location if they set it

We should detect the actual output dir from what build-storybook returns, instead of relying on our passed output dir to have been propagated and used. If the two locations differ, we can additionally warn the user about this.

ghengeveld:
Fixes #162


bin/tasks/build.js
```diff
@@ -37,6 +40,15 @@ export const setSpawnParams = ctx => {
   };
 };
 
+const getOutputDir = buildLog => {
+  const outputIndex = buildLog.lastIndexOf('Output directory: ');
+  if (outputIndex === -1) return undefined;
+  const remainingLog = buildLog.substr(outputIndex + 18);
```
tmeasday, line 46: What's this magic number 18? Is it the length of `'Output directory: '`?

bin/tasks/build.js
```diff
@@ -51,6 +63,13 @@ export const buildStorybook = async ctx => {
   } finally {
     logFile.end();
   }
+
+  const buildLog = fs.readFileSync(ctx.buildLogFile, 'utf8');
+  const outputDir = getOutputDir(buildLog);
+  if (outputDir && outputDir !== ctx.sourceDir) {
+    ctx.log.warn(deviatingOutputDir(ctx, outputDir));
+    ctx.sourceDir = outputDir;
+  }
```
tmeasday, line 72: This process seems a bit brittle (especially when you consider future versions of SB may not have the same log format). What do you think about only doing if it we *don't* find the build in the place we expected?, ie  ala https://github.com/chromaui/chromatic-cli/pull/164)

bin/tasks/build.js
ghengeveld: Yep

bin/tasks/build.js
ghengeveld: Moved to upload task so we only run this if the sourceDir isn't a valid build.

PR #194, opened 2020-10-15
Better ways to retrieve the branch name

ghengeveld:
This should hopefully avoid the `origin/*` prefix as well.


tmeasday (approved):
Ok, I don't know enough about git internals to know if this is ideal but seems good.

PR #200, opened 2020-10-27
improve and upgrade dependencies

ndelangen:
I cleaned up some dependencies, removed unneeded ones, synchronized eslint config with storybook (removing more dependencies), fixed the linting issues

I couldn't upgrade react yet
I couldn't upgrade ts-dedent to 2.0.0 it seems to have a bug now


package.json
```diff
@@ -65,79 +65,64 @@
   "dependencies": {
     "@actions/core": "^1.2.4",
     "@actions/github": "^4.0.0",
-    "@babel/preset-react": "^7.9.4",
-    "@babel/runtime": "^7.9.2",
```
tmeasday, line 69: We don't need this any more? Are you sure @ndelangen?

tmeasday (changes requested):
One thing seems awry, otherwise thanks for this!

package.json
ndelangen: Everything seems to work fine without them?

ghengeveld:
Can you please pull out the ESLint stuff into a separate PR? This is really unwieldy.

package.json
tmeasday: Hmm, I'm taking specifically about `@babel/runtime`. I suspect it could break if you used the package in a project which itself didn't have a dependency on `@babel/runtime`, as `@babel/plugin-transform-runtime` is going to put down a bunch of `require('@babel/runtime')` calls into to the transpiled code.

ndelangen: fixed!

ndelangen:
@ghengeveld that's going to be quite a bit of work. Are you sure it's worth it?

ndelangen:
@ghengeveld can we merge this this week perhaps?

Issue #179, opened 2020-08-27
Don't allow builds with no git history


tmeasday:
If someone runs the CLI in a project with no history, show an error and fail.

In particular for GH actions where the default is to have no history.

ndelangen:
The default in github's checkout v2 action is a history with just a single commit on it

ndelangen:
I'm thinking what we're really after is changing this:
https://github.com/chromaui/chromatic-cli/blob/6ca39d6847c4352e21c38c927557235064e4a9f8/bin/git/git.js#L211-L214

alternatively we check if there's at least 2 commits?
```
await execGitCommand(`git log -n 1 --format="%H,%ct,%ce,%cn" --skip=1`)
```

WDYT @tmeasday ?

tmeasday:
Something like that, although I'm not sure that's the right step to check it, I would have thought somewhere earlier maybe here-ish: https://github.com/chromaui/chromatic-cli/blob/6ca39d6847c4352e21c38c927557235064e4a9f8/bin/git/git.js#L238

ghengeveld:
Fixed in #207

PR #202, opened 2020-11-05
179/fix for no history in actions

ndelangen:
issue: https://github.com/chromaui/chromatic-cli/issues/179

check for a second commit, and if it doesn't exists throw and error

### DIFF
--- a/.github/workflows/action-run.yml
+++ b/.github/workflows/action-run.yml
@@ -10,6 +10,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - run: yarn && yarn build && yarn build-storybook
     - uses: ./
       with: 
--- a/bin/git/git.js
+++ b/bin/git/git.js
@@ -88,6 +88,8 @@ export async function getCommit() {
     await execGitCommand(`git log -n 1 --format="%H,%ct,%ce,%cn"`)
   ).split(',');
 
+  console.log({ commit, committedAt: committedAtSeconds * 1000, committerEmail, committerName });
+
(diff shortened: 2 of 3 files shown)


ghengeveld:
Damn, looks like we both worked on this: https://github.com/chromaui/chromatic-cli/pull/207

ghengeveld:
@ndelangen I like your way of checking the number of commits using `git log --max-count=1 --skip=1`. That seems more efficient than running a command to get the total number of commits like I did.

What I didn't like about your solution is that it throws when there's only one commit, while mine only warns. Technically a repo with only one commit _is_ valid and possible. It's just unlikely to be true.

@tmeasday Do you think we should warn or fail if we find there's only one commit?

FYI we [already fail](https://github.com/chromaui/chromatic-cli/blob/152dc03ad70d00ba79516f89f7b6ed14f0453d99/bin/git/git.js#L35) if there's zero commits. No need to handle that scenario any different.

tmeasday:
> @tmeasday Do you think we should warn or fail if we find there's only one commit?

We decided to **fail**. It's just infinitely more common that someone is using a CI system that only checks out the current commit that somehow a user is using a repo with only one commit in it [1]. 

I think as long as the message says something like "if you legitimately only have one commit in this repo, make another one and run chromatic again".


[1] I suspect this will only ever happen to us during QA.

ghengeveld:
Alright. I've updated my PR with the following:
- Used Norbert's `-n 1 --skip=1` trick to find the previous commit rather than the total count. We don't really need the count anyway.
- Updated all git error messaging to use the UI patterns that we have (and create stories for them)
- Changed the shallow clone warning to an error, so it'll fail.

Closing this in favor of #207.

Issue #196, opened 2020-10-16
Support `pull_request_target` and `pull_request_review`


ghengeveld:
By @NicholasBoll in https://github.com/chromaui/action/issues/56:

Github Actions now support pull requests from forks: https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

More info about the `pull_request_target` event type: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target

However, the code explicitly prevents that type: https://github.com/chromaui/action/blob/c427f36903fb8c899e440bc90cb51822ef2959ed/src/main.ts#L25-L55

Here's an example run failure: https://github.com/Workday/canvas-kit/pull/788/checks?check_run_id=1114827451

`pull_request_target` should be a valid event.

**Note**: from our testing, adding `pull_request_target` or changing `pull_request` to `pull_request_target` in a Pull Request will result in Github Actions not running for that pull request. The event type is supposed to run code from the base branch (the branch the PR is being merged into) which gets Actions into a state where nothing triggers.

---

By @marboro92 in https://github.com/chromaui/action/issues/59:

My team would like to run chromatic checks only after a review is submitted, but currently the pull_request_review event (info on this event: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_review) is not supported by chromaui. 

Would it be possible to add support for pull_request_review?

tmeasday:
@ndelangen can you tell a bit more about what's happening in this ticket? Do you understand what these events mean/when they trigger?

ndelangen:
@tmeasday not yet, but I'll put this on my todo list to find out and assess if it's ok to just add these events 👍

ndelangen:
@ghengeveld can you do a release this week?

penx:
From the linked blog post

> However, instead of running against the workflow and code from the merge commit, the event runs against the workflow and code from the base of the pull request

From what I've seen, the Chromatic action will publish and compare the storybook from the base branch rather than with the proposed changes, resulting in no detected changes.

https://github.com/govuk-react/govuk-react/pull/1077#discussion_r862464284

Any idea if there's a way to use the pull_request_target but still get the visual diff to include the proposed changes?

penx:
Ah perhaps this blog post contains the answer - it is in relation to dependabot, but I think the same approach would work

https://blog.alphasmanifesto.com/2021/11/07/yarn-2-dependabot/


```yml
[...]

PR #204, opened 2020-11-05
ADD pull_request_review to list of allowed events

ndelangen:
issue: #196

- add the event to our list of allowed events


tmeasday (approved):
Lgtm

ndelangen:
really puzzled why `case 'pull_request_review':` doesn't match here 😕

PR #207, opened 2020-11-09
Throw error when running from shallow clone

### LINKED ISSUE #146: Show error message when there's no Git
**Why?** 
We get a lot of support issues about different git capabilities in CI environments. This often takes non-trivial back and forth to resolve. In some cases, it's [not obvious](https://app.intercom.com/a/apps/zj7sn9j1/inbox/inbox/all/conversations/27253866149). We should diagnose git issues and proactively recommend a solution _before_ they contact us. That helps users resolve issues faster by themselves.

**From recent memory:**
- Github actions fetch depth
- Jenkins configuration doesn’t expose git
- Docker images without git

With `git rev-list HEAD --count` we may have a heuristic for determining git capabilities of the users CI environment. If the user doesn't have git: 
- Show a CLI error message that tells them what's going on and how to resolve. (existing error messages [here](https://5d67dc0374b2e300209c41e7-axronkchqy.chromatic.com/?path=/story/cli-messages-errors--build-has-changes))
- Link to docs (we'll need to write these)

ghengeveld:
Fixes #146 and #179

This also extracts and improves other git warnings/errors.


domyen (approved):
lgtm. Left a comment in Chromatic.

PR #216, opened 2020-12-05
Use commit determined by env-ci

### LINKED ISSUE #215: Use commit from env-ci
We have a customer that uses `env-ci` & runs builds on PRs only (in addition to master). When that happens in their CI flow, the commit is not right. Rather than having to set both `CHROMATIC_SHA` and `CHROMATIC_BRANCH`, they are suggesting that we use the `commit` from `envCI` whenever we hit this code block:

https://github.com/chromaui/chromatic-cli/blob/ee1607225351856fd4920057e354efc0a5d8f295/bin/git/getCommitAndBranch.js#L79-L80

There is a complication with `env-ci` and setting `CHROMATIC_BRANCH` in the sense that the branch is determined via some conditional logic:

https://github.com/pvdlg/env-ci/blob/73299d7f729971a2725cff8b582367f108cfb5ea/services/vsts.js#L18

and they would rather not replicate that, which is fair. Is this a feasible option for us?

https://app.intercom.com/a/apps/zj7sn9j1/inbox/inbox/all/conversations/27253889238

travi:
fixes #215


tmeasday (approved):
LGTM. WDYT @ghengeveld ?

ndelangen (approved):

PR #218, opened 2020-12-10
Make sure isChromatic returns a boolean

kylesuss:
This came up in support. I think we have talked about making sure `isChromatic` returns a boolean before -- are there are any gotchas to be aware of?


dist/isChromatic.js
```diff
@@ -1,2 +1,2 @@
-"use strict";Object.defineProperty(exports,"__esModule",{value:!0}),exports["default"]=isChromatic;function isChromatic(){return window.navigator.userAgent.match(/Chromatic/)||window.location.href.match(/chromatic=true/)}
+"use strict";Object.defineProperty(exports,"__esModule",{value:!0}),exports["default"]=isChromatic;function isChromatic(){return!!(window.navigator.userAgent.match(/Chromatic/)||window.location.href.match(/chromatic=true/))}
```
kylesuss, line 1: Is this supposed to be checked into source control?

PR #231, opened 2021-01-02
Fix overriding Storybook version through environment variable

SasanFarrokh:
I think this was a mistake, i could not set CHROMATIC_STORYBOOK_VERSION


ghengeveld (approved):

ghengeveld:
Indeed this is broken, thanks for the fix!

ghengeveld:
Released in [v5.6.0](https://www.npmjs.com/package/chromatic).

PR #233, opened 2021-01-06
Add `--branch-name` flag to override branch name

ghengeveld:
This adds the `--branch-name` flag which skips branch detection in favor of whatever is provided. This is meant to be used only when using an unsupported CI, or when fixing a cross-fork UI comparison. Therefore I think we can omit it in the GitHub action.


ghengeveld:
Released in [v5.6.0](https://www.npmjs.com/package/chromatic).

PR #269, opened 2021-01-29
Record CI service name on build

ghengeveld:
Tracking which CI service was used enables us to better help a customer when they encounter issues.


tmeasday (approved):
LGTM

Issue #273, opened 2021-02-05
Github action get stuck for 10 minutes


hikkyu:
Hi there!

Chromatic step get stuck for 10 minutes. I think 10 minutes is the timeout.
I tried options `exitOnceUploaded: true | "true" ` & `exitZeroOnChanges` but nothing, build always during 10 minutes. Apparently this step only takes 30 secondes. Did I missed something ?

![Capture d’écran 2021-02-05 à 17 00 34](https://user-images.githubusercontent.com/467261/107057411-b8bbae00-67d3-11eb-8e60-b09f619e1537.png)

Here is my Github workflow file : 
```yaml
name: Chromatic

on: push

jobs:
  chromatic-deployment:
    name: Install & Check stories
    runs-on: ubuntu-latest
    steps:
      - name: Checkout repository
        uses: actions/checkout@v2
        with:
          fetch-depth: 0 # 👈 Required to retrieve git history
      - name: Cache node modules
        uses: actions/cache@v2
        env:
          cache-name: cache-node-modules
        with:
          # npm cache files are stored in `~/.npm` on Linux/macOS
          path: ~/.npm
          key: npm-${{ hashFiles('**/package-lock.json') }}
          restore-keys: |
            npm-
      - name: Install dependencies
        run: npm i
      - name: Publish to Chromatic
        uses: chromaui/action@v1
        with:
          token: ${{ secrets.GITHUB_TOKEN }}
          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
          exitOnceUploaded: "true"
```

jjenzz:
We're experiencing the same issue: 

```jsx
name: 'Chromatic'

on: pull_request

jobs:
  chromatic-deployment:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout repository
        uses: actions/checkout@v2
        with:
          fetch-depth: 0
      - name: Install dependencies
        run: yarn
      - name: Publish to Chromatic
        uses: chromaui/action@v1
        with:
          token: ${{ secrets.GITHUB_TOKEN }}
          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
          exitOnceUploaded: true
          exitZeroOnChanges: true

```

benoitgrelard:
We are seeing the same thing as well, it seems to have happen today, as it was faster before.

ccrowley96:
Same issue here!

Buuntu:
Same here.

[...]

PR #278, opened 2021-02-10
Fix/10minute timeout action

ndelangen:
issue: https://github.com/chromaui/chromatic-cli/issues/273


action/main.ts
```diff
@@ -148,19 +145,24 @@ async function run() {
...
+    setOutput('buildUrl', output.buildUrl);
+    setOutput('storybookUrl', output.storybookUrl);
+    setOutput('code', output.code.toString());
 
-    if (code !== 0) {
+    if (output.code !== 0) {
       setFailed('non-zero exit code');
     }
+
+    process.exit(output.code);
```
ndelangen, line 159: here's the fix

ndelangen: inspired from:
https://github.com/chromaui/chromatic-cli/blob/906f4426fc2641e78b556b1c1117d9b1e5301866/bin/main.js#L33-L36

ndelangen:
Looks like it's running in 2m consistency..

action/main.ts
```diff
@@ -148,19 +146,24 @@ async function run() {
       ignoreLastBuildOnBranch: maybe(ignoreLastBuildOnBranch),
     });
 
-    const [{ url, code }] = await Promise.all([chromatic]);
+    const output = await chromatic;
```
ghengeveld: We don't need the `chromatic` variable here, just do `await runChromatic` directly.

action/main.ts
```diff
@@ -148,19 +146,22 @@ async function run() {
...
     }
+
+    process.exit(output.code);
   } catch (e) {
     if (e.message) error(e.message);
     if (e.stack) error(e.stack);
     if (e.description) error(e.description);
     setFailed(e.message);
   }
+  process.exit(1);
```
ghengeveld, line 165: I think this should be inside the `catch` block.

ghengeveld (approved):
LGTM

Issue #286, opened 2021-02-22
Use new index APIs


tmeasday:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1199921881886840/1200015178949739) by [Unito](https://www.unito.io/learn-more)

tmeasday:
Released in v5.7.0.

PR #287, opened 2021-02-22
Update CLI to use new `test` terminology and statuses

tmeasday:
For #286

### DIFF
--- a/bin/tasks/report.js
+++ b/bin/tasks/report.js
@@ -11,7 +11,7 @@ const ReportQuery = `
     app {
       build(number: $buildNumber) {
         number
-        status
+        status(legacy: false)
         webUrl
         cachedUrl
         createdAt
--- a/bin/tasks/report.js
+++ b/bin/tasks/report.js
@@ -16,8 +16,9 @@ const ReportQuery = `
         cachedUrl
         createdAt
         completedAt
-        snapshots {
+        tests {
           status
+          result
           spec {
             name
             component {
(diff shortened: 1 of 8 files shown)


bin/tasks/report.js
```diff
@@ -61,30 +63,35 @@ export const generateReport = async (ctx) => {
...
-        break;
-      default:
+      default: {
+        switch (result) {
+          case 'SKIPPED':
+          case 'PRESERVED':
+            testCase.skipped();
+            break;
+          default:
+        }
```
ghengeveld, line 93: Seems a bit overkill to use a switch for something that could just be a single if statement.

bin/tasks/report.js
```diff
@@ -16,8 +16,9 @@ const ReportQuery = `
         cachedUrl
         createdAt
         completedAt
-        snapshots {
+        tests {
           status
+          result
```
ghengeveld, line 21: Did you add `result` to `testFields` in `authorizedBuildFieldsViaAppCode`?

ghengeveld (changes requested):

bin/tasks/report.js
tmeasday: Good point https://github.com/chromaui/chromatic/pull/4504

bin/tasks/report.js
```diff
@@ -61,30 +63,31 @@ export const generateReport = async (ctx) => {
...
+  build.tests.forEach(({ status, result, spec, parameters }) => {
     const suffix = parameters.viewportIsDefault ? '' : ` [${parameters.viewport}px]`;
     const testCase = suite
       .testCase()
       .className(spec.component.name.replace(/[|/]/g, '.')) // transform story path to class path
       .name(`${spec.name}${suffix}`);
 
     switch (status) {
-      case 'SNAPSHOT_ERROR':
+      case 'FAILED':
```
ghengeveld, line 74: You'll also need to change `status` to `status(legacy: false)` in the gql query above I think.

ghengeveld (changes requested):

tmeasday:
@ghengeveld I think this is ready now

ghengeveld:

PR #288, opened 2021-02-22
Upgrade to husky 5

tmeasday:
I had to do this in order to commit.

I ran `npx github:typicode/husky-4-to-5 --package-manager yarn`.


.husky/.gitignore
```diff
@@ -0,0 +1 @@
+_
```
ghengeveld, line 1: Do we really need this file? We can probably just ignore `.husky/_` from the root gitignore.

ghengeveld (approved):
I guess this is how it's supposed to work? Seems like a hassle to me.

.husky/.gitignore
tmeasday: IDK 🤷 I just ran their script

PR #295, opened 2021-03-02
Gracefully handle git config command in Netlify

ghengeveld:


bin/tasks/gitInfo.js
```diff
@@ -21,8 +21,10 @@ const TesterSkipBuildMutation = `
 export const setGitInfo = async (ctx, task) => {
   const { branchName, patchBaseRef, fromCI: ci } = ctx.options;
   ctx.git = await getCommitAndBranch({ branchName, patchBaseRef, ci, log: ctx.log });
-  ctx.git.slug = ctx.git.slug || (await getSlug());
   ctx.git.version = await getVersion();
+  if (!ctx.git.slug) {
+    ctx.git.slug = await getSlug().catch((e) => ctx.log.warn('Failed to retrieve slug', e));
+  }
```
ndelangen, line 27: What is this slug, and why is it optional? What do we miss if we don't have it?

bin/tasks/gitInfo.js
```diff
@@ -21,8 +21,10 @@ const TesterSkipBuildMutation = `
...
 
   if (ctx.options.ownerName) {
     ctx.git.slug = ctx.git.slug.replace(/[^/]+/, ctx.options.ownerName);
```
ndelangen, line 30: Will this code fail without the slug set?

ndelangen: referring to:
```
 ctx.git.slug = ctx.git.slug.replace(/[^/]+/, ctx.options.ownerName);
 ```

File: bin/tasks/gitInfo.test.js [hunk 3]
```diff
@@ -9,10 +9,10 @@ const log = { debug: jest.fn() };
 
 describe('setGitInfo', () => {
   it('sets the git info on context', async () => {
-    getCommitAndBranch.mockReturnValue({ commit: '123asdf', branch: 'something' });
-    getBaselineCommits.mockReturnValue(['asd2344']);
-    getVersion.mockReturnValue('Git v1.0.0');
-    getSlug.mockReturnValue('user/repo');
+    getCommitAndBranch.mockReturnValue(Promise.resolve({ commit: '123asdf', branch: 'something' }));
```
tmeasday, line 12: can you use `mockResolvedValue` for this?

tmeasday (commented):
I don't really understand what this change is doing; it doesn't seem covered by the test cases.

bin/tasks/gitInfo.js
tmeasday: @ghengeveld why not `try { } catch {}`?

I think it's confusing to use `await` and `.catch()`

ghengeveld: The slug is something like `chromaui/chromatic-cli` and is needed to determine whether we're dealing with a forked repository or not, which we need to know for cross-fork PR support.

I used `.catch` because it's way terser than try/catch and given this is such a benign thing I don't think it deserves to be so verbose.

bin/tasks/gitInfo.js
ghengeveld: Good catch.

bin/tasks/gitInfo.test.js
ghengeveld: Nice one 👍

ghengeveld:

PR #296, opened 2021-03-04
Support yarn execpath

amacneil:
I don't know why, but yarn was not being correctly detected on my system (maybe because I'm using yarn 2?).

In my case, `process.env.npm_execpath` was returning `/path/to/yarn` rather than `/path/to/yarn.js`.


ghengeveld (approved):

ghengeveld:
Thanks for that!

amacneil:
thanks!

PR #298, opened 2021-03-04
Document GitHub action outputs

### LINKED ISSUE #293: Document outputs and publish new version of action?
👋🏽 I just came across this recent change to add `storybookUrl` to the output of the Chromatic action: https://github.com/chromaui/chromatic-cli/commit/e5cf3c6b07bfeaa22f854aab12385aa606008cdd#diff-e0800b9a6c36cb7e1937f05f68784af42488be2266fad90cef27dc394b3b6b6fR152

I'm building a GitHub Actions step that needs access to the `storybookUrl` to work, so this is exactly what I've been looking for!

But, it looks like this change hasn't been published to the [action repo](https://github.com/chromaui/action) yet, which makes it really tricky to consume right now since actions require all dependencies to be bundled ahead of time.

It also looks like none of the currently supported `outputs` are documented in `action.yml`: https://github.com/chromaui/chromatic-cli/blob/next/action.yml

Would it be possible to:

1. document all supported outputs in `action.yml` (I'd be happy to submit a PR for this if it'd help)
1. publish a new version of the action?

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1199921881886840/1200015269989920) by [Unito](https://www.unito.io/learn-more)

mnquintana:
Closes #293 

This PR documents Chromatic's GitHub action's outputs, both in the Readme and in the `action.yml` manifest. It also adds [a missing required field in the GitHub action schema](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#inputsinput_idrequired) – `required`.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -75,13 +75,13 @@ inputs:
 
 outputs:
   url:
-    description: An alias for the build URL (e.g. https://www.chromatic.com/build?appId=<app id goes here>&number=<build number>)
+    description: 'An alias for the build URL (e.g. https://www.chromatic.com/build?appId=<app id goes here>&number=<build number>)'
   buildUrl:
-    description: The build URL (e.g. https://www.chromatic.com/build?appId=<app id goes here>&number=<build number>)
+    description: 'The build URL (e.g. https://www.chromatic.com/build?appId=<app id goes here>&number=<build number>)'
   storybookUrl:
-    description: The Storybook preview URL for your current branch / Pull Request (e.g. https://<app id goes here>-<branch hash>.chromatic.com/)
+    description: 'The Storybook preview URL for your current branch / Pull Request (e.g. https://<app id goes here>-<branch hash>.chromatic.com/)'
   code:
-    description: The exit code for the current run of the Chromatic CLI
+    description: 'The exit code for the current run of the Chromatic CLI'
 
 runs:
   main: action/register.js
--- a/action.yml
+++ b/action.yml
@@ -8,48 +8,70 @@ branding:
 inputs:
   token:
     description: 'Your github token'
+    required: true
   projectToken:
     description: 'Your chromatic project token'
+    required: true
   workingDir:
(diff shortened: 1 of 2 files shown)


action.yml
```diff
@@ -1,55 +1,87 @@
 name: Publish to Chromatic
 author: Chroma Software, Inc.
-description: "Publish your Storybook to Chromatic and run visual regression tests"
+description: 'Publish your Storybook to Chromatic and run visual regression tests'
```
mnquintana, line 3: Tiny side effect: prettier ran on `action.yml` when I saved, hence the double -> single quotes change. Let me know if you want me to change these back!

ghengeveld: I'm fine with switching to single quotes (although be aware there's a difference: `'\n'` is interpreted as a literal `\n` string while `"\n"` is interpreted as a newline. In this case, we don't care because it's all single lines.

I do prefer you add quotes to the output descriptions, for consistency.

ghengeveld (changes requested):

ghengeveld:
Thanks for contributing!

action.yml
mnquintana: @ghengeveld Fixed!

mnquintana:
👋🏽 Just checking in – anything I else I need to do to get this merge-ready?

ghengeveld (approved):

mnquintana:
@ghengeveld Awesome, thanks so much for giving this a look! Would you also be able to publish a new version of the action soon?

ghengeveld:

mnquintana:
Thanks a lot @ghengeveld! Would you also be able to publish a new release with the latest changes at https://github.com/chromaui/action? It looks like the last release was `v1` back in September. Not a huge deal because it's still usable by pointing my GitHub Actions config at the latest commit hash, but it would be nice to express it in terms of a semver range again.

PR #304, opened 2021-03-10
Support only testing components affected by recent git changes

ghengeveld:
This feature is currently in closed beta.

- [x] Add `--only-changed` CLI flag
- [x] Trace affected story files based on changed git files since baseline
- [x] Use previous baseline for re-runs
- [x] Bail if global/config/static files are affected
- [x] Skip build (early exit) on zero changed files
- [x] Skip build (mutation) on zero affected story files
- [x] Rename `baselineCommits` to `parentCommits`
- [x] For `--only-changed`, retrieve actual baseline commits to determine changed files


bin/lib/parseArgs.js
```diff
@@ -22,6 +22,7 @@ export default function parseArgs(argv) {
       --exit-zero-on-changes [branch]  If all snapshots render but there are visual changes, exit with code 0 rather than the usual exit code 1. Only for [branch], if specified. Globs are supported via picomatch.
       --ignore-last-build-on-branch <branch>  Do not use the last build on this branch as a baseline if it is no longer in history (i.e. branch was rebased). Globs are supported via picomatch.
       --only <storypath>  Only run a single story or a subset of stories. Story paths typically look like "Path/To/Story". Globs are supported via picomatch. This option implies --preserve-missing.
```
tmeasday, line 24:
```suggestion
      --only <storypath>  Only run a single story or a subset of stories. Story paths typically look like "Path/To/Story". Globs are supported via picomatch. All other snapshots will be inherited from the prior commit.
```

bin/lib/parseArgs.js
```diff
@@ -22,6 +22,7 @@ export default function parseArgs(argv) {
       --exit-zero-on-changes [branch]  If all snapshots render but there are visual changes, exit with code 0 rather than the usual exit code 1. Only for [branch], if specified. Globs are supported via picomatch.
       --ignore-last-build-on-branch <branch>  Do not use the last build on this branch as a baseline if it is no longer in history (i.e. branch was rebased). Globs are supported via picomatch.
       --only <storypath>  Only run a single story or a subset of stories. Story paths typically look like "Path/To/Story". Globs are supported via picomatch. This option implies --preserve-missing. 
+      --only-changed [branch]  Only run stories affected by files changed since the baseline build. Only for [branch], if specified. Globs are supported via picomatch.
```
tmeasday, line 25:
```suggestion
      --only-changed [branch]  Only run stories affected by files changed since the baseline build. Only for [branch], if specified. Globs are supported via picomatch. All other snapshots will be inherited from the prior commit.
```

bin/lib/getOptions.js
```diff
@@ -41,7 +42,7 @@ export default async function getOptions({ argv, env, flags, log, packageJson })
     exitZeroOnChanges: flags.exitZeroOnChanges === '' ? true : flags.exitZeroOnChanges,
     exitOnceUploaded: flags.exitOnceUploaded === '' ? true : flags.exitOnceUploaded,
     ignoreLastBuildOnBranch: flags.ignoreLastBuildOnBranch,
-    preserveMissingSpecs: flags.preserveMissing || !!flags.only,
+    preserveMissingSpecs: flags.preserveMissing || !!flags.only || flags.onlyChanged !== undefined,
```
tmeasday, line 45: This is now done automatically by the backend, `preserveMissingSpecs` actually has a slightly different effect (e.g. you can't remove stories any more!).

bin/tasks/gitInfo.js
```diff
@@ -72,11 +73,27 @@ export const setGitInfo = async (ctx, task) => {
...
+  // determine affected story files later.
+  // In the unlikely scenario that this list is empty (and not a rebuild), we can skip the build
+  // completely, since we know for certain it wouldn't have any effect.
+  if (baselineCommits.length && matchesBranch(ctx.options.onlyChanged)) {
+    const results = await Promise.all(baselineCommits.map((c) => getChangedFiles(c)));
+    ctx.git.changedFiles = [...new Set(results.flat())].map((f) => `./${f}`);
+    ctx.log.debug(`Found changedFiles:\n${ctx.git.changedFiles.map((f) => `  ${f}`).join('\n')}`);
+    if (ctx.git.changedFiles.length === 0) {
+      ctx.skip = true;
+      transitionTo(skippedRebuild, true)(ctx, task);
```
tmeasday, line 91: Should this transition to `skippedForCommit`? What does `skippedRebuild` do?

In this case you would want to mark the commit as successful, ie call the `skipBuild` mutation. Maybe it is better to let this fall out naturally from calling `getDependentStoryFiles` with an empty list of files?

scripts/trim-stats-file.js
```diff
@@ -0,0 +1,39 @@
...
+    JSON.stringify({ modules: trimmedModules }, null, 2)
+      .replace(/{\n {10}/g, '{ ')
+      .replace(/\n {8}}/g, ' }')
+  );
+
+  // eslint-disable-next-line no-console
+  console.log(`Wrote ${targetFile}`);
+};
+
+main();
```
tmeasday, line 39: What is the purpose of this script?

bin/tasks/build.js
```diff
@@ -36,6 +40,8 @@ export const setSpawnParams = (ctx) => {
       isYarn ? '' : '--',
       '--output-dir',
       ctx.sourceDir,
+      ctx.git.changedFiles && webpackStatsSupported && '--webpack-stats-json',
+      ctx.git.changedFiles && webpackStatsSupported && ctx.sourceDir,
```
tmeasday, line 44:
```suggestion
      ...(ctx.git.changedFiles && webpackStatsSupported ? ['--webpack-stats-json', ctx.sourceDir] : []),
```

bin/tasks/gitInfo.js
```diff
@@ -52,31 +60,61 @@ export const setGitInfo = async (ctx, task) => {
...
+  // In the unlikely scenario that this list is empty (and not a rebuild), we can skip the build
+  // since we know for certain it wouldn't have any effect. We do want to tag the commit.
+  if (parentCommits.length && matchesBranch(ctx.options.onlyChanged)) {
+    const baselineCommits = await getBaselineCommits(ctx, { branch, parentCommits });
+    ctx.log.debug(`Found baselineCommits: ${baselineCommits.join(', ')}`);
+
+    try {
+      const results = await Promise.all(baselineCommits.map((c) => getChangedFiles(c)));
+      ctx.git.changedFiles = [...new Set(results.flat())].map((f) => `./${f}`);
+      ctx.log.debug(`Found changedFiles:\n${ctx.git.changedFiles.map((f) => `  ${f}`).join('\n')}`);
```
tmeasday, line 99:
```suggestion
      // Find the union of all the changes since each baselineCommit.
      // If a file has been modified since one side of a merge but not the other, we don't know which
      // side contains the baseline, so we need to be conservative and assume it has changed since the baseline.
      const results = await Promise.all(baselineCommits.map((c) => getChangedFiles(c)));
      ctx.git.changedFiles = [...new Set(results.flat())].map((f) => `./${f}`);
      ctx.log.debug(`Found changedFiles:\n${ctx.git.changedFiles.map((f) => `  ${f}`).join('\n')}`);
```

tmeasday:
What's going on here? 
image

You added a bunch of stories in this PR right?

bin/tasks/gitInfo.js
ghengeveld: Technically, we could determine the changes for each commit individually. I don't think that's worthwhile though.
I've now implemented this such that we select the most recently committed baseline build as "the baseline" to determine the build status from. Or should we look at the build statuses instead?

bin/git/git.js
```diff
@@ -63,6 +63,21 @@ const TesterHasBuildsWithCommitsQuery = gql`
   }
 `;
 
+const TesterBaselineCommitsQuery = gql`
+  query TesterBaselineCommitsQuery($branch: String!, $parentCommits: [String!]!) {
+    app {
+      baselineBuilds(branch: $branch, parentCommits: $parentCommits) {
+        number
+        status
```
tmeasday, line 71: Needs `legacy:false`!!

ghengeveld: That is such a gotcha 🤦‍♂️

tmeasday: IKR. I'm not sure what we can do about it, beyond renaming the field entirely.

PR #305, opened 2021-03-11
Fix `npx chromatic` timing out on build-storybook

ghengeveld:
This uses `yarn-or-npm` to decide which client to use. It appears the original method doesn't work because I couldn't get it to use Yarn while running `yarn chromatic`. So rather than trying to detect the currently used client, this looks at the presence of yarn.lock (among other things) to determine the correct client.

Important detail: this means we'll be executing `npm run build-storybook` instead of `node /path/to/npm-cli.js run build-storybook`. It remains to be seen whether that will have an impact.

### DIFF
--- a/bin/tasks/build.js
+++ b/bin/tasks/build.js
@@ -28,6 +28,8 @@ export const setSpawnParams = (ctx) => {
   const npmExecPath = process.env.npm_execpath;
   const isJsPath = typeof npmExecPath === 'string' && /\.m?js/.test(path.extname(npmExecPath));
   const isYarn = npmExecPath && /^yarn(\.js)?$/.test(path.basename(npmExecPath));
+  ctx.log.debug('process.env.npm_execpath', process.env.npm_execpath);
+  ctx.log.debug('process.execPath', process.execPath);
   ctx.spawnParams = {
     client: isYarn ? 'yarn' : 'npm',
     platform: process.platform,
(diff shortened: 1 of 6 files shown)


ghengeveld:
Available as a pre-release: `chromatic@5.8.0-canary.0`

bin/tasks/build.js
```diff
@@ -21,22 +22,21 @@ export const setSourceDir = async (ctx) => {
...
-  //   npm/yarn run scriptName (depending on npm_execpath)
-  //   node path/to/npm.js run scriptName (if npm run via node)
-  // Based on https://github.com/mysticatea/npm-run-all/blob/52eaf86242ba408dedd015f53ca7ca368f25a026/lib/run-task.js#L156-L174
-  const npmExecPath = process.env.npm_execpath;
-  const isJsPath = typeof npmExecPath === 'string' && /\.m?js/.test(path.extname(npmExecPath));
-  const isYarn = npmExecPath && /^yarn(\.js)?$/.test(path.basename(npmExecPath));
   ctx.spawnParams = {
-    command: (isJsPath ? process.execPath : npmExecPath) || 'npm',
-    clientArgs: isJsPath ? [npmExecPath, 'run'] : ['run', '--silent'],
+    client: yarnOrNpm(),
```
ghengeveld, line 26: These are cached so it doesn't matter that we call their API multiple times.

tmeasday (approved):
LGTM. I believe the original code was copied from another tool, but if `yarn-or-npm` works I am good with that!

Issue #307, opened 2021-03-15
Consider adding support for `workflow_dispatch` event in GitHub Action


ghengeveld:
Requested [here](https://app.intercom.com/a/apps/zj7sn9j1/inbox/inbox/all/conversations/27253906975).

`workflow_dispatch` is a way to manually trigger a GitHub Action. See https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

This would need to be added here: https://github.com/chromaui/chromatic-cli/blob/62a5b6528c1cb8adecda4942f20003fc76b1edba/action/main.ts#L26-L52

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1199921881886840/1200607016213437) by [Unito](https://www.unito.io)

theKashey:
Yeah, we are generating only "changed" stories and need this escape hatch to be able run full tests when such comparison is not working by any reason.

PR #311, opened 2021-03-18
Support `workflow_dispatch` event in GitHub Action

lyleunderwood:
#307

I was trying to setup a manual chromatic runner using github actions via a github bot, and I wanted to use `workflow_dispatch` but this action didn't seem to allow it. I went ahead and added basic support.


lyleunderwood:
It seems like chromatic only ever wants just the branch name, never the full ref (`refs/heads/my-branch`). Providing the full ref seems to result in chromatic always failing to compare to master. Does anybody have any insight into this point? Is there any reason I'd ever want to allow the full ref to pass through?

ndelangen:
@lyleunderwood could you elaborate this question:

> Is there any reason I'd ever want to allow the full ref to pass through?

ghengeveld:
@lyleunderwood That's right, we basically never expect to see `refs/heads/`. We strip that off [right here](https://github.com/chromaui/chromatic-cli/blob/701f7c8244b419a8b80b5a33a12392c8ea0ef351/bin/git/git.js#L117). Arguably the way we do it is slightly safer because the regex ensures it only strips it at the start of the string, never somewhere half way. It's unlikely someone would have an actual branch name like that, but theoretically possible.

@ndelangen Can you judge whether workflow_dispatch is safe for us to allow?

ndelangen:
> Can you judge whether workflow_dispatch is safe for us to allow? If so, I think this is a nice improvement.

Shall we add a test workflow in this repo and test it.

lyleunderwood:
Hey, sorry I've been on vacation :tropical_drink: 

I went ahead and updated the README to reflect that we really just want a regular branch name.

As for stripping off the ref prefix, it's currently just based on what's being done in the `push` case, let me know if I should remove it.

https://github.com/chromaui/chromatic-cli/blob/71e3c9639e735699d7fe6ebbd89a4bff10ecf747/action/main.ts#L42

Manubi:
Will this get merged any time soon? :)

ndelangen (approved):

PR #313, opened 2021-03-25
Use original baseline for rebuilds (new build for the same commit)

ghengeveld:
If the last build was on the same branch and for the same commit, we're dealing with a rebuild (e.g. restarted CI job). In case this build was passed/accepted, we can skip the build altogether, because it shouldn't lead to any changes. This of course assumes flaky stories won't get accepted.


tmeasday:
@ghengeveld I thought about this a bit more and I realised we had neglected to think about the more common situation where this happens. 

So we've been thinking about people re-running CI jobs due to flakes.

However, when you have Chromatic setup without provider integration, you use `chromatic` without the `--exit-zero-on-changes` flag, and the process for approving snapshots looks like so:

1. Push a commit with changes
2. Chromatic runs, and fails CI due to those changes. You get a red PR.
3. You go into Chromatic and approve those changes. Nothing happens on the PR as you aren't integrated.
4. You re-run the CI job, which this time passes as the new baselines are from the same commit that you are running.

So with that in mind, two things:

1. This change is going to mess things up for folks doing the above, with or without `--onlyChanged`, because the baseline in step 4 won't be updated to the build from step 2.
2. If someone is using `--onlyChanged`, without this change it's quite cool actually -- we use the build from the same commit as the most recent ancestor in step 4, which, given those tests have been approved (and are the baselines) means we preserve every story and set it automatically to `PASSED`, which means the build passes without any captures taken. Which is correct!

---

So I think probably the correct solution here is a "recapture" button in the UI that is smart enough to just rerun the captures in the current build (not the preserved stories). WDYT?

ghengeveld:
@tmeasday I updated this according to what we discussed.

tmeasday (approved):
LGTM, thanks @ghengeveld

PR #344, opened 2021-05-18
Fix determining viewLayer when using transitive dependency

### LINKED ISSUE #342: Could not find a supported Storybook viewlayer as transitive dependency
When using e.g. `@nuxtjs/storybook`, which has a [dependency](https://github.com/nuxt-community/storybook/blob/8306f371d4992c86f64da726c4c45a57daa64363/package.json#L29) on `@storybook/vue`, we fail to detect the viewlayer package.

Previously we would look in `node_modules` to detect the viewlayer, which would work for transitive dependencies. This was changed in #319 to look at `dependencies` in package.json directly.

ghengeveld:
Fixes #342 

This also drops `addons` reporting because we aren't using it and it's not reliable anyway.


bin/lib/getStorybookInfo.js
```diff
@@ -23,59 +20,36 @@ const viewLayers = {
...
-  '@storybook/addon-queryparams': 'queryparams',
-  '@storybook/addon-storyshots': 'storyshots',
-  '@storybook/addon-storysource': 'storysource',
-  '@storybook/addon-viewport': 'viewport',
-};
+// Double inversion on Promise.all means fulfilling with the first fulfilled promise, or rejecting
+// when _everything_ rejects. This is different from Promise.race, which immediately rejects on the
+// first rejection.
+const invert = (promise) => new Promise((resolve, reject) => promise.then(reject, resolve));
+const raceFulfilled = (promises) => invert(Promise.all(promises.map(invert)).then((arr) => arr[0]));
```
tmeasday, line 29: Lol neat trick

tmeasday (changes requested):
This LGTM apart from dropping addons -- we do use the fact that you are using `docs` or `essentials` to decide whether to show the docs tab on the component screen.

We'll need to change that in Chromatic before we can drop it.

bin/lib/getStorybookInfo.js
ghengeveld: It was nice to flex my async muscles again.

ghengeveld:
> This LGTM apart from dropping addons -- we do use the fact that you are using `docs` or `essentials` to decide whether to show the docs tab on the component screen.
> 
> We'll need to change that in Chromatic before we can drop it.

Right, I wasn't aware of that. Reverted.

PR #366, opened 2021-06-15
Support resolving webpack stats in subdirectory

ghengeveld:
This should fix `--only-changed` for monorepos (or any other case where Chromatic is executed from a subdirectory of the git repo).

### DIFF
--- a/bin/lib/getDependentStoryFiles.js
+++ b/bin/lib/getDependentStoryFiles.js
@@ -1,3 +1,4 @@
+import { getWorkingDir } from '../git/git';
 import bailFile from '../ui/messages/warnings/bailFile';
 
 // Bail whenever one of these was changed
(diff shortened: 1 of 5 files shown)


bin/lib/getDependentStoryFiles.js
```diff
@@ -68,7 +71,10 @@ export function getDependentStoryFiles(ctx, stats, changedFiles) {
     }
   }
 
-  changedFiles.forEach(traceName);
+  changedFiles.forEach((gitFilePath) => {
+    const webpackFilePath = workingDir ? gitFilePath.replace(workingDirRegExp, '.') : gitFilePath;
```
tmeasday, line 75:
```suggestion
    const webpackFilePath = workingDir ? path.relative(workingDir, gitFilePath) : gitFilePath;
```

tmeasday (approved):
LGTM with one small improvement

bin/lib/getDependentStoryFiles.js
ghengeveld: Resolved this differently.

PR #370, opened 2021-06-24
Ignore `--only-changed` on rebuild

ghengeveld:


tmeasday (approved):
Thanks @ghengeveld. Can you add it to the docs as a reason why we might do a full build?

ghengeveld:
> Thanks @ghengeveld. Can you add it to the docs as a reason why we might do a full build?

Done.

PR #374, opened 2021-06-28
Fix handling of NO_PROXY environment variable

### LINKED ISSUE #359: Implementation of no_proxy handling is too simplistic
You added proxy handling and at the same time skip it, if the `no_proxy` environment variable is set:
https://github.com/chromaui/chromatic-cli/pull/347/commits/f065a84ca4f21b1915900cc8029d7486ef621ec0#diff-2225d8419a1f54b9c64600dfe5a91c3ad56d1d0c213aedb49476abe623aa0edaR9

Unfortunately having `no_proxy` does not always mean, that no proxy should be used.
In our situation it contains the company internal domain suffix, for which no proxy should be used.
Making calls to hosts outside our company network still requires using a proxy though.

Cf. https://curl.se/libcurl/c/CURLOPT_NOPROXY.html

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1199921881886840/1200606966446678) by [Unito](https://www.unito.io)

ghengeveld:
Fixes #359

This removes the global proxy agent, instead sets up a separate proxy agent for each hostname, and skips it if the request URL matches one of the hostnames specified in the NO_PROXY environment variable.

This also updates all `fetch` calls to use `HTTPClient`, which has the proxy agent configured.


tmeasday (commented):
Would it make sense to use a common library that wraps fetch rather than doing this everywhere? Seems like a lot of redundancy constantly having to pass the URL to `fetch` and the `getProxyAgent` function.

Maybe the `HTTPClient` is the right place? Why do we even have that lib if we are using raw `fetch` all over the place?

ghengeveld:
> Would it make sense to use a common library that wraps fetch rather than doing this everywhere? Seems like a lot of redundancy constantly having to pass the URL to `fetch` and the `getProxyAgent` function.
> 
> Maybe the `HTTPClient` is the right place? Why do we even have that lib if we are using raw `fetch` all over the place?

Done, we're using HTTPClient everywhere now.

PR #391, opened 2021-08-25
minor improvements

ndelangen:
- fix husky
- fix a .env example file

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -6,5 +6,5 @@
 
 CHROMATIC_PROJECT_TOKEN=<place app_code here>
 
-# CHROMATIC_INDEX_URL=https://www.chromaticqa.com
-# CHROMATIC_TUNNEL_URL=http://tunnel.chromaticqa.com
+# CHROMATIC_INDEX_URL=https://www.chromatic.com
+# CHROMATIC_TUNNEL_URL=http://tunnel.chromatic.com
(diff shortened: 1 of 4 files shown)


.env.example
```diff
@@ -6,5 +6,5 @@
 
 CHROMATIC_PROJECT_TOKEN=<place app_code here>
 
-# CHROMATIC_INDEX_URL=https://www.chromaticqa.com
-# CHROMATIC_TUNNEL_URL=http://tunnel.chromaticqa.com
+# CHROMATIC_INDEX_URL=https://www.chromatic.com
+# CHROMATIC_TUNNEL_URL=http://tunnel.chromatic.com
```
ghengeveld, line 10: The tunnel still uses `tunnel.chromaticqa.com` as `tunnel.chromatic.com` doesn't actually exist.

bin/lib/getStorybookInfo.js
```diff
@@ -122,7 +122,7 @@ const findViewlayer = async ({ env, log, options, packageJson }) => {
       resolvePackageJson(pkg)
         .then((json) => ({ viewLayer, version: json.version }))
         .catch(() => Promise.reject(new Error(noViewLayerPackage(pkg)))),
-      timeout(10000),
+      timeout(9000),
```
ghengeveld, line 125: Lol what's up with that one second? Fine with me though.

ghengeveld (approved):
LGTM

bin/lib/getStorybookInfo.js
ndelangen: 10000 is exactly the CI timeout lol, I lowered it, but then found I should really just force the timers to end, will revert these changes.

Issue #166, opened 2020-07-23
Yarn Berry Support


jayarnielsen:
The Chromatic CLI seems to have issues running in a Yarn v2 project. Whenever we run `yarn chromatic` we get this error:

```
UnhandledPromiseRejectionWarning: TypeError: loggly.createClient is not a function
```

Wondering if this is a known issue with Yarn 2 or something else.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1199921881886840/1199953301129733) by [Unito](https://www.unito.io)

tmeasday:
That seems odd, I think we are using https://github.com/loggly/node-loggly-bulk in a fairly straightforward and still the currently recommended way.

I don't know too much about Yarn Berry, do have any idea about how to debug it?

kylesuss:
I had another report of this come through on support today. Yarn v2 user as well.

ghengeveld:
I suspect this is indeed broken, but only gets surfaced by Yarn 2. We should handle rejected promises globally as a catch-all:

```
process.on('unhandledRejection', error => ...);
```

ghengeveld:
I've released an alpha version which adds a promise rejection handler and replaces the default `loggly` import with a named `createClient` import. Hopefully that'll resolve the issue. @jayarnielsen could you verify? You can install it via `yarn add chromatic@alpha` or just use `npx chromatic@alpha`.

jayarnielsen:
Hi @ghengeveld 👋 Thanks for getting back to me so quickly. I'm now getting this error `Unhandled promise rejection: TypeError: createClient is not a function` on chromatic 5.1.0-alpha.3

ghengeveld:
Hi @jayarnielsen, I was able to reproduce and fix the issue with Yarn 2. I've published another alpha which should resolve the issue. Could you verify? Install with `yarn add chromatic@alpha` or run with `npx chromatic@alpha`. Thanks!

ghengeveld:
Probably you're going to run into "Could not find a supported Storybook viewlayer package" now. I'm looking into it.

ghengeveld:
@jayarnielsen I've tried getting Yarn 2 to work for a couple hours but it's currently failing at build-storybook which fails with `Module not found: Error: Can't resolve 'babel-loader'`

I'm out of ideas right now. I did find this supposed workaround, which might work:
https://github.com/yarnpkg/berry/blob/master/packages/plugin-node-modules/README.md

[...]

PR #393, opened 2021-08-25
Bundle code into a single file without dependencies

ndelangen:
- Bundle the chromatic-cli bin and action, so they are a single file without dependencies! 
  No more dependencies need to be installed!
  This should increase 
- Add proper support for yarn berry with pnp enabled
   by asking the yarn cli for storybook packages,
   instead of trying to resolve them all.
   (use the resolving mechanism is asking the yarn cli fails)

This fixes: https://github.com/chromaui/chromatic-cli/issues/166

- Add case handling `workflow_run` for our action

This fixes: https://github.com/chromaui/chromatic-cli/issues/396

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 6.0.0 - unreleased
 
-- [393](https://github.com/chromaui/chromatic-cli/pull/393) Refactor to ESM & Bundle the bin so it's dependency-less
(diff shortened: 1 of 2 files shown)


ndelangen:
I'm, testing this across a few repositories..

ndelangen:
looking good now:
https://github.com/chromaui/chromatic/pull/5475

ndelangen:
I took out the breaking change, so it should be backwards compatible, I tested this on the chromatic repo

PR #411, opened 2021-09-23
Ask yarn for packages instead of trying to resolve them

ndelangen:
Instead of iterating over a list of packages to try and require.resolve I run a single `yarn list --pattern @storybook/` command and extract the package names & version from that.

This means we do not need packages to be installed locally, helping with Yarn 3 compatibility: https://github.com/chromaui/chromatic-cli/issues/166


ndelangen:
tested this here:
https://github.com/chromaui/chromatic/pull/5475

PR #447, opened 2021-11-09
Tech/allow many args CH-789

ndelangen:
https://linear.app/chromaui/issue/CH-789/chromatic-cli-should-accept-multiples-of-the-same-argument-and-pick

@zol I think this should fix the issue..


bin-src/lib/getOptions.js
```diff
@@ -50,7 +50,7 @@ export default async function getOptions({ argv, env, flags, log, packageJson })
     originalArgv: argv,
 
     buildScriptName: flags.buildScriptName,
-    outputDir: flags.outputDir,
+    outputDir: Array.isArray(flags.outputDir) ? takeLast(flags.outputDir) : undefined,
```
tmeasday, line 53: Doesn't `takeLast` already include this ternary?

tmeasday (changes requested):
Small tweak maybe?

bin-src/lib/getOptions.js
ndelangen: The value should always be an array of strings (thanks to `isMultiple: true`), but if it isn't -somehow- I want to return `undefined`.

tmeasday: image

Are you concerned about the possibility of the flag not being an array but also not being `undefined`? Is that actually possible with the library? Would that concern not apply to other options we are `takeLast`-ing? (In which case should `takeLast` just return `undefined` in the non-array case?)

zol:
Progress here looks good. Sadly I don't have the time/expertise to do code reviews anymore so I've taken myself off the list here.

ghengeveld:
@ndelangen What's the status of this? Perhaps you'll want to revisit something based on @tmeasday's comment?

ndelangen:
Very strange for the tests to start failing all of a sudden after merging in `main`? @ghengeveld got any ideas?

ghengeveld:
Sorry, my bad.

ghengeveld:
I don't think this PR is going to completely address the problem. For example, in [this](https://app.intercom.com/a/apps/zj7sn9j1/inbox/inbox/conversation/27253970011#part_id=comment-27253970011-11751369752) example, the `--webpack-stats-json` flag is getting passed to `build-storybook` twice, and getting rejected as a result. We could tweak the Storybook CLI to deal with that, but really Chromatic should prevent the same flag from getting set twice when it invokes `build-storybook`. I think we should resolve [`buildScriptName`](https://github.com/chromaui/chromatic-cli/blob/3e4c4d3d8b59fd5db61df96b800d610cb0b03b72/bin-src/tasks/build.js#L57) to the underlying command and strip `--output-dir` and `--webpack-stats-json` (and their alias / camelCase form) from that command before invoking it.

ndelangen:
I see, this PR is still an improvement I guess, a step into the right direction?

ndelangen:
@ghengeveld could you add your review on this?

bin-src/main.js
```diff
@@ -70,7 +70,7 @@ export async function runBuild(ctx) {
     ctx.options = await getOptions(ctx);
   } catch (e) {
     ctx.log.info('');
-    ctx.log.error(e.message);
+    ctx.log.error(e.stack || e.message);
```
ghengeveld, line 73: This should be wrapped in `fatalError` like below. Just throwing a stack trace at the user isn't good UX.

ghengeveld (approved):

PR #455, opened 2021-11-26
Add `--storybook-base-dir` to support TurboSnap with a prebuilt Storybook originating from a subdirectory

### LINKED ISSUE #403: Incompatible options: --storybook-build-dir (-d), --only-changed
We get this error from the Github action.

Sounds like a bug to me, why would those options be incompatible?

ghengeveld:
Fixes #403

This allows TurboSnap (via `--only-changed`) to be used with a prebuilt Storybook (via `--storybook-build-dir` or `-d`).

In case the prebuilt Storybook does not originate from the same location where Chromatic runs, this would result in zero changed stories getting detected, thus running no tests at all. This PR adds `--storybook-base-dir` to correct for that.

TODO: Show a warning when we detect this discrepancy (using a heuristic) and suggest to use `--storybook-base-dir`.


tmeasday (approved):
OK, I think makes sense!

ghengeveld:
This was released in [6.1.0](https://github.com/chromaui/chromatic-cli/blob/main/CHANGELOG.md#610---2021-11-29).

PR #458, opened 2021-11-26
Track `bailReason`, improve TurboSnap messaging and throw on 0 CSF globs

### LINKED ISSUE #451: Upload a "reason" code if TurboSnap bails
Alongside the build, we should upload some kind of enum detailing *why* we bailed out and captured all snapshots, if we are doing so. Known reasons off the top of my head:

 - Changes to "package" files (i.e `package.json` etc)
 - Changes to Storybook config files or their dependencies.
 - Changes to defined "externals".

### LINKED ISSUE #467: Throw error when 0 CSF globs are detected when using `--only-changed`
If we can't find any CSF globs in the stats file, the Storybook must be misconfigured and that needs to be obvious. If we find 0 CSF globs, Chromatic will break.

ghengeveld:
Fixes #451
Fixes #467
Fixes CH-1206

⚠️ Must be released after deploying https://github.com/chromaui/chromatic/pull/5688.
⚠️ Chromatic builds for this PR will fail until ☝️ is deployed because of API schema changes.

This PR:
- Records the TurboSnap status and "bail reason" and sends them along with the `createBuild` mutation.
- Throws an error when no CSF globs are found in the stats file.
- Improves and extends a log of messaging / logs around TurboSnap, including the number of actual and skipped captures, just like in the webapp.
- Shows message when build has no ancestors (with & without TurboSnap).
- Reveals GraphQL errors if there are any.

Screenshot 2021-12-09 at 08 43 11
Screenshot 2021-12-09 at 08 45 55
Screenshot 2021-12-08 at 17 50 07
Screenshot 2021-12-09 at 13 33 27


bin-src/tasks/upload.js
```diff
@@ -133,16 +135,21 @@ export const traceChangedFiles = async (ctx, task) => {
...
-        .join('\n')}`
-    );
-    transitionTo(traced)(ctx, task);
+    const onlyStoryFiles = await getDependentStoryFiles(ctx, stats, statsPath, changedFiles);
+    if (onlyStoryFiles) {
+      ctx.onlyStoryFiles = onlyStoryFiles;
+      ctx.log.debug(
+        `Found affected story files:\n${Object.entries(onlyStoryFiles)
+          .map(([id, f]) => `  ${f} [${id}]`)
+          .join('\n')}`
```
tmeasday, line 144: Again, I don't think this should be debug, at least not in CI.

bin-src/tasks/verify.js
```diff
@@ -85,11 +88,16 @@ export const createBuild = async (ctx, task) => {
...
   }
 
-  const { createBuild: build } = await client.runQuery(TesterCreateBuildMutation, {
+  const { createBuild: build } = await ctx.client.runQuery(TesterCreateBuildMutation, {
     input: {
       ...commitInfo,
+      rebuildForBuildId,
       ...(only && { only }),
       ...(onlyStoryFiles && { onlyStoryFiles: Object.keys(onlyStoryFiles) }),
+      ...(turboSnap && { turboSnapEnabled: !turboSnap.bailReason }),
```
tmeasday, line 97: Are we calling this `onlyStoryFiles` or `turboSnap` on the backend? Isn't it kind of confusing to use both?

tmeasday (approved):
Love it! Couple of small nits

bin-src/tasks/verify.js
ghengeveld: Well, I tried to avoid calling it `turboSnap` in the CLI for a while, but in the end that didn't seem right. Adding `bailReason` seems too generic (bail on what?) for an API, so I decided to name it what it is.

Initially, `turboSnapEnabled` would just be an indicator of whether `--only-changed` was set, regardless of bailing (i.e. they attempted to use TS). I then decided this was confusing and `turboSnapEnabled` should just be true when it's actually going through with it (not bailing).

I guess we can technically do without the `turboSnapEnabled` field and just infer it from the presence of `onlyStoryFiles` and omission of `turboSnapBailReason`. WDYT?

tmeasday: I get it, thanks @ghengeveld

ghengeveld:
Released as canary: [`6.3.0-canary.0`](https://www.npmjs.com/package/chromatic)

ghengeveld:
This was released in 6.3.0.

PR #459, opened 2021-11-29
Support zip file upload CH-816 CH-817

codykaup:
This PR adds support for uploading a zip of all the Storybook artifacts which should speed up the publish step in most cases! 🎉

### DIFF
--- a/bin-src/lib/compress.js
+++ b/bin-src/lib/compress.js
@@ -1,0 +1,22 @@
+import archiver from 'archiver';
+import fs from 'fs-extra';
+import { join } from 'path';
+import { file as tempFile } from 'tmp-promise';
+
+export default async function makeZipFile(ctx) {
+  const archive = archiver('zip', { zlib: { level: 9 } });
+  const tmp = await tempFile({ postfix: '.zip' });
+  const sink = fs.createWriteStream(null, { fd: tmp.fd });
+  const { paths } = ctx.fileInfo;
+
+  return new Promise((resolve, _) => {
+    sink.on('close', () => {
+      resolve({ path: tmp.path, size: archive.pointer() });
+    });
+    archive.pipe(sink);
+    paths.forEach((path) => {
+      archive.append(fs.createReadStream(join(ctx.sourceDir, path)), { name: path });
+    });
+    archive.finalize();
+  });
+}
--- a/bin-src/lib/compress.js
+++ b/bin-src/lib/compress.js
@@ -9,14 +9,27 @@ export default async function makeZipFile(ctx) {
   const sink = fs.createWriteStream(null, { fd: tmp.fd });
   const { paths } = ctx.fileInfo;
 
-  return new Promise((resolve, _) => {
+  return new Promise((resolve, reject) => {
     sink.on('close', () => {
       resolve({ path: tmp.path, size: archive.pointer() });
     });
-    archive.pipe(sink);
-    paths.forEach((path) => {
-      archive.append(fs.createReadStream(join(ctx.sourceDir, path)), { name: path });
+
+    // 'warning' messages contain non-blocking errors
+    archive.on('warning', (err) => {
+      ctx.log.debug({ err }, 'Received warning when creating zip file');
     });
+    archive.on('error', (err) => {
+      reject(err);
+    });
+    archive.pipe(sink);
+
+    paths.forEach((path) => {
+      const fullPath = join(ctx.sourceDir, path);
+      ctx.log.debug({ fullPath }, 'Adding file to zip archive');
+      archive.append(fs.createReadStream(fullPath), { name: path });
+    });
+
+    ctx.log.debug('Finalizing zip archive');
     archive.finalize();
   });
 }
--- a/bin-src/lib/parseArgs.js
+++ b/bin-src/lib/parseArgs.js
@@ -29,6 +29,7 @@ export default function parseArgs(argv) {
       --preserve-missing  Treat missing stories as unchanged rather than deleted when comparing to the baseline.
       --skip [branch]  Skip Chromatic tests, but mark the commit as passing. Avoids blocking PRs due to required merge checks. Only for [branch], if specified. Globs are supported via picomatch.
       --storybook-base-dir <dirname>  Relative path from repository root to Storybook project root. Use with --only-changed and --storybook-build-dir when your Storybook is located in a subdirectory of your repository.
+      --zip  Attempt to upload Storybook to Chromatic as a single zip file instead of individual content files.
 
     Debug options
       --ci  Mark this build as a CI build. Alternatively, set the CI environment variable (present in most CI systems). This option implies --no-interactive.
@@ -66,6 +67,7 @@ export default function parseArgs(argv) {
         preserveMissing: { type: 'boolean' },
         skip: { type: 'string' },
         storybookBaseDir: { type: 'string' },
+        zip: { type: 'boolean' },
 
         // Debug options
         ci: { type: 'boolean' },
--- a/bin-src/lib/parseArgs.js
+++ b/bin-src/lib/parseArgs.js
@@ -29,7 +29,7 @@ export default function parseArgs(argv) {
       --preserve-missing  Treat missing stories as unchanged rather than deleted when comparing to the baseline.
       --skip [branch]  Skip Chromatic tests, but mark the commit as passing. Avoids blocking PRs due to required merge checks. Only for [branch], if specified. Globs are supported via picomatch.
       --storybook-base-dir <dirname>  Relative path from repository root to Storybook project root. Use with --only-changed and --storybook-build-dir when your Storybook is located in a subdirectory of your repository.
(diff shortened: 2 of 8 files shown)


bin-src/lib/uploadZip.js
```diff
@@ -0,0 +1,77 @@
...
+
+      const res = await ctx.http.fetch(
+        url,
+        {
+          method: 'PUT',
+          body: fs.createReadStream(path).pipe(progressStream),
+          headers: {
+            'content-type': 'application/zip',
+            'content-length': contentLength,
+            'cache-control': 'max-age=31536000',
```
jmhobbs, line 29: I'm sure I just copy-pasted, but I don't think `cache-control` is meaningful on a `PUT` here.

codykaup: Good catch!

bin-src/tasks/upload.js
```diff
@@ -166,6 +175,51 @@ export const uploadStorybook = async (ctx, task) => {
...
+  if (ctx.skip) return;
+  transitionTo(preparing)(ctx, task);
+
+  if (ctx.options.zip) {
+    try {
+      await uploadAsZipFile(ctx, task);
+    } catch (err) {
+      ctx.log.debug('error uploading zip file', err);
+      await uploadAsIndividualFiles(ctx, task);
+    }
```
brockwood, line 218: If someone passes the `--zip` flag in and it fails with the fallback to the individual uploads, is there any indication to the end user outside of running with the debug flag? Not sure if developers would care so that is probably overkill so asking out of curiosity.

brockwood (approved):
LGTM... the question posed is merely for my own information.

bin-src/tasks/upload.js
jmhobbs: The only indication should be a resetting of the upload progress bar, unless `--debug` is passed.

bin-src/tasks/upload.js
```diff
@@ -166,6 +175,51 @@ export const uploadStorybook = async (ctx, task) => {
...
+
+export const uploadStorybook = async (ctx, task) => {
+  if (ctx.skip) return;
+  transitionTo(preparing)(ctx, task);
+
+  if (ctx.options.zip) {
+    try {
+      await uploadAsZipFile(ctx, task);
+    } catch (err) {
+      ctx.log.debug('error uploading zip file', err);
```
ghengeveld, line 216: Our usual pattern for logs is to have an object with `err` property (and potentially other stuff) first, followed by an optional string message. @tmeasday can probably explain why.

```suggestion
      ctx.log.debug({ err }, 'Error uploading zip file');
```

 Also, sentences start with a capital letter. But that's just my OCD.

bin-src/lib/uploadZip.js
```diff
@@ -0,0 +1,76 @@
...
+    async (bail) => {
+      let res;
+      try {
+        res = await ctx.http.fetch(url, {}, { retries: 0, noLogErrorBody: true });
+      } catch (e) {
+        throw new Error('Sentinel file not present.');
+      }
+
+      const result = await res.text();
+      if (SENTINEL_SUCCESS_VALUE !== result) {
```
ghengeveld, line 64: Lol this isn't Java where the constant has to go on the left to avoid NPEs, but I'm down with it.

bin-src/lib/compress.js
```diff
@@ -0,0 +1,22 @@
...
+import { join } from 'path';
+import { file as tempFile } from 'tmp-promise';
+
+export default async function makeZipFile(ctx) {
+  const archive = archiver('zip', { zlib: { level: 9 } });
+  const tmp = await tempFile({ postfix: '.zip' });
+  const sink = fs.createWriteStream(null, { fd: tmp.fd });
+  const { paths } = ctx.fileInfo;
+
+  return new Promise((resolve, _) => {
```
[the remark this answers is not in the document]
ghengeveld: The [docs](https://github.com/archiverjs/node-archiver#quick-start) suggest handing `archive.on('warning', ...)` and `archive.on('error', ...)`

bin-src/lib/compress.js
```diff
@@ -0,0 +1,22 @@
...
+    sink.on('close', () => {
+      resolve({ path: tmp.path, size: archive.pointer() });
+    });
+    archive.pipe(sink);
+    paths.forEach((path) => {
+      archive.append(fs.createReadStream(join(ctx.sourceDir, path)), { name: path });
+    });
+    archive.finalize();
+  });
+}
```
ghengeveld, line 22: I'm missing unit tests for this file.

ghengeveld (changes requested):
Good stuff. I assume you tested the functionality of the progress indicator (i.e. does it actually progress gradually).

bin-src/lib/compress.js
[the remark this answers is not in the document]
ghengeveld: Would also be good to have some debug logging in here.

bin-src/lib/parseArgs.js
```diff
@@ -29,6 +29,7 @@ export default function parseArgs(argv) {
       --preserve-missing  Treat missing stories as unchanged rather than deleted when comparing to the baseline.
       --skip [branch]  Skip Chromatic tests, but mark the commit as passing. Avoids blocking PRs due to required merge checks. Only for [branch], if specified. Globs are supported via picomatch.
       --storybook-base-dir <dirname>  Relative path from repository root to Storybook project root. Use with --only-changed and --storybook-build-dir when your Storybook is located in a subdirectory of your repository.
+      --zip  Attempt to upload Storybook to Chromatic as a single zip file instead of individual content files.
```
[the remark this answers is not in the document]
ghengeveld: Don't forget to add the flag to the [docs](https://github.com/chromaui/chromatic-docs/blob/main/cli.md) as well.

bin-src/lib/uploadZip.js
```diff
@@ -0,0 +1,76 @@
+import retry from 'async-retry';
+import fs from 'fs-extra';
+import progress from 'progress-stream';
+
+const SENTINEL_SUCCESS_VALUE = 'OK';
```
ghengeveld, line 5: Perhaps add a note on where this comes from, and what "sentinel" means.

bin-src/lib/parseArgs.js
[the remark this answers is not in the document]
codykaup: Good point, I'll get that updated!

I was planning on getting this PR merged then going over to update the docs but thanks for the heads up!

bin-src/lib/compress.js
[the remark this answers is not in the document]
codykaup: Good callout, I've added those catches. I figured `warning` could get logged since they're non-blocking but I'm opening to throwing an error in those cases.

bin-src/lib/compress.js
codykaup: Just added some tests for this. Heads up, I added an in-memory file system for testing purposes which brought in another dev dependency.

I also wasn't sure what causes `warning` signals since `ENOENT` errors show up in the `error` signals though. ¯\\\_(ツ)\_/¯

codykaup:
> I assume you tested the functionality of the progress indicator (i.e. does it actually progress gradually).

Yep! The progress bar functions as expected with one exception... since we're unpacking externally, it may sit at 100% during the publish step but it's typically fast enough to not a be an issue.

ghengeveld (approved):
Nice work. Is the sentinel file part of a known pattern? I tried googling it but came up empty.

bin-src/lib/uploadZip.js
```diff
@@ -0,0 +1,79 @@
...
+}
+
+export async function waitForUnpack(ctx, url) {
+  ctx.log.debug(`Waiting for zip unpack sentinel file to appear at '${url}'`);
+
+  return retry(
+    async (bail) => {
+      let res;
+      try {
+        res = await ctx.http.fetch(url, {}, { retries: 0, noLogErrorBody: true });
```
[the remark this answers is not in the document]
codykaup: Great question! I added that piece because the response body is the full XML response from AWS about the 404 status. Since we query that file every few milliseconds, the debug output gets really messy. Skipping that response body cleans things up quite a bit!

codykaup:
> Is the sentinel file part of a known pattern?

It is! It's essentially an altered version of [sentinel value](https://en.wikipedia.org/wiki/Sentinel_value) that uses a file to send a notification rather than updating a value in memory!

bin-src/lib/uploadZip.js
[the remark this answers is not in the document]
ghengeveld: Okay 👍

ghengeveld:
> > Is the sentinel file part of a known pattern?
> 
> It is! It's essentially an altered version of [sentinel value](https://en.wikipedia.org/wiki/Sentinel_value) that uses a file to send a notification rather than updating a value in memory!

Thanks for clarifying! TIL

PR #461, opened 2021-12-02
Add `--untraced` flag to avoid retesting stories that depend on certain files

ghengeveld:
This introduces a new `--untraced` flag (originally `--ignore-changed`) that can be used in conjunction with `--only-changed` (i.e. TurboSnap). Even if any of the matching files have changed since the baseline build, they will be ignored when tracing dependent stories and so are their dependencies. Dependency tracing simply stops where it encounters a matching file. The flag accepts a glob and can be specified multiple times, just like `--externals`.


ghengeveld:
@tmeasday This PR does not really address the problem of `index.js` imports. Ignoring that file will not work if it's actually dependencies of that file which have changed. Should we instead stop the graph traversal when we encounter one of these ignored files? WDYT?

tmeasday:
> Should we instead stop the graph traversal when we encounter one of these ignored files? 

That was what I was thinking we should do here, yes.

bin-src/tasks/gitInfo.js
```diff
@@ -114,14 +114,19 @@ export const setGitInfo = async (ctx, task) => {
     try {
       const results = await Promise.all(baselineCommits.map((c) => getChangedFiles(c)));
       ctx.git.changedFiles = [...new Set(results.flat())];
-      ctx.log.debug(`Found changedFiles:\n${ctx.git.changedFiles.map((f) => `  ${f}`).join('\n')}`);
+      ctx.log.debug(
```
tmeasday, line 117: I think we should change these logs to not be debug, at least in CI 🤷

tmeasday (changes requested):
>> Should we instead stop the graph traversal when we encounter one of these ignored files?
> That was what I was thinking we should do here, yes.

@ghengeveld are we still planning on addressing this?

bin-src/tasks/gitInfo.js
ghengeveld: Done.

ghengeveld:
> > > Should we instead stop the graph traversal when we encounter one of these ignored files?
> > > That was what I was thinking we should do here, yes.
> 
> @ghengeveld are we still planning on addressing this?

Yes, done.

tmeasday (approved):
LGTM!

ghengeveld:
This was released in 6.3.0.

PR #463, opened 2021-12-06
Fix tests of makeZipFile

codykaup:
CircleCI wasn't happy about the tests with `mock-fs` (this worked locally 🙃 ).. this should fix it!


brockwood (approved):
Oh Circle...

PR #468, opened 2021-12-08
Increase unpack wait timeout to 3 minutes

codykaup:
Just a quick update to increase the sentinel file timeout to a minimum of 3 minutes to match the new Lambda timeout.

### DIFF
--- a/bin-src/lib/uploadZip.js
+++ b/bin-src/lib/uploadZip.js
@@ -71,8 +71,8 @@ export async function waitForUnpack(ctx, url) {
       }
     },
     {
-      retries: 30,
-      minTimeout: 500,
+      retries: 185, // 3 minutes and some change (matches the lambda timeout with some extra buffer)
+      minTimeout: 1000,
       maxTimeout: 1000,
     }
   );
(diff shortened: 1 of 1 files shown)


bin-src/lib/uploadZip.js
```diff
@@ -71,7 +71,7 @@ export async function waitForUnpack(ctx, url) {
       }
     },
     {
-      retries: 30,
+      retries: 360, // 3 minutes minimum (matches the lambda timeout)
       minTimeout: 500,
       maxTimeout: 1000,
```
codykaup, line 76: We _could_ decrease this `maxTimeout` but there's some time discrepancy between upload and the lambda kicking off. This gives us some buffer for that time difference.

brockwood (approved):
LGTM

bin-src/lib/uploadZip.js
```diff
@@ -60,6 +60,10 @@ export async function waitForUnpack(ctx, url) {
       try {
         res = await ctx.http.fetch(url, {}, { retries: 0, noLogErrorBody: true });
       } catch (e) {
+        const { response = {} } = e;
+        if (response.status === 403) {
```
codykaup, line 64: Apparently webpack hates optional chaining here and throws errors in our tests. 😭

jmhobbs (approved):

PR #477, opened 2021-12-10
Retry createBuild based on error messages

codykaup:
Connections to the Chromatic backend can get a 503 return error from Heroku after 30 seconds which are then retried by our [HTTPClient](https://github.com/chromaui/chromatic-cli/blob/c3ac4603fb8e1532aae61e1e9081c426a1a14399/bin-src/io/HTTPClient.js#L39-L55). We want to lower our error rates to reduce noise in our monitoring by throwing errors before that occurs. Doing this returns a 200 from the fetch then we throw an error in our [GraphQLClient](https://github.com/chromaui/chromatic-cli/blob/c3ac4603fb8e1532aae61e1e9081c426a1a14399/bin-src/io/GraphQLClient.js#L24-L37). This allows known errors to retry again when creating a build!

### DIFF
--- a/bin-src/io/GraphQLClient.js
+++ b/bin-src/io/GraphQLClient.js
@@ -1,4 +1,7 @@
+import retry from 'async-retry';
 import HTTPClient from './HTTPClient';
+
+const RETRYABLE_ERROR_CODE = 'RETRYABLE_ERROR_CODE';
 
 export default class GraphQLClient {
   constructor({ uri, ...httpClientOptions }) {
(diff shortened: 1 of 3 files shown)


bin-src/io/GraphQLClient.js
```diff
@@ -12,35 +15,46 @@ export default class GraphQLClient {
...
-    if (errors) {
-      if (Array.isArray(errors)) {
-        errors.forEach((err) => {
-          // eslint-disable-next-line no-param-reassign
-          err.name = err.name || 'GraphQLError';
-          // eslint-disable-next-line no-param-reassign
-          err.at = `${err.path.join('.')} ${err.locations
-            .map((l) => `${l.line}:${l.column}`)
-            .join(', ')}`;
+  async runQuery(query, variables, { headers = {}, retries = 0 } = {}) {
```
codykaup, line 18: Adding another field for `retries` felt wrong so I threw it all in one object since `headers` is only used in one other place!

bin-src/tasks/verify.js
```diff
@@ -1,3 +1,4 @@
+import retry from 'async-retry';
```
tmeasday, line 1: Should this be here?

tmeasday (approved):
This LGTM but I do wonder about the previous behaviour. It did retry before (on 5xx) right? Was that automatic behaviour of the `HTTPClient`?

codykaup:
> This LGTM but I do wonder about the previous behaviour. It did retry before (on 5xx) right? Was that automatic behaviour of the HTTPClient?

Yep! It would retry because [this fetch](https://github.com/chromaui/chromatic-cli/blob/c3ac4603fb8e1532aae61e1e9081c426a1a14399/bin-src/io/HTTPClient.js#L42) would throw an error. Since we're throwing the error ourselves, it returns a 200 so the fetch succeeds.

bin-src/tasks/verify.js
codykaup: It should not, good eye!

PR #482, opened 2021-12-17
Set commit details when using env var and warn if it's missing

ghengeveld:
When passing the commit SHA using an environment variable (either explicitly or by using GitHub Actions or Travis), also set the committer name / email or show a warning when the given commit isn't found in history.


ghengeveld:
Should add tests.

yannbf:
Can you add a test for the scenario you wrote?

ghengeveld:
Added a slew of tests.

ghengeveld:
This was released in 6.3.0.

PR #488, opened 2021-12-23
Fix untracing changed package files

ghengeveld:
Previously, `--untraced` would not be respected for changes to package files, making it impossible to ignore those, regardless of where they are in the repository. This fixes that.

I also dropped an outdated comment.


yannbf (approved):
💯 LGTM!

PR #499, opened 2022-01-14
Handle the `CANCELLED` build status separately

ghengeveld:
Currently it's getting lumped in with `FAILED` but they are two different things so I split them up.

Note I used the correct spelling here, but the Chromatic API still gives us the wrong spelling. Not sure what to do about that.

Docs update: https://github.com/chromaui/chromatic-docs/pull/102

### DIFF
--- a/bin-src/tasks/snapshot.js
+++ b/bin-src/tasks/snapshot.js
@@ -8,9 +8,10 @@ import buildPassedMessage from '../ui/messages/info/buildPassed';
 import speedUpCI from '../ui/messages/info/speedUpCI';
 import {
   buildComplete,
-  buildError,
-  buildFailed,
   buildPassed,
+  buildBroken,
+  buildFailed,
+  buildCancelled,
   initial,
   dryRun,
   skipped,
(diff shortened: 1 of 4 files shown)


bin-src/tasks/snapshot.js
```diff
@@ -95,13 +96,17 @@ export const takeSnapshots = async (ctx, task) => {
...
       break;
 
     case 'FAILED':
-    case 'CANCELLED':
       setExitCode(ctx, exitCodes.BUILD_FAILED, true);
-      transitionTo(buildError, true)(ctx, task);
+      transitionTo(buildFailed, true)(ctx, task);
+      break;
+
+    case 'CANCELLED':
```
tmeasday, line 107: Perhaps we should just handled both strings here with the assumption that we'll figure out a way to migrate the API at some stage?

ghengeveld: The only way I see that working is if we release patches for old versions of the CLI. If the index is all the sudden going to return `CANCELED`, old CLI versions are not going to accept that. Not that builds get canceled all that often, but still.

tmeasday: Eventually we will know when older versions have been phased out we can make the switch? Alternatively we could add a second type of "legacy" field on the `Build.status` resolver. I guess that'd be the more correct thing to do?

ghengeveld: We actually send the CLI version (`packageVersion`) when we create the build, so we could use that to send the appropriate response.

ghengeveld: Let's deal with this some other time.

PR #500, opened 2022-01-14
Use commit author info instead of committer info [CH-1282]

ghengeveld:
This updates the log format to yield the author info, which is more correct for merge commits.


tmeasday (commented):
LGTM

tmeasday (approved):

PR #503, opened 2022-01-19
Reformat help text and move `allowConsoleErrors` to deprecated options

ghengeveld:
Revisited the help text. `--allow-console-errors` doesn't do anything anymore (it was originally used with JSDOM) so I moved that to deprecated options. We may want to eventually add `--exit-zero-on-errors` as an alternative.


work933k:
Oh sjeez. Probably better to remove the "--allow-console-errors"  parameter. 
This has sent me on a wild-goose hunt.

PR #504, opened 2022-01-19
Fix consistent naming of flags for `trace` util

ghengeveld:
The main Chromatic CLI has `--storybook-base-dir` and `--storybook-config-dir`. This PR updates the flags for `trace` to use the same names.

This can be a patch release because `trace` is undocumented, effectively in beta, and not part of anyone's CI job.


kylegach (approved):
Thanks!

PR #505, opened 2022-01-20
Migrate to TypeScript

ghengeveld:
This migrates the entire CLI to TypeScript, squashing a couple of small bugs in the process.

Not sure whether this should be a patch release or a major release 🤔 I'm leaning towards patch because nothing should technically change beyond those small fixes and we won't have versioning on the GitHub Action anyway, so releasing it as major will still auto-upgrade everyone who uses that. We're just going to have to do proper QA on this (via the canary).

## QA instructions

Since this PR touches the entire codebase, we should do some extensive QA. Please try (some of) the following:

- Install `chromatic@canary` in one of your projects and have it run a build in CI
- Use `chromaui/action-canary` in your GitHub workflow and have it run a build in CI
- Run a local build using `npx chromatic@canary` and check if the output makes sense
- Run a local build with `--ci` and `--diagnostics` and check the CLI output as well as `chromatic-diagnostics.json`


bin-src/types.ts
```diff
@@ -0,0 +1,276 @@
...
+  runtimeWarnings?: Error[];
+  environment?: string;
+  reportPath?: string;
+  stopApp?: () => void;
+  closeTunnel?: () => void;
+  isPublishOnly?: boolean;
+  isOnboarding?: boolean;
+
+  http: {
+    fetch: (url: string, options?: Record<string, any>, opts?: any) => Promise<Response>;
```
ndelangen, line 142: Is this using node-fetch or something? I would assume so.. might just expose the real types for that.

ndelangen (approved):
So I tried to look all of it, but it's a LOT, I'm not sure I've seen everything, because all changes look really superficial.

I'm going place some trust in the CI and manual QA.

bin-src/types.ts
ghengeveld: Done 👍

PR #510, opened 2022-01-27
also use pathname for creating proxy.

l10rdev:
We need to use our own proxy to push to chromatic. This is specified with a host and subpath. But the pathname gets ignored. Maybe you could included that. We would appreciate.


ghengeveld (approved):

codykaup (approved):

ghengeveld:
Released in [6.4.2](https://github.com/chromaui/chromatic-cli/blob/main/CHANGELOG.md#642---2022-01-28)

PR #513, opened 2022-01-31
Add support for custom npm registry url

### LINKED ISSUE #512: Feature Request: Custom npm registry support
The company I work for disables access to the public registry, so that we're forced to use our internal registry instead - which includes a mirror of the public registry.

This causes [checkForUpdates](https://github.com/chromaui/chromatic-cli/blob/main/bin-src/lib/checkForUpdates.ts) to fail with the following message:

> Could not retrieve package info from registry; skipping update check

Ideally chromatic would automatically use the `registry` in my closest `.npmrc`. Alternatively this could be set with a CLI flag and/or env var.

ghengeveld:
Fixes #512

This attempts to retrieve the current registry url using `npm config get registry` or the Yarn equivalent. Failing that, it falls back to the default registry url for npm.

### DIFF
--- a/bin-src/lib/checkForUpdates.ts
+++ b/bin-src/lib/checkForUpdates.ts
@@ -18,12 +18,14 @@ export default async function checkForUpdates(ctx: Context) {
 
   let latestVersion: string;
   try {
-    const registryUrl = await spawn(['config', 'get', 'registry']).catch(() => '');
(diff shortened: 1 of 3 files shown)


ndelangen (approved):

ghengeveld:

PR #518, opened 2022-02-11
Fix `storybookUrl` by removing `iframe.html` suffix

### LINKED ISSUE #409: GitHub action output `storybookUrl` includes iframe.html
I'm setting up a PR workflow that publishes to Chromatic and then adds a Deployment using the `storybookUrl` output. But it looks like as of 5.10.0, that URL includes `/iframe.html`, which leads to the isolated view with no parameters.

IMO, it should just point to the root URL for the storybook deployment.

ndelangen:
update the action's storybookUrl output, to NOT include `iframe.html` as requested in #409

### DIFF
--- a/action-src/main.ts
+++ b/action-src/main.ts
@@ -107,7 +107,7 @@ async function runChromatic(options): Promise<Output> {
     url: ctx.build?.webUrl,
     code: ctx.exitCode,
     buildUrl: ctx.build?.webUrl,
-    storybookUrl: ctx.build?.cachedUrl,
+    storybookUrl: ctx.build?.cachedUrl.replace(/iframe\.html.*$/, ''),
   };
 }
 


action-src/main.ts
```diff
@@ -107,7 +107,7 @@ async function runChromatic(options): Promise<Output> {
     url: ctx.build?.webUrl,
     code: ctx.exitCode,
     buildUrl: ctx.build?.webUrl,
-    storybookUrl: ctx.build?.cachedUrl,
+    storybookUrl: ctx.build?.cachedUrl.replace(/iframe\.html.*$/, ''),
```
tmeasday, line 110: Why the `.*` at the end?

PS we also have a `build.storybookUrl` that we could request, perhaps could be more reliable?

ndelangen: I added that to also strip off any query params (which would not work for the manager)

ghengeveld: There should never be any query params on cachedUrl I think, but can't hurt to guard against it I suppose.

ghengeveld (approved):

ghengeveld:

PR #521, opened 2022-02-14
Add TurboSnap support for Vite

IanVS:
I'm on a mission to use turbosnap in my own project, which is built using storybook-builder-vite.  It's going to take a few different pieces:

- A rollup plugin to generate something similar to `preview-stats.json` from webpack.  
- Some changes to storybook-builder-vite (https://github.com/eirslett/storybook-builder-vite/pull/239)
- A few tweaks here to chromatic, to account for a few differences between webpack and vite.

### DIFF
--- a/bin-src/lib/getDependentStoryFiles.ts
+++ b/bin-src/lib/getDependentStoryFiles.ts
@@ -31,9 +31,6 @@ const isUserModule = (mod: Module | Reason) =>
 // already contains forward slashes, because that's what git yields even on Windows.
 const posix = (localPath: string) => localPath.split(path.sep).filter(Boolean).join(path.posix.sep);
 
-/** Strips off query params added by rollup/vite to ids, to make paths compatible for comparison with git */
-const stripQueryParams = (filePath: string): string => filePath.split('?')[0];
-
 /**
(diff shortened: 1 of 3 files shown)


bin-src/lib/getDependentStoryFiles.ts
```diff
@@ -145,9 +150,8 @@ export async function getDependentStoryFiles(
   function files(moduleName: string) {
     const mod = modulesByName[moduleName];
     if (!mod) return [moduleName];
-    return mod.modules && MULTI_MODULES.test(mod.name)
```
IanVS, line 148: I'd like to understand why this regex test was needed.  The naming for the vite modules does not include it.

IanVS:
I've been running turbosnap on my own repo with these (and other) changes `patch-package`ed in to my project, and it seems to be working great!  🎉 .  I've marked this as ready for review, and I think it could be merged any time without impacting existing webpack users of turbosnaps.

ghengeveld (approved):

yannbf:
Hey @IanVS thank you so much for all of your efforts! As always the work you've been doing is incredible and we're incredibly thankful for each and every contribution you make! ♥️

Thanks for helping Storybook and its  ecosystem thrive!

ghengeveld:

PR #536, opened 2022-02-25
fix: return slug from getBuildInfo for workflow_dispatch

### LINKED ISSUE #535: Bug: Action: Workflow Dispatch does not pass a slug
Slug is required for the action to run, but the workflow_dispatch config doesn't pass one.

See https://github.com/chromaui/chromatic-cli/blob/main/action-src/main.ts#L66-L72

~Anyone know where we can get the repository name from the context something so we can fix this?~

EDIT: NVM, I was able to propose a fix, see PR attached.

patrick-mcdougle:
This fix is untested but in theory it should behave better than it currently does (which is fail since slug is not defined when it tries to run).

Fixes #535


kylesuss:
Thanks for this @patrick-mcdougle . @ghengeveld -- does this make sense to you?

action-src/main.ts
```diff
@@ -65,7 +65,7 @@ const getBuildInfo = (event: typeof context) => {
 
       return {
         owner: event.payload.repository.owner.login,
-        repo: event.payload.repository.name,
+        slug: event.payload.repository.name,
```
ghengeveld, line 68:
```suggestion
        slug: event.payload.repository.full_name,
```
A slug is supposed to look like `ownerName/repoName` which is what `full_name` supposedly represents (based on [the docs](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#repository)).

action-src/main.ts
```diff
@@ -65,7 +65,7 @@ const getBuildInfo = (event: typeof context) => {
 
       return {
         owner: event.payload.repository.owner.login,
```
ghengeveld, line 67:
```suggestion
```
Don't need this anymore, it's not used anyway.

ghengeveld (changes requested):

ghengeveld (approved):

ghengeveld:
Thanks for contributing!

ghengeveld:
Released in 6.5.3

Issue #530, opened 2022-02-23
TurboSnap Causing Built Storybook Publish to Fail


victor-grajski:
Starting a new issue based on the discussion in #486, especially what @larrifax [mentioned](https://github.com/chromaui/chromatic-cli/issues/486#issuecomment-1004168690).

When I enable TurboSnap, Chromatic fails to publish my built Storybook and throws the following errors: `Could not retrieve dependent story files` and `Did not find any CSF globs in preview-stats.json`.

My `package.json` is at the root-level of my repo so I haven't yet needed to set a base directory, and trying various base directories with TurboSnap enabled haven't solved the issue either locally or using GitHub Actions.

Before enabling TurboSnap, I did not encounter this issue.

Here's the output when I run `npx chromatic --only-changed`:

```
Chromatic CLI v6.5.1
https://www.chromatic.com/docs/cli

  ✔ Authenticated with Chromatic
    → Using project token '********9caf'
  ✔ Retrieved git information
    → Commit 'e9a0c45' on branch '33-Integrate-Chromatic--victor-grajski'; found 1 parent build and 3 changed files
  ✔ Collected Storybook metadata
    → Storybook 6.4.19 for React; supported addons found: Actions, Backgrounds, Essentials, Links
  ✔ Storybook built in 39 seconds
    → View build log at /Users/victorgrajski/Documents/GitHub/mission-components/build-storybook.log
  ✖ Retrieving story files affected by recent changes
    → Did not find any CSF globs in preview-stats.json
    Verify your Storybook
    Test your stories

✖ Did not find any CSF globs in /var/folders/d2/6l5xyjhx20n7wz2p3lfxwf7w0000gn/T/chromatic--7327-uVXYzI5tusFU/preview-stats.json
Check your stories configuration in .storybook/main.js
ℹ Read more at https://storybook.js.org/docs/react/configure/overview

✖ Failed to publish your built Storybook
Could not retrieve dependent story files.
Did not find any CSF globs in preview-stats.json
→ View the full stacktrace below

If you need help, please chat with us at https://www.chromatic.com/docs/cli for the fastest response.
You can also email the team at support@chromatic.com if chat is not an option.

Please provide us with the above CLI output and the following info:
{
  "timestamp": "2022-02-23T00:50:57.017Z",
  "sessionId": "6bfafd31-9837-41a5-905e-32063d708b77",
  "gitVersion": "2.32.0 (Apple Git-132)",
  "nodePlatform": "darwin",
[...]

Issue #486, opened 2021-12-22
6.3.3 Did not find any CSF globs in preview-stats.json (It was working with 6.2.3)


VickyKoblinski:
We're using github actions to add Chromatic to our pipeline. It suddenly stopped working today and after some investigation, the only difference is that the CLI version used by the Github Action bumped up from 6.2.3 to 6.3.3.

```
    → Validating Storybook files
Retrieving story files affected by recent changes
    → Traversing dependencies for 108 files that changed since the last build
✖ Did not find any CSF globs in /tmp/chromatic--1765-AcyfWpX0wn54/preview-stats.json
Check your stories configuration in frontend/.storybook/main.js
Failed to retrieve dependent story files
ℹ Read more at https://storybook.js.org/docs/null/configure/overview
    → Could not retrieve dependent story files.
Did not find any CSF globs in preview-stats.json

✖ Failed to publish your built Storybook
Could not retrieve dependent story files.
Did not find any CSF globs in preview-stats.json
→ View the full stacktrace below

If you need help, please chat with us at https://www.chromatic.com/docs/cli for the fastest response.
You can also email the team at support@chromatic.com if chat is not an option.

Please provide us with the above CLI output and the following info:

{
  "timestamp": "2021-12-22T21:30:26.090Z",
  "sessionId": "3ba52350-0fdc-486b-aec5-418ba46dbb22",

  "gitVersion": "2.34.1",
  "nodePlatform": "linux",

  "nodeVersion": "12.13.1",
  "packageName": "chromatic",

  "packageVersion": "6.3.3",
  "storybook": {
Error: non-zero exit code
    "viewLayer": null,
    "version": null,
    "addons": []
  },
  "flags": {
    "projectToken": "***",
    "workingDir": "frontend",
    "skip": false,
    "onlyChanged": true,
    "fromCI": true,
    "interactive": false,
    "exitZeroOnChanges": true,
    "exitOnceUploaded": false,
    "allowConsoleErrors": false
  },
  "buildScript": "build-storybook",
  "spawnParams": {
    "client": "npm",
    "clientVersion": "8.1.2",
    "nodeVersion": "v16.13.1",
    "platform": "linux",
    "command": "npm",
    "clientArgs": [
      "run",
      "--silent"
    ],
    "scriptArgs": [
      "build-storybook",
      "--",
      "--output-dir",
      "/tmp/chromatic--1765-AcyfWpX0wn54",
      "--webpack-stats-json",
[...]

PR #527, opened 2022-02-21
Fix TurboSnap support for Storybook in a subdirectory

proshunsuke:
## Related

* https://github.com/chromaui/chromatic-cli/issues/486

## Current status

* I'm developing in Monorepo
* I'm managing Storybook in a sub directory of my git root directory.
* Storybook 6.4.19
* Chromatic CLI v6.4.3

example

```
repo/
├── .git
└── sub
    ├── .storybook
    ├── assets
    │   └── js
    │       └── components
    │           └── Button
    │               ├── index.stories.tsx
    │               └── index.tsx
    ├── package.json
    └── yarn.lock
```

## Problem

* When I run TurboSnap on v6.4.3, I get the following error message
  * ```
    Did not find any CSF globs in preview-stats.json
    ```
* In the current configuration, it doesn't seem to be able to find the dependent story files well

## Solution

* When I looked at the output stats file, I found that the entry file was in the format `${baseDir}/generated-stories-entry.js`, so I fixed it that way
  * However, I didn't know why it was in this format

## About this PR

* It's actually working well for my project, but I don't consider this PR to be perfect
* I created this PR in the hope that it will help somebody

## Logs

Some parts have been modified

<details>
<summary>Console log with TurboSnap enabled</summary>

```
Chromatic CLI v6.4.3
https://www.chromatic.com/docs/cli

Authenticating with Chromatic
    → Connecting to https://index.chromatic.com
Authenticated with Chromatic
    → Using project token '********8969'
Retrieving git information
Found 1 changed files:
  sub/assets/js/components/pages/top/IdentificationNotice/IdentificationNoticeExpired/index.tsx
Retrieved git information
    → Commit 'f4f5162' on branch 'turbo-snap'; found 1 parent build and 1 changed file
Collecting Storybook metadata
Collected Storybook metadata
    → Storybook 6.4.19 for React; supported addons found: Docs, Essentials, Viewport
Building your Storybook
    → Running command: /private/var/folders/l_/xkbfxbz12h54zlzfhf1d9pj40000gp/T/xfs-bdbf7468/yarn run --silent build-storybook --output-dir /var/folders/l_/xkbfxbz12h54zlzfhf1d9pj40000gp/T/chromatic--4643-b5qxu2wdjtgI --webpack-stats-json /var/folders/l_/xkbfxbz12h54zlzfhf1d9pj40000gp/T/chromatic--4643-b5qxu2wdjtgI
Storybook built in 48 seconds
[...]

PR #539, opened 2022-02-28
fix: normalize storiesEntryFiles in getDependentStoryFiles

yngvebn:
This PR fixes a bug related to situations where Chromatic (Turbosnap enabled) fails with the error "Did not find any CSF globs in preview-stats.json" (https://github.com/chromaui/chromatic-cli/issues/530 and https://github.com/chromaui/chromatic-cli/issues/486 - not sure if these are caused by the same error)

The error occurs if you're using Storybook 6.4 and pass in configDir as something else than '.' to Chromatic, and the reason stems from this array of predefined paths:

```ts
const storiesEntryFiles = [
    // v6 store (SB <= 6.3)
    `${storybookDir}/generated-stories-entry.js`,
    // v6 store with root as config dir (or SB 6.4)
    `generated-stories-entry.js`,
    // v6 store with .cjs extension (SB 6.5)
    'generated-stories-entry.cjs',
    // v7 store (SB >= 6.4)
    `storybook-stories.js`,
    // vite builder
    `/virtual:/@storybook/builder-vite/vite-app.js`,
];
```
  

In this case, `storybookDir` will be resolved as a combination of `baseDir` + `storybookConfigDir` whereas the rest are not. This fix ensures the storiesEntryFiles respects both `storybookBaseDir` and `storybookConfigDir`

### DIFF
--- a/bin-src/lib/getDependentStoryFiles.test.ts
+++ b/bin-src/lib/getDependentStoryFiles.test.ts
@@ -1,5 +1,4 @@
 import chalk from 'chalk';
-
 import { getDependentStoryFiles, normalizePath } from './getDependentStoryFiles';
 import * as git from '../git/git';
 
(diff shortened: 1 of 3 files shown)


tmeasday (approved):
This changes makes sense to me. We should always be looking for paths relative to the project root.

Thanks @yngvebn. @ghengeveld I'll let you merge if you agree + release 👍

ghengeveld (approved):

ghengeveld:
Thanks. This was similar to the work done in https://github.com/chromaui/chromatic-cli/pull/527 but yours is actually better because it normalizes all entry file paths, making it more robust and future-proof. I ended up using a bit from both PRs and cleaning up a couple things. Nice work!

ghengeveld:
Released in `6.5.2`.

PR #547, opened 2022-03-14
Fix line splitting on Windows for Git output

### LINKED ISSUE #546: Changed Files not detected correctly on Windows
Hi, 

There seems to be a bug in how changed files are detected when running in a Windows environment. I believe this is preventing me from getting the TurboSnap feature to work correctly. 

An example snippet of the broken behaviour is shown below

```
Found 1 changed files:
  .gitignore
package.json
```

(Expected)

```
Found 2 changed files:
  .gitignore
  package.json
```

We've tested the same command (on the same repo etc) in a Linux environment and changes are detected correctly. We've also modified the cli code locally as follows and managed to get it picking up the changes on Windows too (but it's a bit of a hacky fix!). 

```
// exportasync function getChangedFiles(baseCommit: string,headCommit ='')

return files.split(EOL).filter(Boolean);

// changed to...

returnfiles.replace("EOL","\n").split("\n").filter(Boolean);
```

From what we could tell, the EOL value is correct `\r\n` but `returnFiles` (git?) appears to contain `\n` only, hence the bug.

ghengeveld:
Fixes #546

### DIFF
--- a/bin-src/git/git.ts
+++ b/bin-src/git/git.ts
@@ -7,6 +7,8 @@ import gitNoCommits from '../ui/messages/errors/gitNoCommits';
 import gitNotInitialized from '../ui/messages/errors/gitNotInitialized';
 import gitNotInstalled from '../ui/messages/errors/gitNotInstalled';
 
+const newline = /(\r\n|\r|\n)/; // Git may return \n even on Windows, so we can't use EOL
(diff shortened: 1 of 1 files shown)


ghengeveld:
Released in 6.5.3

PR #554, opened 2022-03-25
downcase the slug so we don't accidentally treat origin as fork

JonathanKolnik:
if the slug is capitalized we are accidentally treating it as a fork. Always downcasing the getSlug method should prevent that.
https://linear.app/chromaui/issue/CH-1630/chromatic-cli-slug-isnt-downcased


JonathanKolnik:
https://linear.app/chromaui/issue/CH-1630/chromatic-cli-slug-isnt-downcased

ghengeveld (approved):

PR #559, opened 2022-04-11
Upgrade node-fetch to latest stable

ghengeveld:
Still blocked on `node:http` import which doesn't work with Node 12 and/or CommonJS (see smoke tests).


thafryer:
@ghengeveld closing this. I went ahead and resolved via the dependabot ticket.

PR #565, opened 2022-05-10
Reuben/ch 1867 add tests for untraced flag validation

ethriel3695:
## Index
TS - TurboSnap
docs - documentation
PR - pull request
Global decorator - Story wrapper component in `preview.js`

## Context

I've had a few customers reach out saying that TurboSnap bails out because of the `preview.js-generated-config-entry.js` file and variations of the file `.jsx, .tsx`

This is happening because of the dependencies being changed within a global decorator in the `preview.js` file

For instance, we have the following changed file:
`src/packages/design-system/components/button.jsx` // changed
Which then traces to this file change, which acts as a dynamic import of many components
`src/packages/design-system/component/src/component.jsx`
This traces to the file
`.storybook/decorator.jsx`
The decorator file is imported in `preview.js` as a global decorator and acts as a wrapper for each story

Really, the button is the only file being changed and all other traced changes are expected but not the global decorator because it's functionally created only for use as a wrapper for the story

We would like to advise customers to `add the decorator to the flag `--untraced **/decorator.jsx` like so

## Explanation

These unit tests act as both a validation measure to verify that this functionality does not stop working and also as a sanity check to make sure that customers can realistically use this functionality and we can prove that it works correctly and as intended

I've got specific support examples and trace dependency outputs for internal use if we need more specific context but I do not want to share that here in the PR

## Test cases
Four use cases are being handled with these tests when passing files to the flag `--untraced`:
- Verify that TS ignores untraced files and dependencies with a glob pattern match ('**/stories/**')
- Verify that TS does not bail when we pass global files (`'**/(package**.json|yarn.lock)'`)
- Verify that TS does not bail on untraced Storybook config files that change due to a module sibling (`'**/decorator.jsx'`)
- Verify that TS does not bail on a changed dependency from a dynamic import in an untraced config file (`'**/decorator.jsx'`)


bin-src/lib/getDependentStoryFiles.test.ts
```diff
@@ -470,6 +470,244 @@ describe('getDependentStoryFiles', () => {
...
+    ];
+    const modules = [
+      {
+        id: './src/utils.js',
+        name: './src/utils.js', // changed
+        reasons: [{ moduleName: './src/foo.js' }],
+      },
+      {
+        id: './src/foo.js',
+        name: './src/foo.js', // untraced
```
tmeasday, line 489: Better update this comment

bin-src/lib/getDependentStoryFiles.test.ts
```diff
@@ -470,6 +470,244 @@ describe('getDependentStoryFiles', () => {
...
+    });
+    const res = await getDependentStoryFiles(ctx, { modules }, statsPath, changedFiles);
+    expect(ctx.turboSnap.bailReason).toBeUndefined();
+    expect(res).toEqual({
+      './src/foo.stories.js': ['src/foo.stories.js'],
+    });
+  });
+
+  it('does not bail on untraced Storybook config files', async () => {
+    const changedFiles = ['src/utils.js', '.storybook/preview.js'];
```
tmeasday, line 514: We don't have a module for `.storybook/preview.js` so I don' think this test is "valid"

bin-src/lib/getDependentStoryFiles.test.ts
```diff
@@ -470,6 +470,244 @@ describe('getDependentStoryFiles', () => {
...
+        modules: [
+          { name: './path/to/storybook-config/preview.js-generated-config-entry.js' },
+          { name: './src/docs-wrapper.jsx' },
+        ],
+        reasons: [],
+      },
+      {
+        id: './src/docs-wrapper.jsx',
+        name: './src/docs-wrapper.jsx', // untraced
+        reasons: [{ moduleName: './src/docs-wrapper.stories.js' }],
```
tmeasday, line 562: We should add a module for this story too

bin-src/lib/getDependentStoryFiles.test.ts
```diff
@@ -470,6 +470,244 @@ describe('getDependentStoryFiles', () => {
...
+        id: CSF_GLOB,
+        name: CSF_GLOB,
+        reasons: [{ moduleName: './path/to/storybook-config/generated-stories-entry.js' }],
+      },
+    ];
+    const ctx = getContext({
+      configDir: 'path/to/storybook-config',
+      untraced: [
+        '**/docs-wrapper.jsx',
+        '**/path/to/storybook-config/preview.js-generated-config-entry.js',
```
tmeasday, line 589: Would we expect users to add the `generated-config-entry` to their untraced list?

bin-src/lib/getDependentStoryFiles.test.ts
```diff
@@ -470,6 +470,244 @@ describe('getDependentStoryFiles', () => {
...
+        modules: [
+          { name: './path/to/storybook-config/preview.js-generated-config-entry.js' },
+          { name: './src/docs-wrapper.jsx' },
+        ],
+        reasons: [],
+      },
+      {
+        id: './src/docs-wrapper.jsx',
+        name: './src/docs-wrapper.jsx',
+        reasons: [{ moduleName: './src/packages/design-system/components/button.jsx' }],
```
tmeasday, line 614: This is backwards I think? This is saying that `./src/packages/design-system/components/button.jsx` imports `./src/docs-wrapper.jsx` -- but if I understand the test correctly, I think you mean the opposite?

tmeasday (commented):
Some questions/comments

bin-src/lib/getDependentStoryFiles.test.ts
ethriel3695: No, this was a mistake on my part.
A user would never need to add the `generated-config-entry` to their untraced list.

bin-src/lib/getDependentStoryFiles.test.ts
ethriel3695: Done

bin-src/lib/getDependentStoryFiles.test.ts
ethriel3695: Yeah, this was an incorrect naming choice on the config file

ghengeveld (approved):

PR #572, opened 2022-05-20
Create `release` script and document it

ghengeveld:
This should prevent mistakes during the release process, and ensure the `next` tag and `chromaui/action-next` are updated whenever a `latest` version is released (so people who use `next` will get updated to `latest` as well).


tmeasday (approved):
LGTM

PR #577, opened 2022-05-25
Fix Issue with Getting Storybook Info in CLI

thafryer:


bin-src/lib/getStorybookInfo.ts
```diff
@@ -35,7 +37,8 @@ const findViewlayer = async ({ env, log, options, packageJson }) => {
   // Allow setting Storybook version via CHROMATIC_STORYBOOK_VERSION='@storybook/react@4.0-alpha.8' for unusual cases
   if (env.CHROMATIC_STORYBOOK_VERSION) {
     const [, p, v] = env.CHROMATIC_STORYBOOK_VERSION.match(/(.+)@(.+)$/) || [];
-    const version = semver.valid(v); // ensures we get a specific version, not a range
+    const version = semver.valid(v) || semver.validRange(v); // ensures we get a specific version, not a range
```
JonathanKolnik, line 40: Can we update this comment because I think it now might return a range right?

.storybook/main.js
```diff
@@ -3,4 +3,20 @@ module.exports = {
...
+    builder: 'webpack5',
+  },
+  staticDirs: ['../static'],
+  webpackFinal: async (config, { configType }) => {
+    // eslint-disable-next-line no-param-reassign
+    config.resolve = {
+      ...config.resolve,
+      fallback: {
+        ...config.resolve.fallback,
+        os: false,
```
JonathanKolnik, line 16: Can you explain this a bit more, I don't know this part of webpack well. Is this cause of the issue?

JonathanKolnik (commented):
Couple questions

bin-src/lib/getStorybookInfo.ts
thafryer: Actually, I'll revert this change. Understanding the context of using this flag, it's to explicitly use a specific version of Storybook.

thafryer: @JonathanKolnik I have reverted this change!

JonathanKolnik (approved):
Nice!!! :shipit:

bin-src/lib/getInstalledStorybookPackages.ts
```diff
@@ -0,0 +1,35 @@
+import execa from 'execa';
+import { viewLayers } from './viewLayers';
+import { supportedAddons } from './supportedAddons';
+
+export const getInstalledStorybookInfo = async () => {
+  const { all } = await execa.command('yarn list --pattern @storybook/ --json', {
```
ghengeveld, line 6: This is dangerous. People might not have Yarn installed (locally, or in CI).

package.json
```diff
@@ -49,7 +49,7 @@
     "release": "node scripts/release.js",
     "trace": "node -r esm bin-src/trace.js",
     "trim-stats": "node -r esm bin-src/trim-stats-file.js",
-    "storybook": "start-storybook -p 9009 -s static",
```
ghengeveld, line 52: Don't remove `-s`, Chromatic doesn't support `staticDirs` in main.js yet (for TurboSnap). Sucks, I know.

bin-src/lib/getInstalledStorybookPackages.ts
```diff
@@ -0,0 +1,35 @@
...
+    env: {},
+    timeout: 10000,
+    all: true,
+    shell: true,
+  });
+
+  const { data } = JSON.parse(all as any);
+  const trees: { name: string; version: string }[] = data.trees.map((p: any) => {
+    const [, name, version] = p.name.match(/(.+)@(.+)/);
```
ghengeveld, line 15: This needs a fallback in case the regex matches nothing, otherwise you'll be destructuring `undefined` which is a runtime error. E.g.:
```js
const [, name, version] = p.name.match(/(.+)@(.+)/) || [];
```

bin-src/lib/getInstalledStorybookPackages.ts
ghengeveld: Also the list command does not exist in Yarn 2+

ghengeveld: `yarn info @storybook/** --json` might do it in v2+.

PR #585, opened 2022-05-31
SB Metadata Retrieval

thafryer:
This PR updates how we retrieve the metadata for the Storybook being used in Chromatic.
1. If the Storybook is being uploaded via `-s `flag and the `project.json` file exists (meaning Storybook is v6.5+), we will use the data in that `project.json` file.
2. We also now have access to which builder was used.
3. We updated the way that we retrieve the package.json information.


bin-src/lib/getPrebuiltStorybookMetadata.ts
```diff
@@ -17,20 +21,41 @@ export interface SBProjectJson {
...
+  if (sbProjectJson.builder) {
+    const { builder } = sbProjectJson;
+    name = sbProjectJson.builder.name;
+    return {
+      name,
+      packageVersion: sbProjectJson.storybookPackages[builders[name]].version,
+    };
+  }
+
+  // Stroybook uses Webpack4 if no builder is explicitly set.
```
tmeasday, line 39:
```suggestion
  // Storybook uses Webpack4 if no builder is explicitly set.
```

tmeasday (commented):
I like it! Would like @ghengeveld to take a look though.

bin-src/lib/getStorybookInfo.ts
```diff
@@ -1,138 +1,25 @@
...
   ctx: Context
 ): Promise<Partial<Context['storybook']>> {
   try {
-    const info = await Promise.all([findAddons(ctx), findConfigFlags(ctx), findViewlayer(ctx)]);
-    return info.reduce((acc, obj) => Object.assign(acc, obj), {});
+    if (ctx.options.storybookBuildDir) {
+      const projectJsonPath = path.resolve(ctx.options.storybookBuildDir, 'project.json');
+      // This test makes sure we fall through if the file does not exist.
+      if (fs.pathExistsSync(projectJsonPath)) {
+        return await getStorybookMetadataFromProjectJson(projectJsonPath);
```
ghengeveld, line 17: Maybe add a comment to clarify `await` is required here in order for the `catch` block to apply to the result of this promise (typically `return await` is a mistake).

bin-src/lib/getPrebuiltStorybookMetadata.ts
```diff
@@ -0,0 +1,65 @@
...
+      name,
+      packageVersion: sbProjectJson.storybookPackages[builders[name]].version,
+    };
+  }
+
+  // Storybook uses Webpack4 if no builder is explicitly set.
+  return {
+    name: 'webpack4',
+    packageVersion: sbProjectJson.storybookVersion,
+  };
```
ghengeveld, line 43:
```suggestion
  const { builder, storybookPackages, storybookVersion } = sbProjectJson;
  return builder ? {
    name: builder.name,
    packageVersion: storybookPackages[builders[name]].version,
  } : {
    name: 'webpack4', // the default builder for Storybook v6
    packageVersion: storybookVersion,
  };
```

bin-src/lib/getPrebuiltStorybookMetadata.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  return {
+    viewLayer: sbProjectJson.framework.name ?? null,
+    version: sbProjectJson.storybookPackages[viewLayerPackage].version,
+    builder,
+    addons: Object.keys(sbProjectJson.addons).map((addon) => ({
+      name: supportedAddons[addon],
+      packageName: addon,
+      packageVersion: sbProjectJson.addons[addon].version,
+    })),
```
ghengeveld, line 63:
```suggestion
    addons: Object.entries(sbProjectJson.addons).map(([packageName, addon]) => ({
      name: supportedAddons[packageName],
      packageName,
      packageVersion: addon.version,
    })),
```

bin-src/lib/getStorybookMetadata.ts
```diff
@@ -0,0 +1,187 @@
...
+};
+
+export const findBuilder = async (mainConfig) => {
+  let name: string;
+  if (mainConfig?.core.builder) {
+    const { builder } = mainConfig.core;
+    name = typeof builder === 'string' ? builder : builder.name;
+  } else {
+    name = 'webpack4';
+  }
```
ghengeveld, line 165:
```suggestion
  let name = 'webpack4'; // default builder in Storybook v6
  if (mainConfig?.core.builder) {
    const { builder } = mainConfig.core;
    name = typeof builder === 'string' ? builder : builder.name;
  }
```

bin-src/lib/getStorybookInfo.test.ts
```diff
@@ -100,19 +140,63 @@ describe('getStorybookInfo', () => {
...
   });
 
   describe('with --storybook-build-dir', () => {
+    beforeAll(() => {
+      mock({
+        'storybook-static': {
+          'project.json':
+            '{"generatedAt":1654054690474,"builder":{"name":"webpack5"},"hasCustomBabel":false,"hasCustomWebpack":true,"hasStaticDirs":false,"hasStorybookEslint":false,"refCount":0,"packageManager":{"type":"yarn","version":"1.22.18"},"features":{"postcss":false},"storybookVersion":"6.5.6","language":"typescript","storybookPackages":{"@storybook/addon-essentials":{"version":"6.5.6"},"@storybook/builder-webpack5":{"version":"6.5.6"},"@storybook/eslint-config-storybook":{"version":"3.1.2"},"@storybook/linter-config":{"version":"3.1.2"},"@storybook/manager-webpack5":{"version":"6.5.6"},"@storybook/react":{"version":"6.5.6"}},"framework":{"name":"react"},"addons":{"@storybook/addon-viewport":{"version":"6.5.6"}}}',
+        },
+      });
```
ghengeveld, line 155: Is this mock even used, considering you're mocking out `./getPrebuiltStorybookMetadata` at the top? I don't see you disabling that mock to call through to the actual module. If it is, maybe you can change some mock data and the assertions along with it so that they don't overlap with the other mock (e.g. use webpack5 v6.5.4 for these tests, rather than 6.5.6).

bin-src/lib/getStorybookInfo.test.ts
```diff
@@ -1,8 +1,30 @@
...
+          packageName: '@storybook/addon-essentials',
+          packageVersion: '6.5.5',
+        },
+      ],
+      builder: { name: 'webpack5', packageVersion: '6.5.6' },
+      version: '6.5.5',
+      viewLayer: 'react',
+    })),
+  };
+});
```
ghengeveld, line 26: I'm not a fan of mocking at this level. Better to mock out "outgoing" APIs like the file system in this case. Otherwise we'll have a unit test rather than an integration test and those are much less valuable.

ghengeveld (changes requested):
Impressive work! I'm happy I didn't have to figure this out myself.
I left a bunch of suggestions for improvements, but my primary feedback is the (lack of) testing of `getPrebuiltStorybookMetadata`, because it's getting mocked out.

package.json
```diff
@@ -142,6 +146,7 @@
     "prop-types": "^15.7.2",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
+    "read-pkg-up": "^7.0.1",
```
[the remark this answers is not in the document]
thafryer: Exactly for that reason!

bin-src/main.ts
```diff
@@ -27,20 +26,18 @@ import runtimeError from './ui/messages/errors/runtimeError';
...
-const { readFile } = jsonfile;
-
 export async function main(argv: string[]) {
   const sessionId: string = uuid();
   const env = getEnv();
   const log = createLogger(sessionId, env);
 
-  const packagePath = await pkgUp(); // the user's own package.json
-  if (!packagePath) {
+  const pkgInfo = readPkgUp.sync({ cwd: process.cwd() });
```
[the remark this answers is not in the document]
thafryer: Good point! I'll switch to async version

bin-src/lib/getStorybookInfo.test.ts
thafryer: Resolved this by pointing at a mock project.json file that allows that actually uses the functions instead of mocking.

bin-src/lib/getStorybookInfo.ts
thafryer: Comment added!

ghengeveld (approved):

PR #587, opened 2022-06-01
Implement async build creation process (#2)

tmeasday:
Reverts chromaui/chromatic-cli#581

Also sends all TS information (apart from a `turboSnapEnabled` boolean) on the `publishBuild` mutation.

Depends on https://github.com/chromaui/chromatic/pull/6189


tmeasday (commented):
LGTM

PR #590, opened 2022-06-15
Better Onboarding Support in CLI

thafryer:
This PR
- leverages the isOnboarding flag from Index api
- forces rebuild on uncommitted changes if still onboarding.

1. In order to test this PR, you need to create a new project in Chromatic.
2. Pull my branch locally.
3. Follow the onboarding steps.
4. The CLI url to review changes or view build details should have /setup in path until onboarding is complete.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "chromatic",
-  "version": "6.6.3",
+  "version": "6.6.4-canary.0",
   "description": "Automate visual testing across browsers. Gather UI feedback. Versioned documentation.",
   "keywords": [
     "storybook-addon",


bin-src/tasks/auth.ts
```diff
@@ -8,9 +8,22 @@ const CreateAppTokenMutation = `
...
+  }
+`;
 interface CreateAppTokenMutationResult {
   createAppToken: string;
 }
+interface AppQueryResult {
+  app: {
+    isOnboarding: boolean;
+  };
+}
```
tmeasday, line 26: Does this need to be a separate query? Could we fetch this flag alongside the `lastBuild` etc?

tmeasday (commented):
I like the fix but I wonder if we could make it slightly more efficient.

bin-src/tasks/auth.ts
thafryer: The issue with fetching alongside `lastBuild` is that the BuildQuery happens too late. We need to know the onboarding state before we perform the `gitInfo` task so that we can automatically force the rebuild. If we do not know this the onboarding state before then, the build will get skipped if running against the same commit as the sole parent commit. See [diff](https://github.com/chromaui/chromatic-cli/pull/590/files#diff-95b2f101ab2de95280c00ec0784ff419bfd283e0a357ad61cfab005059e763a4R117).

thafryer: I considered returning isOnboarding in CreateAppTokenMutation, but it felt wrong to add there.

tmeasday: @thafryer hmm, couldn't we fetch the `isOnboarding` field [right there](https://github.com/chromaui/chromatic-cli/blob/b11fc488da4d574769464b7418bc6c9071a060f3/bin-src/tasks/gitInfo.ts#L100-L103) when we check if the `lastBuild` is `PASSED`?

thafryer: We can't do it there it because we need to check if the app is onboarding even if they have more than 1 parent commit. We basically need to check if the user is onboarding on every run. We could move this query up out of the if statement but this seemed to be a more expensive query to run every time than the one I added.

thafryer: But after sleeping on it some, it would not hurt to run that query every time too. Thoughts?

tmeasday: I'm not too bothered by it, I guess it'll generally be fine, if it needs to be a separate query then I guess that's fine.

andrewortwein (approved):
Looks good to me! 🔛🛹🏂

PR #592, opened 2022-06-17
Filter out unsupported addons from sb metadata

thafryer:
You can test this by running `yarn test` and reviewing the "getStorybookInfo does not return with --storybook-build-dir unsupported addons in metadata" test.

### DIFF
--- a/bin-src/lib/getPrebuiltStorybookMetadata.ts
+++ b/bin-src/lib/getPrebuiltStorybookMetadata.ts
@@ -52,11 +52,11 @@ export const getStorybookMetadataFromProjectJson = async (
     version: sbProjectJson.storybookPackages[viewLayerPackage].version ?? null,
     builder,
     addons: Object.entries(sbProjectJson.addons)
+      .filter(([packageName]) => supportedAddons[packageName])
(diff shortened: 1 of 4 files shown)


bin-src/lib/getPrebuiltStorybookMetadata.ts
```diff
@@ -51,10 +51,12 @@ export const getStorybookMetadataFromProjectJson = async (
     viewLayer: sbProjectJson.framework.name ?? null,
     version: sbProjectJson.storybookPackages[viewLayerPackage].version ?? null,
     builder,
-    addons: Object.entries(sbProjectJson.addons).map(([packageName, addon]) => ({
-      name: supportedAddons[packageName],
-      packageName,
-      packageVersion: addon.version,
-    })),
+    addons: Object.entries(sbProjectJson.addons)
+      .map(([packageName, addon]) => ({
```
yannbf, line 55: How about filtering the unsupported addons first, thus making the intentions more clear?

ghengeveld (approved):

yannbf (approved):

PR #596, opened 2022-06-24
Handle redirects when using `--storybook-url`

ghengeveld:
Rather than just assuming `iframe.html` will exist where we expect it to be, actually resolve the provided URL and add `iframe.html` to the resolved (redirected) URL.

### DIFF
--- a/bin-src/lib/startStorybook.ts
+++ b/bin-src/lib/startStorybook.ts
@@ -3,11 +3,11 @@ import { spawn } from 'cross-spawn';
 import path from 'path';
 import { Context } from '../types';
 
-export async function resolveIsolatorUrl(ctx: Context, url: string) {
+export async function resolveIsolatorUrl(ctx: Context, storybookUrl: string) {
   try {
     // Allow invalid certificates, because we might be running against localhost.
     const options = { proxy: { rejectUnauthorized: false } };
-    const { url: resolvedUrl } = await ctx.http.fetch(url, {}, options);
+    const { url: resolvedUrl } = await ctx.http.fetch(storybookUrl, {}, options);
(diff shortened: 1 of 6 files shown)


bin-src/lib/startStorybook.ts
```diff
@@ -3,19 +3,22 @@ import { spawn } from 'cross-spawn';
...
-export async function checkResponse(ctx: Context, url: string) {
+export async function resolveIsolatorUrl(ctx: Context, storybookUrl: string) {
   try {
-    // Allow invalid certificates, because we're running against localhost
-    await ctx.http.fetch(url, {}, { proxy: { rejectUnauthorized: false } });
-    return true;
+    // Allow invalid certificates, because we might be running against localhost.
+    const options = { proxy: { rejectUnauthorized: false } };
+    const { url: resolvedUrl } = await ctx.http.fetch(storybookUrl, {}, options);
+    const isolatorUrl = resolvedUrl.replace(/index\.html$/, '').replace(/\/?$/, '/iframe.html');
```
tmeasday, line 11: Should we just drop anything with a file extension?

tmeasday (approved):
OK, LGTM!

bin-src/lib/startStorybook.ts
ghengeveld: I had considered that at first, but decided against it because that doesn't align with directory index behavior, and we might end up stripping a part of the path which was intentional. A directory name can technically look like a filename.

ghengeveld:
Released in [6.6.4](https://github.com/chromaui/chromatic-cli/blob/main/CHANGELOG.md#664---2022-06-27).

PR #598, opened 2022-06-29
Enable bash mode for globs when matching branches with picomatch

andrewortwein:
Several customers have reported that using the `--only-changed` flag with a glob (e.g. `"!(main)"`) doesn't work properly if they are running Chromatic on a branch that has a forward slash in its name (e.g. "andrew/feature"). In these scenarios, TurboSnap is disabled and the customers are producing far more snapshots than they expected.

It turns out that picomatch, the package being used to match branch names against the given glob, is strict about how it uses forward slashes. By default, a catch-all glob like "*" will **not** match a string with a forward slash, like "andrew/feature". You must use something like "**" to catch the afore-mentioned branch along with other simpler branch names.

By enabling the `bash` configuration option, we tell picomatch to follow bash matching rules more strictly, treating single stars more like globstars. This fixes our issue with branch names containing forward slashes, though there might be performance implications. More info on that here: https://github.com/micromatch/picomatch#matching-behavior-vs-bash.

### DIFF
--- a/bin-src/tasks/gitInfo.test.ts
+++ b/bin-src/tasks/gitInfo.test.ts
@@ -80,6 +80,21 @@ describe('setGitInfo', () => {
     expect(ctx.git.replacementBuildIds).toEqual([]);
   });
 
+  it('sets changedFiles on a branch name containing a slash', async () => {
+    getCommitAndBranch.mockResolvedValue({
+      ...commitInfo,
+      branch: 'something/else',
+    });
+    getBaselineBuilds.mockResolvedValue([{ commit: '012qwes' } as any]);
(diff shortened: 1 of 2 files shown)


ethriel3695 (approved):
Great work here, Andrew!
I love that you added the test too.

ethriel3695:
@andrewortwein
I think you are correct.
If we end up implementing this change, we will have to let users know because patterns like this one: !(foo)* became memory inefficient because it would match foo and foobar according to the documentation.

thafryer (approved):
Looks Good! Let's get with Dom/João on the docs update!

tmeasday (approved):
LGTM but let's get @ghengeveld to agree!

PR #612, opened 2022-07-21
Add change count to buildPassed message if changes exist.

thafryer:
When using the Github Action, the flag `exitZeroOnChanges` is equal to true when not set explicitly equal to false. So, if the build status is `ACCEPTED`, `PENDING` or `DENIED`, we always return 'No visual changes were found in this build.' even if there was changes. This PR updates the message so if changes are found, the number of changes are shown in action output.


tmeasday (approved):
LGTM! Makes sense.

PR #613, opened 2022-07-22
6.7.1

thafryer:


JonathanKolnik:
how does this relate to the version published to npm? https://www.npmjs.com/package/chromatic

did that get picked up correctly?

thafryer:
@JonathanKolnik Yes. The version on NPM is now 6.7.1. We currently do not release the CLI via automation so when you deploy to latest, the version gets bumped so this is to commit the version bump to git.

PR #614, opened 2022-07-26
Update CHANGELOG.md

thafryer:


skitterm (approved):
Looks good! Seems some other entries are also missing, in case we need to tweak the release script in the future.

PR #615, opened 2022-07-26
Changed trim stats to stream and added test

ethriel3695:
## Context
This PR aims to address the issue that we are seeing for some of our customers, where the Webpack preview stats file is too large to parse through for the TurboSnap process.

In Storybook, we already write a stream for the stats file so it makes sense that we could read a stream and get the file contents accordingly.
Here is the file reference in Storybook
https://github.com/storybookjs/storybook/blob/next/code/lib/core-server/src/utils/output-stats.ts

## Changes
I added a function to process the reading of the stats file so I could add a test for that functionality
Added a jest test to verify that the `readStatsFile` function is returning an Array of objects as expected with an `id` property.

Added a new library that we also use in Storybook
https://github.com/discoveryjs/json-ext#stringifystreamvalue-replacer-space

## Testing
1. I ran `npx chromatic trim-stats-file` on my branch and saved the file locally in a different location
2. I then pushed my latest changes
3. I switched back to my main branch, pulled the latest code, and ran `yarn` to update dependencies
4. Then, I ran `npx chromatic trim-stats-file` again and compared the file with the other file I had saved locally and which was built and trimmed off my branch.

The image shows an exact match other than white space
Screen Shot 2022-07-27 at 9 30 27 PM


bin-src/trim-stats-file.test.ts
```diff
@@ -0,0 +1,17 @@
...
+import { execSync } from 'child_process';
+import { main } from './trim-stats-file';
+// eslint-disable-next-line jest/no-mocks-import
+import * as trimmedStats from './__mocks__/previewStatsJson/preview-stats.trimmed.json';
+// eslint-disable-next-line jest/no-mocks-import
+import * as baseTrimmedStats from './__mocks__/previewStatsJson/baseline-preview-stats.trimmed.json';
+
+describe('Trim Stats File', () => {
+    it('returns expected file output', async () => {
+        execSync('yarn build-storybook --webpack-stats-json', { encoding: 'utf-8' });
```
yannbf, line 11: I think these tests should probably be reworked a little because right now they execute the actual `build-storybook` command as well as `chromatic trim-stats-file`. Probably what you might want is to, or rather, what I'd suggest is to:

- Store your `baseTrimmedStats` as mocks
- Run `yarn build-storybook --webpack-stats-json`, and store that value as mock for Storybook's `preview-stats.json`
- Mock file system so the `./storybook-static/preview-stats.json` path always returns that mock 
- Spy on `outputFile` from `fs-extra`
- run `await main(['./storybook-static/preview-stats.json'])`;
- Use that `outputFile` spy and expect it to be called with `baseTrimmedStats`

ethriel3695: Thanks, Yann.

After our conversation today, I rewrote the test and ran the file comparison in VSCode to verify the changes are non-breaking.

tmeasday (approved):
Awesome! LGTM @ethriel3695

package.json
```diff
@@ -51,6 +51,7 @@
     "trim-stats": "node -r esm bin-src/trim-stats-file.js",
     "storybook": "start-storybook -p 9009 -s static",
     "test": "jest",
+    "test:file": "jest -t",
```
yannbf, line 54: If I understand this correctly, you're using this command to run a specific test? Like `yarn test:file trim`.
If that's the case, I believe you can delete this line. `yarn test trim` would work as well for that use case

bin-src/trim-stats-file.test.ts
```diff
@@ -0,0 +1,21 @@
...
+afterEach(() => {
+    mockfs.restore();
+})
+
+describe('Trim Stats File', () => {
+    it('readStatsFile returns expected output', async () => {
+        const result = await readStatsFile('./storybook-static/preview-stats.json');
+        expect(result.modules.some(({ id }) => id === './bin-src/ui/components/icons.stories.ts')).toBe(true);
+    });
+});
```
ghengeveld, line 21: What's up with the 4 space indent in this file? Maybe you need to update your editor settings?

ghengeveld (approved):
Nice, I like it!

bin-src/trim-stats-file.test.ts
ethriel3695: Thanks Gert!

PR #621, opened 2022-07-29
Add addon-interactions to supportedAddons in cli

JonathanKolnik:
This should be considered a supported addon and I think it was left out by mistake


thafryer (approved):
LGTM!

PR #624, opened 2022-08-03
Update stats process to use stream

ethriel3695:
This builds off of the changes to the `trim-stats-file`.
We are now moving to change all instances of the stats file to use `readStatsFile` which will read stats files from a stream.
This will allow handling larger stats files and avoid hitting a file size limit.

### DIFF
--- a/bin-src/tasks/upload.test.ts
+++ b/bin-src/tasks/upload.test.ts
@@ -1,7 +1,7 @@
 import * as fs from 'fs-extra';
 import progressStream from 'progress-stream';
 
-import * as trimStatsFile from '../trim-stats-file';
+import * as readStatsFile from './read-stats-file';
 import { getDependentStoryFiles as dependentStoryFiles } from '../lib/getDependentStoryFiles';
 import { validateFiles, traceChangedFiles, uploadStorybook } from './upload';
 
@@ -91,7 +91,7 @@ describe('validateFiles', () => {
 });
 
 describe('traceChangedFiles', () => {
-  jest.spyOn(trimStatsFile, 'readStatsFile').mockReturnValueOnce(Promise.resolve(undefined));
+  jest.spyOn(readStatsFile, 'readStatsFile').mockReturnValueOnce(Promise.resolve(undefined));
 
   it('sets onlyStoryFiles on context', async () => {
     const deps = { 123: ['./example.stories.js'] };
--- a/bin-src/tasks/upload.test.ts
+++ b/bin-src/tasks/upload.test.ts
@@ -1,6 +1,7 @@
 import * as fs from 'fs-extra';
 import progressStream from 'progress-stream';
 
+import * as trimStatsFile from '../trim-stats-file';
 import { getDependentStoryFiles as dependentStoryFiles } from '../lib/getDependentStoryFiles';
 import { validateFiles, traceChangedFiles, uploadStorybook } from './upload';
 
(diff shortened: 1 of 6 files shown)


bin-src/trim-stats-file.ts
```diff
@@ -8,8 +8,12 @@ const isUserCode = ({ name, moduleName = name }: { name?: string; moduleName?: s
   !moduleName.startsWith('(webpack)') &&
   !moduleName.match(/(node_modules|webpack\/runtime)\//);
 
-export const readStatsFile = async (filePath: string): Promise<Stats> =>
-  parseChunked(createReadStream(filePath));
+export const readStatsFile = async (filePath: string): Promise<Stats> => {
+  if (existsSync(filePath)) {
```
ethriel3695, line 12: I added this check because I found that in the `upload.test.ts` test for traced files, it was erroring out because the file path does not return an existing file.

I attempted to mock the file path but because we mock `fs-extra` it overrides all my mock tests and cannot find the file.

The logic check ensures that we don't error out on a stats file with a directory that we cannot process.

However, once the trace or upload reaches the point where it's processing the stats file, we have already checked for a file path and that the file exists so we should be able to reasonably assume that the file exists and this is probably an indicator that the test needs to be refactored.

ethriel3695: What is strange is that when you include a combination of the `sourceDir` and `statsPath` the upload process builds a path that would most likely not exist in production.
```
sourceDir: '/static/',
fileInfo: { statsPath: '/static/preview-stats.json' },
```
This is the path: `static/static/preview-stats.json`
This is the existing test so I'm hesitant to change it but that file path does not seem like something that would ever exist in a real project.

So is the test incorrect, is the path join incorrect, or is this expected?
@ghengeveld @tmeasday

yannbf (approved):
This all looks good to me, thanks for the good work @ethriel3695 !

thafryer (approved):
I like!! Good work @ethriel3695 !!

bin-src/tasks/upload.ts
```diff
@@ -29,6 +29,7 @@ import {
 } from '../ui/tasks/upload';
 import { Context, Task } from '../types';
 import { exitCodes, setExitCode } from '../lib/setExitCode';
+import { readStatsFile } from '../trim-stats-file';
```
tmeasday, line 32: I think this code should probably live somewhere else, perhaps `read-stats-file.ts`?

bin-src/tasks/upload.test.ts
```diff
@@ -90,6 +91,8 @@ describe('validateFiles', () => {
 });
 
 describe('traceChangedFiles', () => {
+  jest.spyOn(trimStatsFile, 'readStatsFile').mockReturnValueOnce(Promise.resolve(undefined));
```
tmeasday, line 94: How was this file previously mocking the reading of JSON?

tmeasday (commented):
Couple comments!

bin-src/tasks/upload.test.ts
yannbf: It wasn't. Just resolved to undefined and did not break. The tests could be improved if there are use cases that require stats files to contain something though

bin-src/tasks/upload.ts
ethriel3695: This is complete

bin-src/trim-stats-file.ts
tmeasday: @ethriel3695 reading where `statsPath` comes from, it is relative to the `sourceDir`: https://github.com/chromaui/chromatic-cli/blob/0d1bd21f838decd4709d66db55c525870ab0bef5/bin-src/tasks/upload.ts#L77-L117

So I think that simply the test is wrong here.

tmeasday (approved):
🙌

PR #629, opened 2022-08-17
Deprecate `--preserve-missing` and raise a warning if it's being used

ghengeveld:
Screen Shot 2022-08-17 at 15 00 47

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -85,7 +85,7 @@ inputs:
     description: 'Storybook is already running at (external) url (implies -S)'
     required: false
   preserveMissing:
-    description: 'Pass the baselines forward and treat all missing stories as “preserved” without re-capturing them'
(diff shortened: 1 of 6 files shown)


ShaunLloyd (approved):
Looks good, @ghengeveld!

thafryer (approved):
LGTM! I'll create a ticket to remove it for v7.0.0 .

adybionka:
Where can I find more information, on why this flag is going to be deprecated? 

It will be possible to send only a subset of stories Chromatic, and the rest will be considered as unchanged?

ghengeveld:
We're working on the updated documentation for this. Should be out soon. I'll also update the changelog with some details.

The gist is that preserveMissing is troublesome when upgrading capture infrastructure, makes baseline logic more complicated and leads to stories never getting removed from Chromatic. As a replacement we offer TurboSnap (`onlyChanged`), `onlyStoryFiles` and `onlyStoryNames` (previously `only`).

PR #630, opened 2022-08-17
Rename `--only` to `--only-story-names` but keep it as a deprecated alias

ghengeveld:
Screen Shot 2022-08-17 at 15 44 05

Contrary to `--only`, `--only-story-names` can be specified multiple times.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -43,10 +43,13 @@ inputs:
     description: 'Relative path from where you run Chromatic to your Storybook config directory'
     required: false
   only:
-    description: 'Only run a single story or a subset of stories'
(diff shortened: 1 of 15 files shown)


action.yml
```diff
@@ -43,11 +43,14 @@ inputs:
     description: 'Relative path from where you run Chromatic to your Storybook config directory'
     required: false
   only:
-    description: 'Only run a single story or a subset of stories'
+    description: 'Deprecated, use onlyStoryNames instead'
```
thafryer, line 46: Small note. Let's choose between Deprecated or [Deprecated] to keep it more uniform.

thafryer (approved):

PR #635, opened 2022-09-01
Retry queries by default and allow even more retries for `FirstCommittedAtQuery`

ghengeveld:
This should deal with network flake and handle brief outages at the upstream Git provider (for up to several minutes).

### DIFF
--- a/bin-src/git/getParentCommits.ts
+++ b/bin-src/git/getParentCommits.ts
@@ -182,10 +182,11 @@ export async function getParentCommits(
   const { branch, commit, committedAt } = git;
 
   // Include the latest build from this branch as an ancestor of the current build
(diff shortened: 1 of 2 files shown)


tmeasday (approved):
Makes sense. I guess we had the zero retries from way back when the index couldn't handle multiple calls to the creation mutation. Is it ok now to do that?

ghengeveld:
The `AnnounceBuildMutation` and `PublishBuildMutation` are already configured with 3 retries. The only mutations affected are `SkipBuildMutation` and `GetUploadUrlsMutation`.

For `skipBuild` it doesn't matter if it's retried because all it does is set a commit status (which may actually be good to retry, since it connects to an upstream provider). For `uploadUrls` I don't think it matters either, because it's just generating a bunch of random strings. It doesn't matter if we don't end up using them.

So I think we're all good.

PR #636, opened 2022-09-01
Suggest using `--force-rebuild` when skipping rebuild

ghengeveld:
This should clear up some confusion and help people on their way.


tmeasday (approved):
Love it 😍

PR #637, opened 2022-09-07
Finer-grained TurboSnap: Determines dependent story files based on package.json dependency version changes

skitterm:
Note: Code is pretty rough, this is more to get the approach out there for review (I need to add tests, get better variable names, move things around, etc).

I have an initial approach (thanks @ghengeveld for pairing on this) for keeping TurboSnap enabled when a dependency version in a `package.json` file changes. This will enable us to avoid bailing if a change happens to a package file or a lockfile. We may eventually look in lockfiles for changes, but the same pattern should apply.

Adds dependency [`parse-diff`](https://www.npmjs.com/package/parse-diff) to make it easier to tell what lines actually changed.

To test:
 - Create a placeholder repo with a storybook and Chromatic project, using TurboSnap
 - yarn link this repo to that project
 - Add a dependency to a component in that project
 - Commit and run a build
 - Tweak the dependency in package.json (doesn't matter for now if it's valid)
 - (without committing) run another build (temporary thing)
 - Verify that the story/stories for the component using that dependency is snapshotted, but nothing else

### DIFF
--- a/bin-src/lib/getDependentStoryFiles.ts
+++ b/bin-src/lib/getDependentStoryFiles.ts
@@ -1,26 +1,16 @@
 import path from 'path';
-
+import parseDiff from 'parse-diff';
+import fs from 'fs-extra';
 import { matchesFile } from './utils';
-import { getRepositoryRoot } from '../git/git';
+import { getRepositoryRoot, execGitCommand } from '../git/git';
 import bailFile from '../ui/messages/warnings/bailFile';
 import noCSFGlobs from '../ui/messages/errors/noCSFGlobs';
 import tracedAffectedFiles from '../ui/messages/info/tracedAffectedFiles';
 import { Context, Module, Reason, Stats } from '../types';
 
-// Bail whenever one of these was changed
-const GLOBALS = [
-  /^package\.json$/,
-  /^package-lock\.json$/,
-  /^yarn\.lock$/,
-  /\/package\.json$/,
-  /\/package-lock\.json$/,
-  /\/yarn\.lock$/,
-];
-
 // Ignore these while tracing dependencies
-const EXTERNALS = [/^node_modules\//, /\/node_modules\//, /\/webpack\/runtime\//, /^\(webpack\)/];
+const EXTERNALS = [/\/webpack\/runtime\//, /^\(webpack\)/];
 
-const isPackageFile = (name: string) => GLOBALS.some((re) => re.test(name));
 const isUserModule = (mod: Module | Reason) =>
   (mod as Module).id !== undefined &&
   (mod as Module).id !== null &&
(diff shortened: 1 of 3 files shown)


bin-src/lib/getDependentStoryFiles.ts
```diff
@@ -31,6 +21,55 @@ const isUserModule = (mod: Module | Reason) =>
...
+      })
+    );
+
+    console.log('changed', dependenciesChanged);
+  } catch (err) {
+    console.error(err);
+  }
+
+  return dependenciesChanged;
+};
```
skitterm, line 71: This function needs to move to its own dedicated file, and I need to finesse the string-splitting and other stuff. Also, for now it's actually using the unstaged package.json's changes for the diff, instead of diffing based off of the baseline's commit. Will work on that next week.

bin-src/lib/getDependentStoryFiles.ts
```diff
@@ -64,6 +103,8 @@ export async function getDependentStoryFiles(
     untraced = [],
   } = ctx.options;
 
+  const changedModules = await getExternalChangedDirectDependencies(changedFiles);
+
```
skitterm, line 107: As per conversation with @ghengeveld I'll supply this as a parameter to `getDependentStoryFiles` for easier testing.

bin-src/lib/getDependentStoryFiles.ts
```diff
@@ -31,6 +21,55 @@ const isUserModule = (mod: Module | Reason) =>
...
+        const changes = await execGitCommand(`git --no-pager diff ${packageFilePath}`);
+        const filez = parseDiff(changes);
+        filez[0].chunks.forEach((chunk) => {
+          const theChanges = chunk.changes.filter((item) => item.type === 'add');
+          // make sure this change is a proper change... should be easy enough...
+          // honestly just pull out the package name
+          // (later check that it is in proper format, maybe just using semver)
+          const refinedChanges = theChanges.map((aChange) =>
+            aChange.content.split('+')[1].split(':')[0].trim().replace(/"/g, '')
+          );
```
tmeasday, line 42: I feel like there might be some ways this could go wrong (what if the dependency goes over two lines somehow, etc)?

Would we be better of actually just getting the before and after JSON of the `package.json` and comparing directly?

bin-src/lib/getDependentStoryFiles.ts
```diff
@@ -74,6 +115,16 @@ export async function getDependentStoryFiles(
...
   const staticDirs = staticDir.map((dir: string) => normalize(posix(dir)));
 
+  const getPackageName = (modulePath) => {
+    const [, scopedName] = modulePath.match(/\/node_modules\/(@[\w-]+\/[\w-]+)\//) || [];
+    if (scopedName) {
+      return scopedName;
+    }
+
+    const [, unscopedName] = modulePath.match(/\/node_modules\/([\w-]+)\//) || [];
+    return unscopedName;
```
tmeasday, line 125: One thing to watch out for here is `node_modules/x/node_modules/y` -- we want this to be `y` not `x`. So you'd want to do a negative lookahead for any more `node_modules` I think?

bin-src/lib/getDependentStoryFiles.ts
```diff
@@ -90,6 +141,11 @@ export async function getDependentStoryFiles(
   ].map(normalize);
 
   const modulesByName: Record<string, Module> = {};
+  // stores the normalized paths (like modulesByName does), but groups all package files under the same package name.
+  // Why? In monorepo, may have multiple node_modules directories, but we want to update any components using an updated dependency
+  // (regardless of where it's used). So we use the package name as the key, and the value is all files of that package name,
+  // regardless of where they are.
+  const thirdPartyModules: Record<string, string[]> = {};
```
tmeasday, line 148: I find this comment a bit confusing, but I think I understand the point based on how it's used below. This is mapping a package name (e.g. `@storybook/addon-actions`) to all the module names we find that we think are part of that package.

I wonder if maybe using some type aliases (so instead of `Record<string, string[]>` it's `Record<PackageName, NormalizedModulePath[]>` or something), might make this easier to understand at a glance.

bin-src/lib/getDependentStoryFiles.ts
```diff
@@ -153,7 +217,11 @@ export async function getDependentStoryFiles(
     return mod.modules?.length ? mod.modules.map((m) => normalize(m.name)) : [normalize(mod.name)];
   }
 
-  const tracedFiles = changedFiles.filter(untrace);
+  // append the third-party modules (respecting any untraced rules)
+  const tracedFiles = changedModules
+    .reduce((acc, mod) => acc.concat(thirdPartyModules[mod]), changedFiles)
+    .filter(untrace);
```
tmeasday, line 223:
```suggestion
  const tracedFiles = [...changedModules.flatMap(mod => thirdPartyModules[mod]), ...changedFiles]
    .filter(untrace);
```

Might be a bit simpler?

bin-src/lib/getDependentStoryFiles.ts
```diff
@@ -213,6 +278,7 @@ export async function getDependentStoryFiles(
     checkedIds[id] = true;
     reasonsById[id].filter(untrace).forEach((reason) => traceName(reason, tracePath));
   }
+  // TODO: should rename (this is webpack modules) -- maybe affectedStoryFiles?
```
tmeasday, line 281: You might want to consider calling it `changedPackages` instead of `changedModules` also. Even though the directory is called `node_modules`, I'm pretty sure when people say "module" they usually mean *a single file*. Eg, from [the node docs](https://nodejs.org/api/modules.html):

> In Node.js, each file is treated as a separate module

And ESM definitely is referring to single files as the ES "modules". Webpack too calls each file a module. I'm not sure what the history of "`node_modules`" is, but it was a mistake, I guess!

tmeasday (changes requested):
The approach is very much in line with what I was thinking, a few things that might be improved.

bin-src/lib/getDependentStoryFiles.ts
skitterm: Thanks for the heads-up, will give this a look.

skitterm: @tmeasday thanks for this point, just to make sure I'm on the same page -- you're suggesting to do something like `git show` on the file at a given commit, and then diff the keys of the dependencies (and devdeps/peerdeps) to find the difference? Sounds good, just wanted to make sure I understood your point.

tmeasday: Yep, that's it! I realised as I wrote that comment that that approach might be a lot easier than trying to parse + apply the diff.

thafryer:
@ghengeveld @skitterm Closing this since there's a new PR from Gert.

PR #646, opened 2022-09-28
Fix Issues finding SB metadata

thafryer:
We are having issues getting Storybook metadata to show up in the CLI output. There were two reasons.

1. If the user did not have a `core.builder` defined in their SB config, `mainConfig?.core.builder` would return undefined. So, we added optional chaining to resolve this.
2. Because we bundle the CLI via webpack, we needed to use __non_webpack_require__ in order to use the actual require functionality of node. However, for some reason, it was having issues resolving node_modules and so we could never find the installed package.

### DIFF
--- a/bin-src/lib/getStorybookMetadata.ts
+++ b/bin-src/lib/getStorybookMetadata.ts
@@ -76,7 +76,6 @@ const findViewlayer = async ({ env, log, options, packageJson }) => {
       // Note that `version` can be a semver range in this case.
       return { viewLayer, version };
     }
-
     // Verify that the viewlayer package is actually present in node_modules.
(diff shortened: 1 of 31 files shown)


bin-src/lib/getStorybookMetadata.ts
```diff
@@ -181,5 +176,6 @@ export const getStorybookMetadata = async (ctx: Context) => {
     findViewlayer(ctx),
     findBuilder(mainConfig),
   ]);
+  ctx.log.debug(info);
```
JonathanKolnik, line 179: just checking this debug is intentional?

JonathanKolnik (approved):
👍

PR #648, opened 2022-10-06
TurboSnap not bailed if package manifest change isn't dependency-related

skitterm:
TurboSnap doesn't bail if a package.json changes non-dependency fields (like `scripts`, `version`, `license`, etc). It still bails if a dependency field changes.

<s>Currently I'm only looking for `dependencies`, `devDependencies`, and `peerDependencies`. Technically there are [many](https://docs.npmjs.com/cli/v8/configuring-npm/package-json/#peerdependenciesmeta) [more](https://classic.yarnpkg.com/lang/en/docs/package-json/#toc-resolutions) dependency-related fields that I could add to the list. Happy to loop back and add them in if we want to cover our bases.</s> We look for changes in any dependency-related package.json field.

Also changes wording for the package-json-related dependency changes ( `Found a package file change in package.json` -> `Found a dependency change in package.json`)


bin-src/lib/getDependentStoryFiles.ts
```diff
@@ -172,8 +165,16 @@ export async function getDependentStoryFiles(
...
-  const changedPackageFiles = tracedFiles.filter(isPackageFile);
-  if (changedPackageFiles.length) ctx.turboSnap.bailReason = { changedPackageFiles };
+  const changedPackageFiles = tracedFiles.filter(isPackageLockFile);
+  if (changedPackageFiles.length) {
+    ctx.turboSnap.bailReason = { changedPackageFiles };
+    // If package.json dependencies changed, we still want to use the same TurboSnap bail reason
+    // for now.
+  } else if (ctx.git.packageManifestsWithDependencyChanges?.length) {
+    ctx.turboSnap.bailReason = {
+      changedPackageFiles: ctx.git.packageManifestsWithDependencyChanges,
```
skitterm, line 175: I didn't want to add a new bail reason to the data model since we'll likely soon be avoiding bailing for package.json changes altogether.

bin-src/tasks/gitInfo.ts
```diff
@@ -175,6 +176,17 @@ export const setGitInfo = async (ctx: Context, task: Task) => {
...
+      const changedPackageManifests = ctx.git.changedFiles.filter((fileName) =>
+        [/^package\.json$/, /\/package\.json$/].some((re) => re.test(fileName))
+      );
+
+      if (changedPackageManifests.length > 0) {
+        ctx.git.packageManifestsWithDependencyChanges = await getPackageManagerChanges(
+          ctx.build,
+          changedPackageManifests
+        );
+      }
```
skitterm, line 189: @ghengeveld should I be looking over all the baseline commits here (instead of just the most recent), similar to our approach today?

bin-src/lib/comparePackageManifests.test.ts
```diff
@@ -0,0 +1,184 @@
...
+  ).toBe(true);
+});
+
+it('returns false if dependency objects are added', () => {
+  expect(arePackageDependenciesEqual({}, { dependencies: { a: '1' } })).toBe(false);
+});
+
+it('returns false if dependency objects are removed', () => {
+  expect(arePackageDependenciesEqual({ dependencies: { a: '1' } }, {})).toBe(false);
+});
```
ghengeveld, line 184: One more scenario: moving a dep from `dependencies` to `devDependencies`, all else staying equal.
We could allow this because no versions are changing (the same packages end up getting installed), but it might cause issues if devDeps explicitly aren't installed (e.g. in CI) when they build their Storybook, in which case there might actually be visual changes as a result. So probably safest to just bail in this scenario.

bin-src/tasks/gitInfo.ts
ghengeveld: Good point. What we should be doing is check against the baseline build where the manifest file was actually changed. We [currently take the union](https://github.com/chromaui/chromatic-cli/blob/83f281d2cdcc05df6fa5db97bd664b5ba819a65b/bin-src/tasks/gitInfo.ts#L157-L163) of changed files from each baseline as the total set of `changedFiles`. If `package.json` is in there, we don't know whether it was changed on one baseline, the other, or both.

I suggest we introduce a new `ctx.git.changedPackageManifestsByCommit` value which contains a Map of commit -> changed manifest files. We'd create that in the same [place](https://github.com/chromaui/chromatic-cli/blob/83f281d2cdcc05df6fa5db97bd664b5ba819a65b/bin-src/tasks/gitInfo.ts#L157-L163) where we set `ctx.git.changedFiles` right now.

bin-src/ui/tasks/upload.ts
```diff
@@ -50,10 +50,18 @@ export const tracing = (ctx: Context) => {
...
   const { changedPackageFiles, changedStorybookFiles, changedStaticFiles } =
     ctx.turboSnap.bailReason;
-  const [firstFile, ...otherFiles] =
-    changedPackageFiles || changedStorybookFiles || changedStaticFiles;
+  const changedFiles = changedPackageFiles || changedStorybookFiles || changedStaticFiles;
+
+  // if all changed files are package.json, message this as a dependency change.
+  const allChangedFilesArePackageJson = changedFiles.every(
+    (changedFile) => changedFile === 'package.json'
+  );
```
ghengeveld, line 58: Maybe we should extract this into a util (including the regex thing).

ghengeveld (changes requested):

bin-src/lib/comparePackageManifests.test.ts
skitterm: Sounds good, shouldn't need a behavior change, I'll add a test!

bin-src/tasks/gitInfo.ts
```diff
@@ -175,6 +176,17 @@ export const setGitInfo = async (ctx: Context, task: Task) => {
             .join('\n')}`
         );
       }
+
+      const changedPackageManifests = ctx.git.changedFiles.filter((fileName) =>
+        [/^package\.json$/, /\/package\.json$/].some((re) => re.test(fileName))
+      );
+
+      if (changedPackageManifests.length > 0) {
+        ctx.git.packageManifestsWithDependencyChanges = await getPackageManagerChanges(
```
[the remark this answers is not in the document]
skitterm: `getPackageManagerChanges` is defined outside `setGitInfo`, so it'd be defined before `setGetInfo` gets called.

bin-src/tasks/gitInfo.ts
skitterm: I went ahead and accounted for multiple baselines. I'm only storing the file names that changed (not the commits), since we only need to know that there are some package files changed (at the `gitInfo` level I'm caring about commits of course, so we can checkout the files to diff them).

bin-src/lib/comparePackageManifests.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (keyA !== keyB || valueA !== valueB) {
+      return false;
+    }
+  }
+
+  return true;
+};
+
+export const arePackageDependenciesEqual = (packageObjA, packageObjB) => {
+  const fields = ['dependencies', 'devDependencies', 'peerDependencies'];
```
yannbf, line 32: I know you mentioned about peerDependenciesMeta and resolutions, but I wanted to point possible important fields that are not here:
- optionalDependencies
- overrides (npm 8)
- resolutions (yarn)

yannbf:
FYI the unit tests are failing:

```
bin-src/lib/getDependentStoryFiles.test.ts (9.802 s)
  ● getDependentStoryFiles › detects direct changes to CSF files

    TypeError: Cannot read properties of undefined (reading 'changedPackageManifests')

      189 |     // If package.json dependencies changed, we still want to use the same TurboSnap bail reason
      190 |     // for now.
    > 191 |   } else if (ctx.git.changedPackageManifests?.length) {
          |                      ^
      192 |     ctx.turboSnap.bailReason = { changedPackageFiles: ctx.git.changedPackageManifests };
      193 |   }
      194 |

      at getDependentStoryFiles (bin-src/lib/getDependentStoryFiles.ts:191:22)
      at Object.<anonymous> (bin-src/lib/getDependentStoryFiles.test.ts:51:17)
```

bin-src/tasks/gitInfo.ts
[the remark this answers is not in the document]
ghengeveld: That's function hoisting, which only applies when using the `function` keyword, not when using an arrow function.
Although I personally like the "chronological" order of writing code like that (defining functions _after_ they are used, so implementation details are further down in the file), it's generally frowned upon.

bin-src/tasks/gitInfo.ts
```diff
@@ -175,6 +209,16 @@ export const setGitInfo = async (ctx: Context, task: Task) => {
             .join('\n')}`
         );
       }
+
+      const packageManifestDependencyChanges = await Promise.all(
+        packageManifestChanges.map(async ({ commit: buildCommit, changedFiles }) => {
+          return getPackageManagerChanges(buildCommit, changedFiles);
```
ghengeveld, line 215:
```suggestion
          return getPackageManagerChanges(ctx, buildCommit, changedFiles);
```

bin-src/tasks/gitInfo.ts
```diff
@@ -51,6 +52,23 @@ interface LastBuildQueryResult {
...
+      const fileA = await execGitCommand(`git show ${commit}:${fileName}`);
+      const fileB = await execGitCommand(`git show HEAD:${fileName}`);
+
+      // put in empty entry for equal-dependency packages so we only have fileNames for the non-equal ones
+      return arePackageDependenciesEqual(JSON.parse(fileA), JSON.parse(fileB)) ? [] : [fileName];
+    })
+  );
+
+  return allChanges.flat();
+};
```
ghengeveld, line 70: This function could use some tests.

bin-src/tasks/gitInfo.ts
```diff
@@ -175,6 +209,16 @@ export const setGitInfo = async (ctx: Context, task: Task) => {
...
+
+      const packageManifestDependencyChanges = await Promise.all(
+        packageManifestChanges.map(async ({ commit: buildCommit, changedFiles }) => {
+          return getPackageManagerChanges(buildCommit, changedFiles);
+        })
+      );
+
+      const flattenedChanges = packageManifestDependencyChanges.flat();
+      // remove duplicate entries (if have multiple ancestors and both changed the same package.json, for example)
+      ctx.git.changedPackageManifests = Array.from(new Set(flattenedChanges));
```
ghengeveld, line 221: Probably a good idea to extract all of this to a separate function (and file), along with the other stuff (L183-196) above and the `getPackageManagerChanges` function. I'm fine if you then test it as a whole rather than test `getPackageManagerChanges` separately.

bin-src/ui/messages/warnings/bailFile.ts
```diff
@@ -9,9 +10,17 @@ const docsUrl = 'https://www.chromatic.com/docs/turbosnap#how-it-works';
...
-  let type = changedPackageFiles ? 'package' : 'static';
+  // if all changed files are package.json, message this as a dependency change.
+  const allChangedFilesArePackageJson = changedFiles.every((changedFile) =>
+    isPackageManifestFile(changedFile)
+  );
+
+  const [firstFile, ...files] = changedFiles;
+
+  let type = changedPackageFiles ? 'package file' : 'static file';
+  if (allChangedFilesArePackageJson) type = 'dependency';
```
ghengeveld, line 23: 👍

ghengeveld (changes requested):
Looking good. Needs a bit of cleanup/refactor to keep our sanity though.

bin-src/lib/getDependentStoryFiles.test.ts
```diff
@@ -356,11 +357,40 @@ describe('getDependentStoryFiles', () => {
...
     expect(res).toEqual(null);
+    expect(ctx.turboSnap.bailReason).toEqual({
+      changedPackageFiles: ['src/yarn.lock'],
+    });
+    expect(ctx.log.warn).toHaveBeenCalledWith(
+      expect.stringContaining(chalk`Found a package file change in {bold src/yarn.lock}`)
+    );
+  });
+
+  it('bails on dependency changes to package manifest file', async () => {
```
tmeasday, line 368: Should we add a test for *not* bailing on non-dependency changes?

tmeasday (approved):
Looks great @skitterm! I have just one more suggestion for a further test.

PR #649, opened 2022-10-07
Dealing with URL Params in stats file module names

thafryer:
If the name of a module in the Webpack Stats file has a URL parameter (e.g. ?ngResource) in it, it causes an issue with tracing the file. This PR resolves the issue.

### DIFF
--- a/bin-src/lib/getDependentStoryFiles.ts
+++ b/bin-src/lib/getDependentStoryFiles.ts
@@ -70,8 +70,8 @@ export async function getDependentStoryFiles(
   // Convert a "webpack path" (relative to storybookBaseDir) to a "git path" (relative to repository root)
   // e.g. `./src/file.js` => `path/to/storybook/src/file.js`
   const normalize = (posixPath: string) => {
-    const CSF_REGEX = /.\/src.*?sync/g;
-    const URL_PARAM_REGEX = /(\?.*)|(#.*)/g;
+    const CSF_REGEX = /\s+sync\s+/g;
+    const URL_PARAM_REGEX = /(\?.*)/g;
     let newPath = normalizePath(posixPath, rootPath, baseDir);
     // This regex test is to ensure file names do not include url parmeters
     // or match the CSF glob we get back in the stats file. We added this because
--- a/bin-src/lib/getDependentStoryFiles.ts
+++ b/bin-src/lib/getDependentStoryFiles.ts
@@ -99,6 +99,10 @@ export async function getDependentStoryFiles(
     .filter((mod) => isUserModule(mod))
     .forEach((mod) => {
       let { name } = mod;
+      // This regex test is to ensure file names do not include url parmeters
+      // or match the CSF glob we get back in the stats file. We added this because
+      // CSS/SCSS files were getting ?ngResource appended on the end of file names
+      // in the stats file.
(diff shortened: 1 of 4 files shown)


bin-src/lib/getDependentStoryFiles.ts
```diff
@@ -93,17 +93,28 @@ export async function getDependentStoryFiles(
...
   const csfGlobsByName: Record<string, true> = {};
-
+  const CSF_REGEX = /.\/src.*?sync/g;
+  const URL_PARAM_REGEX = /(\?.*)|(#.*)/g;
   stats.modules
     .filter((mod) => isUserModule(mod))
     .forEach((mod) => {
-      const normalizedName = normalize(mod.name);
+      let { name } = mod;
+      if (URL_PARAM_REGEX.test(mod.name) && !CSF_REGEX.test(mod.name)) {
```
thafryer, line 102: We need this `CSF_REGEX` test because the stats file generates a very weird blob for CSF files which seemed to have this "./src ... sync"and the  ? and # characters, which cause the file glob to also match the URL parameter test.

bin-src/lib/getDependentStoryFiles.ts
```diff
@@ -93,17 +93,28 @@ export async function getDependentStoryFiles(
   const namesById: Record<number, string> = {};
   const reasonsById: Record<number, string[]> = {};
   const csfGlobsByName: Record<string, true> = {};
-
+  const CSF_REGEX = /.\/src.*?sync/g;
+  const URL_PARAM_REGEX = /(\?.*)|(#.*)/g;
```
tmeasday, line 97: I'm not sure that webpack "urls" can have hashes in them so I'd be inclined to keep it simple and leave hashes out.

bin-src/lib/getDependentStoryFiles.ts
```diff
@@ -93,17 +93,28 @@ export async function getDependentStoryFiles(
   const namesById: Record<number, string> = {};
   const reasonsById: Record<number, string[]> = {};
   const csfGlobsByName: Record<string, true> = {};
-
+  const CSF_REGEX = /.\/src.*?sync/g;
```
tmeasday, line 96: maybe `\s+sync` -- I suspect the whitespace in the string is crucial (and not allowed in normal "urls").

The `./src` part will be project specific so I'd leave that off.

tmeasday (changes requested):
Can we move the logic inside `normalize` so we don't need to repeat it?

bin-src/lib/getDependentStoryFiles.ts
```diff
@@ -69,7 +69,20 @@ export async function getDependentStoryFiles(
 
   // Convert a "webpack path" (relative to storybookBaseDir) to a "git path" (relative to repository root)
   // e.g. `./src/file.js` => `path/to/storybook/src/file.js`
-  const normalize = (posixPath: string) => normalizePath(posixPath, rootPath, baseDir);
+  const normalize = (posixPath: string) => {
+    const CSF_REGEX = /\s+sync\s+/g;
+    const URL_PARAM_REGEX = /(\?.*)/g;
+    let newPath = normalizePath(posixPath, rootPath, baseDir);
+    // This regex test is to ensure file names do not include url parmeters
```
tmeasday, line 76:
```suggestion
    // This regex test is to ensure file names do not include url parameters
```

tmeasday (approved):
LGTM!

PR #650, opened 2022-10-10
Ensure all GitHub Action outputs are exposed

ghengeveld:
This avoids dumb mistakes by automatically exposing all properties on Chromatic's return value in GitHub Actions.


JReinhold (approved):
Nice solution!
LGTM

PR #660, opened 2022-10-19
Update action to use node16

bryanjos:
Node 12 actions are deprecated. This update the node version used to `node16`


thafryer:
Thanks @bryanjos for your contribution. It is very much appreciated!

PR #667, opened 2022-10-26
Edit package file detection process for the untraced flag

ethriel3695:
A regression was introduced with the untraced flag with the updates from 6.11.0

We previously had a `GLOBALS` const that included both root-level `package.json` files and nested `package-json` files in subdirectories. This was removed so non-dependency changes to package.jsons don't cause full rebuilds. However, we weren't making sure untraced package.jsons didn't cause full rebuilds in this case.

Here is the CLI output from my local chromatic run

```
Starting partial build
    → Snapshots will be limited to 0 story files affected by recent changes
✔ TurboSnap enabled
Capturing 0 snapshots and skipping 224 snapshots.
Started build 6915
    → View build details at https://www.chromatic.com/build?appId=******&number=6915
Running 0 tests affected by recent changes (skipping 224 tests)
    → This may take a few minutes
```

### DIFF
--- a/bin-src/lib/getDependentStoryFiles.ts
+++ b/bin-src/lib/getDependentStoryFiles.ts
@@ -191,6 +191,7 @@ export async function getDependentStoryFiles(
     ctx.turboSnap.bailReason = { changedPackageFiles };
     // If package.json dependencies changed, we still want to use the same TurboSnap bail reason
     // for now.
+    // if package.jsons are untraced, don't bail, even when its dependency fields have changed
   } else if (ctx.git.changedPackageManifests?.length && tracedFiles.filter(isPackageFile).length) {
     ctx.turboSnap.bailReason = { changedPackageFiles: ctx.git.changedPackageManifests };
   }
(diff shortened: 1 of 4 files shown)


bin-src/lib/getDependentStoryFiles.ts
```diff
@@ -188,7 +191,7 @@ export async function getDependentStoryFiles(
     ctx.turboSnap.bailReason = { changedPackageFiles };
     // If package.json dependencies changed, we still want to use the same TurboSnap bail reason
     // for now.
-  } else if (ctx.git.changedPackageManifests?.length) {
+  } else if (ctx.git.changedPackageManifests?.length && tracedFiles.filter(isPackageFile).length) {
```
skitterm, line 194:
```suggestion
    // if package.jsons are untraced, don't bail, even when its dependency fields have changed  
  } else if (ctx.git.changedPackageManifests?.length && tracedFiles.filter(isPackageFile).length) {
```

skitterm (commented):
Looks good @ethriel3695 , thanks for catching this.

skitterm (approved):

PR #670, opened 2022-10-31
Trace command errors for package manifest change

skitterm:
In #648 we started looking within package manifest files and only bailing from TurboSnap if the change was dependency-related. However the `trace` command was failing because we were looking for info that didn't exist. 

Now we throw an error if a person attempts to trace a package.json file, since we'd need to know the file contents (and not just the file path) to know if that file would cause changes to story files.

New error message:
Screen Shot 2022-10-31 at 4 38 24 PM

### DIFF
--- a/bin-src/trace.ts
+++ b/bin-src/trace.ts
@@ -92,7 +92,7 @@ export async function main(argv: string[]) {
   const packageManifestFile = changedFiles.find((item) => isPackageManifestFile(item));
   if (packageManifestFile) {
     throw new Error(
-      'Unable to trace package manifest file as that would require diffing file contents.'
+      `Unable to trace package manifest file (${packageManifestFile}) as that would require diffing file contents.`
     );
   }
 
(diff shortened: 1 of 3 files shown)


bin-src/trace.ts
```diff
@@ -88,5 +89,12 @@ export async function main(argv: string[]) {
   const stats = await readStatsFile(flags.statsFile);
   const changedFiles = input.map((f) => f.replace(/^\.\//, ''));
 
+  const packageManifestFile = changedFiles.find((item) => isPackageManifestFile(item));
+  if (packageManifestFile) {
+    throw new Error(
+      'Unable to trace package manifest file as that would require diffing file contents.'
```
tmeasday, line 95:
```suggestion
      'Unable to trace package manifest file (`${packageManifestFile}`) as that would require diffing file contents.'
```

tmeasday (approved):
LGTM! Small suggestion for improvement

PR #675, opened 2022-11-10
Output progress updates in non-interactive mode every 10 seconds

ghengeveld:
In non-interactive mode (i.e. a CI environment), we wouldn't print any progress status updates during the verify and upload progress. This could lead to the CI script failing for lack of output over time ([example](https://app.circleci.com/pipelines/github/chromaui/chromatic/18053/workflows/5a325be5-caf7-4b24-ad8e-00d4b9fe8fd9/jobs/166735)).

> Too long with no output (exceeded 10m0s): context deadline exceeded

The exact timeout differs per CI environment (this is CircleCI).

This PR ensures we also print the usual progress bar in non-interactive mode, but throttled to 10 seconds in order to avoid flooding the logs. For long-running processes that have no progress indicator (build-storybook, verify), render a "bouncing" activity bar. This should resolve the script timeout issue, provide useful progress updates and keep the number of log lines to a minimum.

The `CHROMATIC_OUTPUT_INTERVAL` environment variable can be used to override the output interval (in ms). Default is 10,000 (10 seconds). Documentation update is [here](https://github.com/chromaui/chromatic-docs/pull/185).

Screen Shot 2022-12-09 at 21 05 13


jmhobbs (approved):
LGTM

PR #676, opened 2022-11-14
Log message when TurboSnap is unavailable

ghengeveld:
Screen Shot 2022-11-14 at 15 06 52

Depends on https://github.com/chromaui/chromatic/pull/6730


ghengeveld:
This can be safely released before the feature gating is deployed, because we're not querying any new fields (`turboSnapEnabled` is already available).

tmeasday (approved):
LGTM

PR #683, opened 2022-11-25
Enhanced TurboSnap: Trace dependency changes instead of bailing out

ghengeveld:
Currently (as of v6.11.0), TurboSnap bails out if it finds a dependency-related change in `package.json`, or _any_ change to `package-lock.json` or `yarn.lock`. Although this is much better than bailing out on _any_ change to `package.json`, this behavior is still lacking because it means we do a full rebuild even if the updated dependency isn't used in any or some of the stories (directly or indirectly).

This update avoids bailing out of TurboSnap for changes to lock files. Instead it parses and compares lock files to their previous (baseline) version, to retrieve a set of **changed dependency names**. Those dependency names are then cross-referenced with the Webpack dependency tree to trace those changed dependencies to the story files that are affected by them. This essentially uses the existing "changed files" behavior of TurboSnap, but extends it to include files in `node_modules` (for dependencies we know have been added/updated/removed) in the set of changed files.

If we fail to find or parse a lock file (or its previous version) then we fall back to the v6.11.0 behavior of bailing out on any dependency-related change to package.json.

### How to test

Scenarios to test:
- [x] Component change => tests are skipped as usual (except for the changed component)
- [x] npm dependency update (package-lock.json change) => only tests stories which use this dependency
- [x] Yarn dependency update (yarn.lock change) => only tests stories which use this dependency
- [x] package.json dependency change (lockfile missing) => bails out with `changedPackageFiles`
- [x] package.json other change (lockfile missing) => skips all tests
- [x] package.json dependency change (lockfile unchanged) => skips all tests
- [x] Monorepo setup (Storybook / package.json in subdirectory):
	- [x] Component change => tests are skipped as usual (except for the changed component)
	- [x] Dependency change in Storybook project (local lockfile change) => skips tests that don't use this dependency
	- [x] Dependency change in another project (root lockfile change) => skips tests that don't use this dependency


bin-src/tasks/gitInfo.ts
```diff
@@ -150,44 +146,54 @@ export const setGitInfo = async (ctx: Context, task: Task) => {
...
+      ctx.git.changedFiles = Array.from(
+        new Set(changedFilesWithInfo.flatMap(({ changedFiles }) => changedFiles))
       );
-      ctx.git.changedFiles = Array.from(new Set(results.flatMap((r) => r.changedFiles)));
 
-      ctx.git.replacementBuildIds = results
+      // Track changed package manifest files along with the commit they were changed in.
+      ctx.git.packageManifestChanges = changedFilesWithInfo.flatMap(({ build, changedFiles }) => {
+        const manifestFiles = changedFiles.filter(isPackageManifestFile);
+        return manifestFiles.length ? [{ changedFiles: manifestFiles, commit: build.commit }] : [];
```
tmeasday, line 174: I think you'd want to use the `replacementBuild.commit` if it was set over the `build.commit`.

bin-src/lib/findChangedDependencies.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (acc.has(value)) acc.delete(value);
+    else acc.add(value);
+    return acc;
+  }, new Set(left));
+
+// Yields a list of dependency names which have changed since the baseline.
+// E.g. ['react', 'react-dom', '@storybook/react']
+export const findChangedDependencies = async (baselineCommits: string[]) => {
+  const manifestName = 'package.json';
+  const lockfileName = hasYarn() ? 'yarn.lock' : 'package-lock.json';
```
tmeasday, line 26: Are we going to bail in the case of `pnpm` etc?

bin-src/tasks/upload.ts
```diff
@@ -153,10 +155,28 @@ export const traceChangedFiles = async (ctx: Context, task: Task) => {
   transitionTo(tracing)(ctx, task);
 
   const statsPath = join(ctx.sourceDir, ctx.fileInfo.statsPath);
-  const { changedFiles } = ctx.git;
+  const { baselineCommits, changedFiles, packageManifestChanges } = ctx.git;
   try {
+    const changedDependencyNames = await findChangedDependencies(baselineCommits).catch((e) => {
+      ctx.log.debug(e);
+    });
```
tmeasday, line 162: Would it make sense to also figure out the `packageLockChanges` ahead of time and only diff those commits? [Similarly we'd need to use replacement commits when the commit was missing]

bin-src/git/git.ts
```diff
@@ -211,6 +211,15 @@ export async function checkout(ref: string) {
   return execGitCommand(`git checkout ${ref}`);
 }
 
+export async function checkoutFile(
+  ref: string,
+  fileName: string,
+  targetFileName = `${ref}.${fileName}`
```
tmeasday, line 217: Shouldn't this be in a temp dir or something?

tmeasday (changes requested):
This is awesome @ghengeveld! 🚀 

I just want to make sure we get the replacement builds thing right. Is it possible to get a test for those in? I'm not really seeing a full E2E test for this, is that too difficult?

bin-src/lib/findChangedDependencies.ts
ghengeveld: Yes, pnpm and Yarn pnp is out of scope.

bin-src/git/git.ts
ghengeveld: Ah yes, of course 👍 Wouldn't want to clobber up their working dir.

bin-src/lib/findChangedDependencies.ts
tmeasday: Sure, but we should bail out at least if we see a lockfile change for one of those.

skitterm (commented):
Nice work, thanks for taking this on. I assume we'll want to also (not in this PR) change the [TurboSnap docs](https://www.chromatic.com/docs/turbosnap#full-rebuilds) related to full rebuilds.

bin-src/lib/findChangedDependencies.ts
ghengeveld: I think this will work out because we wouldn't be able to find any node_modules in the stats file (because they're located elsewhere), and we [already have a mechanism](https://github.com/chromaui/chromatic-cli/blob/feb1b5e87942d0bf76d6f3a4304c8b26d56422b5/bin-src/lib/getDependentStoryFiles.ts#L205-L213) to bail in case we don't find any dependency files in the stats.

bin-src/tasks/upload.ts
```diff
@@ -153,10 +155,28 @@ export const traceChangedFiles = async (ctx: Context, task: Task) => {
...
+    if (!changedDependencyNames) {
```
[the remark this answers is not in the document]
ghengeveld: No, an empty array is a valid response (no dependencies changed). It can be falsy (`undefined`) if it hits the `catch` clause.

bin-src/lib/findChangedDependencies.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  const headTree = await buildDepTreeFromFiles('.', manifestName, lockfileName, true);
```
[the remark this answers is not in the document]
ghengeveld: Good call. We need to handle that.

bin-src/tasks/upload.ts
ghengeveld: Are you suggesting we run `findChangedDependencies` earlier on in the process, or are you suggesting to determine what baselines have lockfile changes, and feed that into `findChangedDependencies` so we don't have to check _all_ baselines?

tmeasday: The latter. I only suggested it because we do that optimisation for package file changes.

ghengeveld: Going to punt on that for now. This PR has been dragging on for too long as it is.

bin-src/lib/getDependentStoryFiles.test.ts
```diff
@@ -340,8 +373,8 @@ describe('getDependentStoryFiles', () => {
     ).rejects.toEqual(new Error('Did not find any CSF globs in preview-stats.json'));
   });
 
-  it('bails on changed global file', async () => {
-    const changedFiles = ['src/foo.stories.js', 'src/yarn.lock'];
+  it('does not bail on changed global file', async () => {
```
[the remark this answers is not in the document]
ghengeveld: Done. They're in `upload.test.ts`.

skitterm (approved):
Looks great to me, thanks for all the comments in the tests!

ghengeveld:
All tested ✅ Wow!
I used the `shapes-light` and `chromatic` projects.

PR #694, opened 2022-12-09
Remove tunnel flags

paulelliott:
This cleans out all the deprecated flags related to tunnel builds.

Fixes IN-712


paulelliott:
@ghengeveld @tmeasday Can one of you smoke test these changes locally?

bin-src/lib/getOptions.ts
```diff
@@ -181,70 +150,28 @@ export default function getOptions({ argv, env, flags, log, packageJson }: Conte
...
-  }
-
-  return {
-    ...options,
-    noStart,
-    useTunnel: true,
-    url: storybookUrl,
-    scriptName,
-  };
+  return options;
```
ghengeveld, line 176: These lines are unreachable after the `throw` above.

The check for `onlyChanged` was there because it's incompatible with starting Storybook rather than using a built Storybook. We can remove this check because we already check for the existence of the stats file and will just bail with `missingStatsFile` if it doesn't exist, and continue running the build regardless.

bin-src/lib/getOptions.ts
```diff
@@ -181,70 +150,28 @@ export default function getOptions({ argv, env, flags, log, packageJson }: Conte
...
-          ) || [];
-        if (key) buildScriptName = key;
-      }
-    }
-    if (scripts && buildScriptName && scripts[buildScriptName]) {
-      return { ...options, noStart: true, useTunnel: false, buildScriptName };
+  // Build Storybook
+  if (storybookBuildDir) {
+    return { ...options };
+  }
```
ghengeveld, line 156:
```suggestion
  // Use prebuilt Storybook
  if (storybookBuildDir) {
    return options;
  }

  // Look for a build-storybook script to build Storybook ourselves
```

bin-src/tasks/index.ts
```diff
@@ -24,25 +22,15 @@ export const runUploadBuild = [
...
-  verify,
-  snapshot,
-];
 
-export const runPatchBuild = (runBuild: typeof runUploadBuild | typeof runTunnelBuild) => [
+export const runPatchBuild = (runBuild: typeof runUploadBuild) => [
   prepareWorkspace,
   ...runBuild,
   restoreWorkspace,
 ];
```
ghengeveld, line 30:
```suggestion
export const runPatchBuild = [
  prepareWorkspace,
  ...runUploadBuild,
  restoreWorkspace,
];
```

bin-src/tasks/index.ts
```diff
@@ -24,25 +22,15 @@ export const runUploadBuild = [
...
 
 export default (options: Context['options']): Listr.ListrTask<Context>[] => {
-  const runBuild = options.useTunnel ? runTunnelBuild : runUploadBuild;
-  const tasks = options.patchHeadRef && options.patchBaseRef ? runPatchBuild(runBuild) : runBuild;
+  const tasks =
+    options.patchHeadRef && options.patchBaseRef ? runPatchBuild(runUploadBuild) : runUploadBuild;
```
ghengeveld, line 34:
```suggestion
  const tasks =
    options.patchHeadRef && options.patchBaseRef ? runPatchBuild : runUploadBuild;
```

ghengeveld (approved):
Nice!

tmeasday:
🚀

PR #695, opened 2022-12-12
Omit `inputs` for `workflow_dispatch` event in GitHub Action

108EAA0A:
The support for `workflow_dispatch` added in #311, but it is a pain to set `inputs` manually.
This should be possible to automate, so I modified the implementation.


arikmaor (approved):
LGTM

thafryer (approved):
Looks good. Tested this and worked with no issue. I will update you once a new CLI version is released.

PR #706, opened 2023-01-11
Add JUnit Report to Github Action Inputs

thafryer:


thafryer:
@ghengeveld Let me know when https://github.com/chromaui/chromatic-cli/pull/710 is ready and I will pull in changes from your PR.

ghengeveld:
@thafryer It's merged.

I think there may be a reason we don't support the XML report on the GitHub Action. The usual recommendation with GitHub is to use exitOnceUploaded, but that wouldn't work with the XML report because that requires one to run the CLI to the very end. That's not a reason to omit the option though, so I wonder if there's another reason. Either way this requires proper QA.

ghengeveld:
I see no other reason not to support this besides exiting late, so this is OK with me. Needs QA though.

PR #710, opened 2023-01-20
Add `--repository-slug` flag to CLI, and `repositorySlug` + `branchName` options to Action

ghengeveld:
This allows customers to override the repository owner and/or name through either the `repositorySlug` flag/option, or the `branchName` flag/option, whichever is more convenient. Either of them can affect the `ownerName`, which is used to determine whether a build originates from a fork.

We didn't previously expose the `branchName` option on the GitHub Action, now we do.

For example:
- `repositorySlug: "chromaui/chromatic"` sets the owner name to `chromaui` and repository name to `chromatic`
- `branchName: "chromaui:feature"` sets the owner name to `chromaui` and the branch name to `feature`.

If both specify an owner name, they must match.

### DIFF
--- a/action-src/main.ts
+++ b/action-src/main.ts
@@ -136,30 +136,30 @@ async function run() {
 
   try {
     // Remember to keep this list in sync with ../action.yml
-    const projectToken = getInput('projectToken') || getInput('appCode'); // backwards compatibility
-    const workingDir = getInput('workingDir') || getInput('workingDirectory');
+    const allowConsoleErrors = getInput('allowConsoleErrors');
+    const autoAcceptChanges = getInput('autoAcceptChanges');
     const buildScriptName = getInput('buildScriptName');
-    const skip = getInput('skip');
+    const debug = getInput('debug');
+    const diagnostics = getInput('diagnostics');
     const dryRun = getInput('dryRun');
+    const exitOnceUploaded = getInput('exitOnceUploaded');
+    const exitZeroOnChanges = getInput('exitZeroOnChanges');
+    const externals = getInput('externals');
     const forceRebuild = getInput('forceRebuild');
-    const storybookBaseDir = getInput('storybookBaseDir');
-    const storybookConfigDir = getInput('storybookConfigDir');
+    const ignoreLastBuildOnBranch = getInput('ignoreLastBuildOnBranch');
     const only = getInput('only');
     const onlyChanged = getInput('onlyChanged');
-    const onlyStoryNames = getInput('onlyStoryNames');
     const onlyStoryFiles = getInput('onlyStoryFiles');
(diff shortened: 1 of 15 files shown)


bin-src/lib/getOptions.ts
```diff
@@ -32,7 +27,8 @@ const undefinedIfEmpty = <T>(array: T[]) => {
 export default function getOptions({ argv, env, flags, log, packageJson }: Context): Options {
   const fromCI = !!flags.ci || !!process.env.CI;
   const [patchHeadRef, patchBaseRef] = (flags.patchBuild || '').split('...').filter(Boolean);
-  const [branchName, ownerName] = (flags.branchName || '').split(':').reverse();
+  const [branchName, branchOwner] = (flags.branchName || '').split(':').reverse();
+  const [repositoryOwner] = flags.repositorySlug ? flags.repositorySlug.split('/') : [];
```
JonathanKolnik, line 31: Can we write a test for this?

JonathanKolnik: and maybe how it relates to which one takes precedence on line 67

thafryer (approved):
This looks good. I noticed we have a few more flags missing in the GitHub Action, like the JUnit report one. We should formalize that anytime we add an option to the CLI we add it to the Action unless circumstances say we shouldn't.

bin-src/lib/getOptions.ts
ghengeveld: Done

action-src/main.ts
```diff
@@ -136,68 +136,73 @@ async function run() {
 
   try {
     // Remember to keep this list in sync with ../action.yml
```
itwasmattgregg, line 138: 👍

action.yml
```diff
@@ -6,35 +6,44 @@ branding:
   color: 'orange'
 
 inputs:
```
itwasmattgregg, line 8: Would be sweet if this auto-generated the docs page.

itwasmattgregg (approved):
This all looks good. Curious though what happens if you specify `repositorySlug` with a different owner than what gets generated from `branchName` as the owner.

action.yml
ghengeveld: Our docs are based on Markdown, and live in a different repository. It's definitely possible, but not really worth the effort and complexity. We don't change options all that often.

ghengeveld:
> This all looks good. Curious though what happens if you specify `repositorySlug` with a different owner than what gets generated from `branchName` as the owner.

There's a check to verify that if both are specified (and `branchName` contains an owner name prefix), they must match. Otherwise an error is thrown and no build will run.

JonathanKolnik (approved):

PR #718, opened 2023-02-23
Update changedFiles target to fix broken trace

ethriel3695:
Currently when we run the trace utility to debug builds, we get an error on `changedFiles` within the `getDependentStoryFiles` function.

This PR changes the reference of changedFiles from ctx.git.changedFiles to changedFiles which is passed to the function


tmeasday (approved):
LGTM. Would be great to get a test of the trace so it doesn't break again, but doesn't have to be in this PR.

thafryer (approved):

PR #730, opened 2023-03-29
Disconnect Preserve missing from `--only-story-names` flag

thafryer:
Only Story Names was using preserve missing in the background. This updates the CLI so that the CLI flag no causes preserveMissingSpecs to be set.


ethriel3695 (approved):
Looks great @thafryer

PR #733, opened 2023-04-04
Chromatic telemetry reports null safeguard

JonathanKolnik:


bin-src/lib/getStorybookMetadata.ts
```diff
@@ -170,12 +170,35 @@ export const getStorybookMetadata = async (ctx: Context) => {
...
-    findConfigFlags(ctx),
-    findViewlayer(ctx),
-    findBuilder(mainConfig),
-  ]);
-  ctx.log.debug(info);
-  return info.reduce((acc, obj) => Object.assign(acc, obj), {});
+  let addons = {};
+  try {
+    addons = await findAddons(ctx, mainConfig);
+  } catch (e) {
```
thafryer, line 176: Would it be helpful to log out the stack trace too? Maybe we should add Sentry to CLI.

thafryer (approved):

bin-src/lib/getStorybookMetadata.ts
JonathanKolnik: adding stack trace now

re: sentry, I think that's a good idea -- I'd probably want to chat with Gert about where it makes sense to inject that into the context etc.

JonathanKolnik:
Closing in favor of #735

Issue #668, opened 2022-10-28
[github action] issue_comment event is not supported in this action


dominikwilkowski:
**Feature request**

We need to run our CI vis test on a comment that is being added to a PR.
We have a CI workflow that does just this https://github.com/Shopify/polaris/blob/main/.github/workflows/vistest.yml but it fails with the above mentioned error message

This fails right now: https://github.com/Shopify/polaris/actions/runs/3342136481/jobs/5534047760#step:10:12

**Proposed solution**

Please add support for the `issue_comment` event

JReinhold:
I think that's doable, but the caveat is that the `issue_comment` event is always invoked from the default branch, so even if someone comments on a PR, the Chromatic CLI would act as if it was an event in the default branch. This might be a bit confusing, and could lead to users getting baseline builds when they expected regular PR builds.

We could however just assume that users know what they're doing, and alternatively provide a warning in that case.

cc @thafryer 

## Additional info

It would be a matter of adding a switch case to `getBuildInfo` https://github.com/chromaui/chromatic-cli/blob/main/action-src/main.ts#L26

Docs: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment

Example payloads: https://github.com/octokit/webhooks/tree/master/payload-examples/api.github.com/issue_comment

alex-page:
I think `issue_comment` uses the `default branch` in issues. When it is triggered in a pull request you can pass through the branch. Is this a chromatic problem?

dominikwilkowski:
Yes we have been using the `issue_comment` event company wide and it triggers on each PR on main branches or non-main branches. There may be complexities on chromatic side here but from an event perspective this works like a charm for us

codykaup:
Thanks for the suggestion @dominikwilkowski!

I went ahead and created an internal ticket to add support for `issue_comment` in this project so we can track it a bit better!

ArtyEmsee:
Any update here?

codykaup:
Sorry for the delay all!

Looking at our internal ticket, there's been some discussion but the holidays may have gotten the best of us! I poked the PM on that team to get things rolling again.

thafryer:
@ArtyEmsee We have released a new version of the action that supports the issue_comment event. You should be able to update your workflow file to add the event without any change to the Chromatic GitHub Action. Let us know if you run into any more issues.

PR #734, opened 2023-04-05
Add `release` event to GitHub Action

0x2b3bfa0:
In the same vein as #196 and #668, but this time for the `release` event, preventing the following error:

```console
Error: release event is not supported in this action
```

### DIFF
--- a/action-src/main.ts
+++ b/action-src/main.ts
@@ -62,6 +62,13 @@ const getBuildInfo = (event: typeof context) => {
         slug: event.payload.repository.full_name,
       };
     }
+    case 'release': {
+      return {
+        sha: event.sha,
+        branch: event.ref.replace('refs/tags/', ''),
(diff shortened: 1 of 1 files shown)


action-src/main.ts
```diff
@@ -62,6 +62,13 @@ const getBuildInfo = (event: typeof context) => {
         slug: event.payload.repository.full_name,
       };
     }
+    case 'release': {
+      return {
+        sha: event.sha,
+        branch: event.ref.replace('refs/tags/', ''),
```
0x2b3bfa0, line 68: Maybe we should use the tag as the `branch` name instead of the target commitish?

```suggestion
        branch: event.payload.release.target_commitish('refs/tags/', ''),
```

Precha24: Good

Precha24 (commented):
Good

PR #735, opened 2023-04-05
Add framework support to cli sb metadata

JonathanKolnik:
Our telemetry reports storybook_version = null for a large proportion of our build events. It turns out these are SB v7 build events and the `findBuilder` function was erroring out causing the entire metadata Promise.all to fail. This PR updates the logic for finding the builder which should restore the telemetry overall.


thafryer:
@JonathanKolnik, I think you need to rebase from the main branch. We do not have the node 12 job anymore, but we moved it to 14

JonathanKolnik:
> @JonathanKolnik, I think you need to rebase from the main branch. We do not have the node 12 job anymore, but we moved it to 14

Weird it says everything is up to date on my branch

tmeasday (changes requested):
This LGTM but won't this code below still throw:

image

As I read it, in SB7 project, `name` will end up `webpack4`, which isn't installed, leading to the `packageDoesNotExist` error getting thrown.

thafryer:
> This LGTM but won't this code below still throw:
> 
> 
> 
> image
> 
> 
> 
> As I read it, in SB7 project, `name` will end up `webpack4`, which isn't installed, leading to the `packageDoesNotExist` error getting thrown.

In SB7 project, it should get caught by the earlier code block.

In SB6, that will still work since it does the builder package lookup (i.e. `builders[name]`) assuming webpack4 points at a valid package.

tmeasday:
Oh, right! I missed the earlier `return`.

JonathanKolnik:
closing in favor of cleaner PR #738

PR #737, opened 2023-04-06
Better discovery for TurboSnap trace-changed and related directories

ethriel3695:
This adds better tracing for diagnosing TurboSnap tracing in sub-directories

### DIFF
--- a/.github/workflows/smoke-test-yarn-berry.yml
+++ b/.github/workflows/smoke-test-yarn-berry.yml
@@ -5,14 +5,16 @@ jobs:
   chromatic:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/setup-node@v1
+      - uses: actions/checkout@v3
         with:
-          node-version: '14.x'
-      - uses: actions/checkout@v1
+          fetch-depth: 0
+      - uses: actions/setup-node@v3
+        with:
+            node-version: '16'
       - run: yarn install
       - run: yarn set version berry
       - name: run chromatic
-        run: npx -p . chromatic --build-script-name build-test-storybook --exit-zero-on-changes --force-rebuild
+        run: npx -p . chromatic --build-script-name build-test-storybook --exit-zero-on-changes
         env:
           LOG_LEVEL: debug
           DEBUG: chromatic-cli
--- a/.github/workflows/smoke-test-yarn-berry.yml
+++ b/.github/workflows/smoke-test-yarn-berry.yml
@@ -7,10 +7,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          node-version: '16.x'
-      - uses: actions/checkout@v2
-        with:
           fetch-depth: 0
+      - uses: actions/setup-node@v3
+        with:
+            node-version: '16'
       - run: yarn install
       - run: yarn set version berry
       - name: run chromatic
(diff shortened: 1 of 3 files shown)


.github/workflows/smoke-test-yarn-berry.yml
```diff
@@ -5,16 +5,16 @@ jobs:
...
         with:
           fetch-depth: 0
+      - uses: actions/setup-node@v3
+        with:
+            node-version: '16'
       - run: yarn install
       - run: yarn set version berry
       - name: run chromatic
-        run: npx -p . chromatic --build-script-name build-test-storybook --exit-zero-on-changes --force-rebuild
+        run: npx -p . chromatic --build-script-name build-test-storybook --exit-zero-on-changes
```
tevanoff, line 17: Why did you need to remove `force-rebuild`? Won't this be required to rerun on the same commit?

bin-src/ui/messages/info/tracedAffectedFiles.ts
```diff
@@ -38,7 +38,33 @@ export default (
 
   const changed = pluralize('changed files', changedFiles.length, true);
   const affected = pluralize('affected story files', Object.keys(affectedModules).length, true);
-  const summary = chalk`${info} Traced {bold ${changed}} to {bold ${affected}}`;
+  const bailReason = ctx.turboSnap.bailReason
```
tevanoff, line 41: Can this be run in cases where `ctx.turboSnap` will be undefined, or can we guarantee that will be set in here?

.github/workflows/smoke-test-yarn-berry.yml
ethriel3695: I don't love the idea of `--force-rebuild` being in a script all the time, but I agree that if we decide to change it, this should be outside the scope of this PR.

ethriel3695: @tevanoff I reverted the change 🙏🏼

bin-src/ui/messages/info/tracedAffectedFiles.ts
ethriel3695: @tevanoff 
The function call is gated behind traceChanged
https://github.com/chromaui/chromatic-cli/blob/main/bin-src/lib/getDependentStoryFiles.ts#L260

Also, we call `ctx.turbosnap` all over without the `?` conditional operand.

I'd be happy to add it everywhere if you think it's worth it.

tevanoff: Oh ok, all good, just wanted to check. Thanks!

bin-src/trace.test.ts
```diff
@@ -1,18 +1,63 @@
...
+
+    // Execute the script as a child process
+    const output = execSync(`yarn ${scriptName}`).toString().trim();
+
+    const baseDirNote = `The base directory (The relative path from the root to the storybook config root):`;
+    const storybookDirNote = `The storybook directory (The directory can either be at the root or in a sub-directory):`;
+    const traceSuggestions = `If you are having trouble with tracing, please check the following:\n
+  1. Make sure you have the correct root path, base path, and storybook path.\n
+  2. Make sure you have the correct storybook config file.\n
+  3. Make sure you have the correct storybook config file path.\nYou can either set the flags storybook-base-dir or storybook-config-dir to help TurboSnap find the correct storybook config file.\n`;
```
thafryer, line 38: This text seems duplicated. Can we export it in tracedAffectedFiles so that it's reusable and makes the tests more resilient?

thafryer (commented):
Looks good. We should centralize the long helper text so that we don't have to tweak it in multiple places if changes need to be made.

bin-src/trace.test.ts
ethriel3695: @thafryer 
I updated the text

thafryer (approved):

PR #738, opened 2023-04-06
get the builder name and version via the mainConfig for SB v7+

JonathanKolnik:
Our telemetry reports storybook_version = null for a large proportion of our build events. It turns out these are SB v7 build events and the findBuilder function was erroring out causing the entire metadata Promise.all to fail. This PR updates the logic for finding the builder which should restore the telemetry overall.

### DIFF
--- a/bin-src/lib/getStorybookMetadata.ts
+++ b/bin-src/lib/getStorybookMetadata.ts
@@ -181,35 +181,12 @@ export const getStorybookMetadata = async (ctx: Context) => {
   const r = typeof __non_webpack_require__ !== 'undefined' ? __non_webpack_require__ : require;
   const mainConfig = await r(path.resolve(configDir, 'main'));
 
-  let addons = {};
-  try {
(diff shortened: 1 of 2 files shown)


thafryer (approved):
LGTM!

tmeasday (commented):
LGTM although I wonder if either of these fail we should still be recording whatever other metadata we have?

bin-src/lib/getStorybookMetadata.ts
```diff
@@ -170,12 +181,35 @@ export const getStorybookMetadata = async (ctx: Context) => {
...
+  } catch (e) {
+    ctx.log.info('Unable to determine viewLayer used in this project', e);
+  }
+
+  let builder = {};
+  try {
+    builder = await findBuilder(mainConfig);
+  } catch (e) {
+    ctx.log.info('Unable to determine builder based in this project', e);
+  }
```
thafryer, line 210: As an alternative to this, you could use promise.allSettled. It would be less verbose.

PR #750, opened 2023-05-15
Use `CHROMATIC_PROJECT_TOKEN` environment variable rather than hardcoded value

ghengeveld:
This brings our project setup in line with our general advice to not hardcode project tokens.

Before we merge this, Azure Pipelines needs to be configured with the env var.

We should cycle our token once this PR is merged.

### DIFF
--- a/.github/workflows/chromatic-prod.yml
+++ b/.github/workflows/chromatic-prod.yml
@@ -20,7 +20,7 @@ jobs:
           traceChanged: true
           diagnostics: true
         env:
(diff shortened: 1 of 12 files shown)


ghengeveld:
@thafryer where did we end up with Azure?

thafryer:
> @thafryer where did we end up with Azure?

@ghengeveld We are removing it for now.

tevanoff (approved):

ghengeveld:
✅ Secrets have been cycled

PR #753, opened 2023-06-02
update getStorybookMetadata to safely record version even if fails to parse mainConfig

JonathanKolnik:
Even though we reproduced, tested, and fixed a mainConfig parsing bug, it is possible that certain formats are still not readable. In that event we want to at least still record the metadata that we can gather.


tmeasday (approved):
OK, LGTM. @zol @thafryer this will mean that in some proportion of cases (it looks like maybe as many as 1/3) where we fail to parse `main.js`, we report the SB + CH versions correctly, but we won't have a valid addons list or builder (they'll be "unknown").

In order to investigate further we'll probably need to start uploading the errors we are getting trying to parse the file. That's a bit tricky because we need to sanitize the errors etc. We should consider if we want the data enough to do that.

thafryer (approved):
LGTM. Tests are failing. It's likely because of the tweak to catch block.

PR #754, opened 2023-06-02
Add type statement

kk3939:
I changed three points.(small improvement)

First,  I saw the below error when `yarn install`.
Therefore, I added `any` as a tentative response.
```
TS2339: Property 'app' does not exist on type '{}'.
```

Second, I added the type to ctx.
I think we need to unify action and bin, so we have explicitly aligned the types.
https://github.com/chromaui/chromatic-cli/blob/f767962c1f32b618b2cb1e228d3c2b8925b561ba/bin-src/main.ts#L49

Third, I deleted the unused `Info` of actions/core.


ghengeveld:
Thanks! I reverted your lockfile changes because those seemed unnecessary (and risky). I also merged in the latest updates from `main` which already fixed some of the types. It's all looking good now though.

PR #755, opened 2023-06-02
Add a node entry point

tmeasday:
Add a node entry point at `./api` that exposes the `runChromatic` command to be consumed from node:

```ts
import { runChromatic } from 'chromatic/api';

await runChromatic({ ... });
```

This PR is a draft as it has a couple of issues:

1. For some reason the export isn't actually exported. I haven't looked into it too far because it is something to do with the current webpack setup, and I am not sure it makes sense:

2. In this version we end up bundling all the same code into the `bin/`, `action/` and `node/` directories. I think the `action/` directory isn't published to npm, but we would want to publish the `node/` dir.

I would make the `node/` compile target a simple shim that calls into `../bin` but `bin/main.cjs` of course assumes it is being running as a CLI. So I'm not quite sure what to do.

Probably the most sensible thing would to have `bin/main.js` be a thin wrapper around `node/`. If we do that, it'd probably make sense to move the source code into `node-src` also. WDYT @ghengeveld?

3. I wonder if we should just migrate to `tsup` rather than webpack. @ndelangen has setup many repos to bundle with it successfully. But that might be a distraction.

4. There are no tests yet. Once I've sorted out the packing above, my plan is to make a new github workflow that basically just does the code snippet above. Does that sound good @ghengeveld?


package.json
```diff
@@ -19,6 +19,15 @@
     "url": "https://github.com/chromaui/chromatic-cli.git"
   },
   "license": "MIT",
+  "exports": {
+    ".": {
+      "require": "./isChromatic.js",
+      "import": "./isChromatic.mjs"
+    },
+    "./api": {
+      "import": "./node/main.cjs"
```
JReinhold, line 28: while the docs on this topic has always been incredibly elusive, I think `import` is for ESM and `require is for CJS, so this looks strange to me. Although if you say it works. 🤷
```suggestion
      "require": "./node/main.cjs"
```

tmeasday: Oh lol, no I probably had it backwards. Node doesn't use these entry points so I didn't really test them yet :)

bin-src/git/findAncestorBuildWithCommit.ts
```diff
@@ -55,15 +55,17 @@ export async function findAncestorBuildWithCommit(
   let skip = 0;
   while (skip < limit) {
     // eslint-disable-next-line no-await-in-loop
-    const { app } = await client.runQuery(AncestorBuildsQuery, {
+    const { app } = (await client.runQuery(AncestorBuildsQuery, {
       buildNumber,
       skip,
       limit: Math.min(page, limit - skip),
-    });
+    })) as AncestorBuildsQueryResult;
```
ghengeveld, line 62: This was already fixed [here](https://github.com/chromaui/chromatic-cli/commit/b7a250f8a292befee3441cce5d0514ecf7e115a6).

ghengeveld:
> Add a node entry point at ./api that exposes the runChromatic command to be consumed from node:
> `import { runChromatic } from 'chromatic/api';`

I'm not a fan of `chromatic/api` since it conflates with our (eventual) public API. We may at some point ship an API client, so I'd like to avoid any confusion. I think `chromatic/node` would be more appropriate. Also let's rename `runChromatic` to just `run`.

> Probably the most sensible thing would to have bin/main.js be a thin wrapper around node/. If we do that, it'd probably make sense to move the source code into node-src also. WDYT @ghengeveld?

Yes that seems like the best solution.

> I wonder if we should just migrate to tsup rather than webpack. @ndelangen has setup many repos to bundle with it successfully. But that might be a distraction.

Yes I'd love to migrate to tsup. This seems like an appropriate time to me.

tmeasday:
@ghengeveld can you take another look? I'm still bundling both bin and node into separate bundles, but the sizes are much smaller now so it's probably less of a concern at least in this iteration. We could use tsup to do code splitting too I think (although we might have to be careful it doesn't split the action out too.

tmeasday:
Published a canary release as `6.19.0-canary.0` successfully:
image

tmeasday:
@ghengeveld I'll await further feedback from you on this. In particular:

1. Should I split the tsup config into two (one for `bin-src` + `node-src`, building to `dist/`, which I'll code-split to avoid duplication, one for `action-src`, building to `action/`)?

2. Should I move the bulk of the code to `node-src`, and have `bin-src` call into it, or just leave it as is (seems OK this way once I code split 🤷).

Note that the action publishing is [currently broken](https://github.com/chromaui/chromatic-cli/blob/e31080b0933d6e4fcac4a3180f61ade9368819c3/scripts/release.js#L46-L49) on this branch, but I don't need it, so I wont fix it until 1 is answered.

Also there are some Azure checks going on that I can't log into to see what's going on.

.gitignore
```diff
@@ -11,6 +11,7 @@ subdir-static
 test-storybook
 chromatic-build-*.xml
 chromatic-diagnostics.json
-# the dist folder IS included so we can test every version in the CI without publishing
```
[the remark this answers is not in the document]
ghengeveld: This was [added](https://github.com/chromaui/chromatic-cli/commit/229ca6c919dd4856850f46c3caa7d3fc0f22c1af) a long time ago. I don't know how relevant it is. I suspect this is intended to prevent `dist` from being added to the repo by a CI script that builds the CLI for testing. I'm not sure whether we have such a script anymore. Can't hurt to keep ignoring `dist` but the comment can go 🤷

tsup.config.ts
```diff
@@ -0,0 +1,21 @@
+import { defineConfig } from 'tsup';
+
+export default defineConfig((options) => ({
+  entry: {
+    action: 'action-src/register.js',
+    bin: 'bin-src/register.js',
+    node: 'node-src/index.ts',
```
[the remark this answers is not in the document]
ghengeveld: Yes a separate tsup for the action seems appropriate.

tsup.config.ts
```diff
@@ -0,0 +1,21 @@
...
+  },
+  splitting: false,
+  minify: !options.watch,
+  format: ['cjs'],
+  dts: {
+    entry: ['node-src/index.ts'],
+    resolve: true,
+  },
+  treeshake: true,
+  sourcemap: false,
```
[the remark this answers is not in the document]
ghengeveld: Works for me.

ghengeveld:
> 1. Should I split the tsup config into two (one for `bin-src` + `node-src`, building to `dist/`, which I'll code-split to avoid duplication, one for `action-src`, building to `action/`)?

Yes that seems like the best approach, as it allows us to fine-tune config for the action separately.

> 2. Should I move the bulk of the code to `node-src`, and have `bin-src` call into it, or just leave it as is (seems OK this way once I code split 🤷).

I suspect conceptually most code should be in node-src now, right? Even if you "fixed" that with code splitting, I feel the repo structure should represent the conceptual structure. In other words I think you should move the code over to node-src.

> Note that the action publishing is [currently broken](https://github.com/chromaui/chromatic-cli/blob/e31080b0933d6e4fcac4a3180f61ade9368819c3/scripts/release.js#L46-L49) on this branch, but I don't need it, so I wont fix it until 1 is answered.

We'll deal with that later. I want to revisit how we publish the action anyway (with versioning).

> Also there are some Azure checks going on that I can't log into to see what's going on.

Me neither. @thafryer is looking into replacing the Azure stuff.

> Also, also, let me know any thoughts on a testing strategy.

I don't have any right now beyond what we're already doing, and doing a lot of QA. What did you have in mind?

tmeasday:
>  What did you have in mind?

Just making a new github workflow that calls out to a node script that does something like:

```ts
import { run } from 'chromatic/api';

await run({ ... });
```

(Ultimately similar to our other test workflows)

ghengeveld:
Yeah that makes sense. Basically it'd be an example of how one can build a custom integration using the node API whilst also being an integration test for us.

tsup.config.ts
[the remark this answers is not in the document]
ndelangen: defineConfig takes an array as well!

.gitignore
```diff
@@ -5,12 +5,12 @@ node_modules
...
 storybook-static
 subdir-static
 test-storybook
 chromatic-build-*.xml
 chromatic-diagnostics.json
-# the dist folder IS included so we can test every version in the CI without publishing
+dist
 bin
 action
+node
```
ghengeveld, line 16: No newline at end of file?

action-src/register.js
```diff
@@ -1,6 +1 @@
-const Observable = require('zen-observable');
-
-global.Observable = Observable;
-require('any-observable/register')('global.Observable');
-
```
ghengeveld, line 5: This warrants a major version bump I think.

tmeasday: I'm still registering an observable in the node code

tmeasday: Here: https://github.com/chromaui/chromatic-cli/blob/39642116974da41d4c023ec547bbee4aab8689ab/node-src/index.ts#L3

.gitignore
tmeasday: 🤷

scripts/run-via-node.ts
```diff
@@ -0,0 +1,15 @@
+import process from 'process';
+import { run } from '../dist/node';
+
+async function go() {
```
ghengeveld, line 4: Why wrap this in an extra function?

ghengeveld (approved):

node-src/index.ts
```diff
@@ -0,0 +1,107 @@
+import readPkgUp from 'read-pkg-up';
+import { v4 as uuid } from 'uuid';
+import 'any-observable/register/zen';
```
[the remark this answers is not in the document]
ghengeveld: Right you answered this already

PR #756, opened 2023-06-02
Added `onTaskComplete` callback option

tmeasday:
Telescoping on https://github.com/chromaui/chromatic-cli/pull/755

And ability to pass options to the node API.

This will allow us to get access to the context at the completion of each task, in particular to get the build id at the end of the `announce` task.

### DIFF
--- a/node-src/lib/tasks.ts
+++ b/node-src/lib/tasks.ts
@@ -16,7 +16,7 @@ export const createTask = ({ title, steps, ...config }): Listr.ListrTask<Context
       await step(ctx, task);
     }
 
-    ctx.options.onTaskComplete?.(ctx);
+    ctx.options.onTaskComplete?.({ ...ctx });
   },
   ...config,
 });
--- a/node-src/lib/tasks.ts
+++ b/node-src/lib/tasks.ts
@@ -15,6 +15,8 @@ export const createTask = ({ title, steps, ...config }): Listr.ListrTask<Context
       // eslint-disable-next-line no-await-in-loop
       await step(ctx, task);
     }
+
+    ctx.options.onTaskComplete?.(ctx);
   },
   ...config,
 });
--- a/node-src/types.ts
+++ b/node-src/types.ts
@@ -86,6 +86,9 @@ export interface Options {
   branchName: string;
   patchHeadRef: string;
   patchBaseRef: string;
+
+  /** A callback that is called at the completion of each task */
+  onTaskComplete?: (ctx: Context) => void;
 }
 
 export interface Context {


bin-src/main.ts
```diff
@@ -77,12 +77,12 @@ export async function runAll(ctx) {
...
 }
 
-export async function runBuild(ctx: Context) {
+export async function runBuild(ctx: Context, extraOptions?: Partial<Options>) {
   ctx.log.info('');
   ctx.log.info(intro(ctx));
 
   try {
-    ctx.options = await getOptions(ctx);
+    ctx.options = await getOptions(ctx, extraOptions);
```
ghengeveld, line 85:
```suggestion
    ctx.options = await getOptions(ctx);
    if (extraOptions) ctx.options = { ...ctx.options, ...extraOptions };
```
This makes it more apparent that "all bets are off" at this point.

bin-src/lib/tasks.ts
```diff
@@ -15,6 +15,8 @@ export const createTask = ({ title, steps, ...config }): Listr.ListrTask<Context
       // eslint-disable-next-line no-await-in-loop
       await step(ctx, task);
     }
+
+    ctx.options.onTaskComplete?.(ctx);
```
ghengeveld, line 19:
```suggestion
    ctx.options.onTaskComplete?.({ ...ctx });
```
This should ensure `ctx.title` still refers to the title of the current (finished) task rather than the next one.

node-src/index.ts
```diff
@@ -24,7 +24,13 @@ interface Output {
...
 }
 
-export async function runChromatic(flags: Flags): Promise<Output> {
+export async function runChromaticFull({
+  flags = {},
+  options = {},
+}: {
+  flags?: Flags;
+  options?: Partial<Options>;
+}): Promise<Output> {
```
ghengeveld, line 33: I'm not digging the name `runChromaticFull`. What does "full" mean in this context? I think we should treat this node entry point as a public API for anyone who wants to integrate more deeply with our CLI. I think it should either be a descriptive and truthful name, or be something quite generic like `run` or `main`.

ghengeveld (changes requested):

bin-src/lib/getOptions.ts
```diff
@@ -75,6 +78,7 @@ export default function getOptions({ argv, env, flags, log, packageJson }: Conte
     branchName,
     patchHeadRef,
     patchBaseRef,
+    ...extraOptions,
```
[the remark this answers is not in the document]
tmeasday: @ghengeveld the main issue with your approach is then the checks below (e.g. for missing project token) start hassling me if I pass the token in via options (rather than flags). We should have a chat about how we want this to work.

action-src/main.ts
```diff
@@ -109,34 +109,35 @@ async function run() {
...
-      projectToken,
-      repositorySlug: maybe(repositorySlug),
-      skip: maybe(skip),
-      storybookBaseDir: maybe(storybookBaseDir),
-      storybookBuildDir: maybe(storybookBuildDir),
-      storybookConfigDir: maybe(storybookConfigDir),
-      traceChanged: maybe(traceChanged),
-      untraced: maybe(untraced),
-      zip: maybe(zip, false),
-      junitReport: maybe(junitReport, false),
```
ghengeveld, line 139: Was this removal intentional?

ghengeveld (approved):

action-src/main.ts
tmeasday: It's just moved to the flags option

PR #757, opened 2023-06-04
Fix types and Yarn classic compatibility

ghengeveld:
This restores compatibility with native Yarn 1, since most of us are still on v1. There was also a TypeScript issue I fixed.


ethriel3695 (approved):
Looks great, thank you!

PR #764, opened 2023-06-13
Move all depenendencies to dev deps

tmeasday:
I'm not sure why these utilities were in deps, no dev deps. I think that TSUP will never bundle deps, whereas maybe previously webpack would happily bundle everything? Not sure.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,11 +81,7 @@
   "resolutions": {
     "any-observable": "^0.5.1"
   },
-  "dependencies": {
-    "@discoveryjs/json-ext": "^0.5.7",
-    "@storybook/csf-tools": "^7.0.12",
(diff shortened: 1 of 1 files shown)


tmeasday:
Going to hard merge this one to fix the action (confirmed on canary).

mattseddon:
See https://tsup.egoist.dev/#excluding-packages.

tmeasday:
That's a good move if we want to revert this PR. Will seek clarification as to whether that's the right direction from others.

ndelangen:
@tmeasday so the action absolutely needs everything bundled in, because it has no install step (for performance reasons).

The node-api might be different, because you might want to share dependencies.

Having them all bundled in might not be a bad situation though, this PR seems fine as is for the foreseeable future.

PR #770, opened 2023-06-14
Ensure we exit with a code at the end

tmeasday:
I missed this bit from the old code: https://github.com/chromaui/chromatic-cli/blob/b3fbf71511beb2dcb39da08cecc331e29614f7d3/bin-src/main.ts#L59-L60

Not sure if the log is still necessary but this solves the immediate problem

### DIFF
--- a/bin-src/main.ts
+++ b/bin-src/main.ts
@@ -1,5 +1,7 @@
 import { run } from '../node-src';
 
 export async function main(argv: string[]) {
-  return run({ argv });
+  const { code } = await run({ argv });
+
+  process.exit(code);
 }


bin-src/main.ts
```diff
@@ -1,5 +1,7 @@
 import { run } from '../node-src';
 
 export async function main(argv: string[]) {
-  return run({ argv });
+  const { code } = await run({ argv });
+
+  process.exit(code);
```
ndelangen, line 6: This would normally not be necessary, something is keeping node open, you should probably investigate what. This `.exit()` method can prevent errors from getting fully printed..

https://www.npmjs.com/package/why-is-node-running

Maybe we add a ticket to investigate this, or add a TODO comment?

skitterm (approved):
This looks good to me, restores the `process.exit()` that we've had for awhile. Seems we can wait on making improvements to that part.

ghengeveld (approved):

PR #775, opened 2023-06-21
Increase the timeout for executing Git commands

thafryer:
Ensure that Git commands are not timing out when running builds.


Precha24 (approved):
Go

Precha24 (commented):
Go

PR #793, opened 2023-07-26
Update snyk parser and added tests for yarn berry

ethriel3695:
I've had a couple of users reach out saying that the manifest dependency tracing does not work for yarn berry.

It appears that yarn2 / lockfile v2 with resolutions is specifically what is not working.

Given a package.json with a resolution like so:
"resolutions": {
  "chalk": "^4.1.2"
}

In v1, yarn would list every version of chalk needed by the various dependencies and then specify the resolved version.

In v2, this has been reduced to just the forced version.

```
OutOfSyncError: Dependency chalk@^4.1.2 was not found in yarn.lock. 
Your package.json and yarn.lock are probably out of sync. Please run "yarn install" and try again.
```

Recent versions of https://github.com/snyk/nodejs-lockfile-parser have changed how lockfile v2 is parsed.

I also added tests to verify that the yarn berry lock file format is read successfully, and, changes between v2 lock files are detected as expected.

Screen Shot 2023-07-26 at 4 11 30 PM


ethriel3695:
I don't think this is working, yet.

When I removed the older versions of the package in the lock file, with the hopes of relying on the resolution, it outputs the OutOfSyncError

```
"any-observable@npm:^0.5.1":
  version: 0.5.1
  resolution: "any-observable@npm:0.3.0"
  peerDependenciesMeta:
    rxjs:
      optional: true
    zen-observable:
      optional: true
  checksum: 7ead38d198a7ae5d72ba1a696e0a80cb51a62f3784faa46ab7c69f93c1f6750b2d4eadef9afa2c58acfe27dde54c7e02a051320ae5dbc62bd0e220a34b2085b1
  languageName: node
  linkType: hard
```

ethriel3695:
@ghengeveld @tmeasday 

I updated the lock files to match the resolution format and the structure for yarn berry in `pnp` mode.
I also added tests to verify the behavior and verify we trace the packages correctly.

Please let me know what you think.

tmeasday (approved):
LGTM! thanks for updating the tests.

Do we have a sample repo we can test this against?

ghengeveld (approved):

ghengeveld:
Released in [6.21.0](https://www.npmjs.com/package/chromatic/v/6.21.0).

PR #794, opened 2023-08-02
add getGitInfo function exported by the node-api

ndelangen:
To be used in https://github.com/chromaui/addon-visual-tests/pull/24

### DIFF
--- a/node-src/tasks/index.ts
+++ b/node-src/tasks/index.ts
@@ -11,7 +11,6 @@ import snapshot from './snapshot';
 import storybookInfo from './storybookInfo';
 import upload from './upload';
 import verify from './verify';
-import { getBranch, getCommit, getSlug } from '../git/git';
 
 export const runUploadBuild = [
   auth,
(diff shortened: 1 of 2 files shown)


node-src/tasks/index.ts
```diff
@@ -30,3 +31,20 @@ export default (options: Context['options']): Listr.ListrTask<Context>[] => {
...
+export async function getGitInfo(): Promise<GitInfo> {
+  const branch = await getBranch();
+  const { commit } = await getCommit();
+  const slug = await getSlug();
+
+  const [ownerName, repoName, ...rest] = slug ? slug.split('/') : [];
+  const isValidSlug = !!ownerName && !!repoName && !rest.length;
+
+  return { branch, commit, slug: isValidSlug ? slug : '' };
+}
```
tmeasday, line 50: This isn't a task, I think it should be in `node-src/index.ts` (in fact I'm unclear on how this actually works exported from here).

tmeasday (changes requested):

node-src/tasks/index.ts
ndelangen: Ow, I put it in the wrong file 😓

tmeasday (approved):

ghengeveld:
Released in [6.21.0](https://www.npmjs.com/package/chromatic/v/6.21.0).

PR #795, opened 2023-08-08
Add `--local` flag and pass to builds as `isLocalBuild`

tmeasday:
I'm not sure if we want the flag (it could just be a node API option), but I thought it could be helpful for debugging. I didn't add it as an input to the action. Not sure if that's confusing.

Depends on https://github.com/chromaui/chromatic/pull/7576

### DIFF
--- a/node-src/lib/parseArgs.ts
+++ b/node-src/lib/parseArgs.ts
@@ -27,6 +27,7 @@ export default function parseArgs(argv: string[]) {
       --exit-zero-on-changes [branch]           If all snapshots render but there are visual changes, exit with code 0 rather than the usual exit code 1. Only for [branch], if specified. Globs are supported via picomatch.
       --externals <filepath>                    Disable TurboSnap when any of these files have changed since the baseline build. Globs are supported via picomatch. This flag can be specified multiple times. Requires --only-changed.
       --ignore-last-build-on-branch <branch>    Do not use the last build on this branch as a baseline if it is no longer in history (i.e. branch was rebased). Globs are supported via picomatch.
+      --local                                   Create a "local" build, that will not affect other user's baselines.
       --only-changed [branch]                   Enables TurboSnap: Only run stories affected by files changed since the baseline build. Only for [branch], if specified. Globs are supported via picomatch. All other snapshots will be inherited from the prior commit.
       --only-story-files <filepath>             Only run a single story or a subset of stories by their filename(s). Specify the full path to the story file relative to the root of your Storybook project. Globs are supported via picomatch. This flag can be specified multiple times.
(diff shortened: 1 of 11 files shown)


node-src/lib/parseArgs.ts
```diff
@@ -74,6 +75,7 @@ export default function parseArgs(argv: string[]) {
         exitOnceUploaded: { type: 'string' },
         exitZeroOnChanges: { type: 'string' },
         externals: { type: 'string', isMultiple: true },
+        local: { type: 'boolean' },
         ignoreLastBuildOnBranch: { type: 'string' },
```
ghengeveld, line 79:
```suggestion
        ignoreLastBuildOnBranch: { type: 'string' },
        local: { type: 'boolean' },
```
Alphabetical order.

node-src/lib/parseArgs.ts
```diff
@@ -27,6 +27,7 @@ export default function parseArgs(argv: string[]) {
       --exit-zero-on-changes [branch]           If all snapshots render but there are visual changes, exit with code 0 rather than the usual exit code 1. Only for [branch], if specified. Globs are supported via picomatch.
       --externals <filepath>                    Disable TurboSnap when any of these files have changed since the baseline build. Globs are supported via picomatch. This flag can be specified multiple times. Requires --only-changed.
       --ignore-last-build-on-branch <branch>    Do not use the last build on this branch as a baseline if it is no longer in history (i.e. branch was rebased). Globs are supported via picomatch.
+      --local                                   Create a "local" build, that will not affect other user's baselines.
```
ghengeveld, line 30:
```suggestion
      --local                                   Create a "local" build, which will not affect other user's baselines.
```

node-src/tasks/initialize.ts
```diff
@@ -65,6 +65,7 @@ export const announceBuild = async (ctx: Context) => {
         ...commitInfo,
         committedAt: new Date(committedAt),
         ciVariables: ctx.environment,
+        isLocalBuild,
```
ghengeveld, line 68:
```suggestion
        isLocalBuild: local,
```

node-src/types.ts
```diff
@@ -69,6 +70,7 @@ export interface Options {
   autoAcceptChanges: true | string;
   exitZeroOnChanges: true | string;
   exitOnceUploaded: true | string;
+  isLocalBuild: boolean;
```
ghengeveld, line 73:
```suggestion
  local: boolean;
```

ghengeveld (approved):
Some minor nitpicks but otherwise LGTM.

tmeasday:
@ghengeveld as discussed on Slack we dropped the `--local` option and now the only way to get a local build is to use the node API. We can revisit this later if we like.

PR #796, opened 2023-08-08
Pass `gitUserEmailHash` up with new builds and use `localBuilds` filter in baseline calculations

tmeasday:
Telescoping on https://github.com/chromaui/chromatic-cli/pull/795

We want to mark who created each build[1] so we can filter out local builds created by anyone else apart from the current user.

As such we:

 - Send `gitUserEmaiHash ` as part of the `AnnounceBuild` mutation
 - Send `localBuilds: { localBuildEmailHash: gitUserEmaiHash || committerHash }` as part of all the queries we use for baseline calculations.

This depends on https://github.com/chromaui/chromatic/pull/7576

[1] The creator is the current git user for local builds and the committer otherwise.

### DIFF
--- a/node-src/lib/emailHash.ts
+++ b/node-src/lib/emailHash.ts
@@ -1,4 +1,3 @@
-// This is duplicated from https://github.com/chromaui/chromatic/blob/1ceabdd81936b883a2a9ddc804ed13bfa0919c47/services/index/model/lib/emailHash.ts#L1-L7
 import { createHash } from 'crypto';
 
 // Inspired by https://en.gravatar.com/site/implement/hash
--- a/node-src/lib/emailHash.ts
+++ b/node-src/lib/emailHash.ts
@@ -1,0 +1,7 @@
+// This is duplicated from https://github.com/chromaui/chromatic/blob/1ceabdd81936b883a2a9ddc804ed13bfa0919c47/services/index/model/lib/emailHash.ts#L1-L7
+import { createHash } from 'crypto';
+
+// Inspired by https://en.gravatar.com/site/implement/hash
+export function emailHash(email: string) {
+  return createHash('md5').update(email.trim().toLowerCase()).digest('hex');
+}
--- a/node-src/lib/localBuildsSpecifier.ts
+++ b/node-src/lib/localBuildsSpecifier.ts
@@ -1,0 +1,12 @@
+import { Context } from '../types';
+import { emailHash } from './emailHash';
+
+export function localBuildsSpecifier(ctx: Pick<Context, 'options' | 'git'>) {
+  if (ctx.options.isLocalBuild) return { localBuildEmailHash: emailHash(ctx.git.gitUserEmail) };
+
+  // For non local builds, if we have a committer hash, we want only local builds from that person
+  if (ctx.git.committerEmail) return { localBuildEmailHash: emailHash(ctx.git.committerEmail) };
+
+  // If we don't know, we fall back to *no local builds at all*
+  return { isLocalBuild: false };
+}
--- a/node-src/lib/localBuildsSpecifier.ts
+++ b/node-src/lib/localBuildsSpecifier.ts
@@ -4,7 +4,7 @@ import { emailHash } from './emailHash';
 export function localBuildsSpecifier(ctx: Pick<Context, 'options' | 'git'>) {
   if (ctx.options.isLocalBuild) return { localBuildEmailHash: emailHash(ctx.git.gitUserEmail) };
 
-  // For non local builds, if we have a committer hash, we want only local builds from that person
+  // For global builds, we only want local builds from the committer (besides global builds)
   if (ctx.git.committerEmail) return { localBuildEmailHash: emailHash(ctx.git.committerEmail) };
 
   // If we don't know, we fall back to *no local builds at all*
--- a/node-src/tasks/gitInfo.ts
+++ b/node-src/tasks/gitInfo.ts
@@ -1,7 +1,7 @@
 import picomatch from 'picomatch';
 
 import getCommitAndBranch from '../git/getCommitAndBranch';
-import { getSlug, getVersion } from '../git/git';
+import { getSlug, getUserEmail, getVersion } from '../git/git';
 import { getParentCommits } from '../git/getParentCommits';
 import { getBaselineBuilds } from '../git/getBaselineBuilds';
 import { exitCodes, setExitCode } from '../lib/setExitCode';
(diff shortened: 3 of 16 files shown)


node-src/tasks/gitInfo.ts
```diff
@@ -59,11 +59,19 @@ export const setGitInfo = async (ctx: Context, task: Task) => {
...
     fromCI: ci,
     interactive,
+    isLocalBuild,
   } = ctx.options;
 
+  const commitAndBranchInfo = await getCommitAndBranch(ctx, { branchName, patchBaseRef, ci });
+
+  const creatorEmail = isLocalBuild
+    ? await getUserEmail()
+    : commitAndBranchInfo.committerEmail || 'unknown';
```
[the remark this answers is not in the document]
ghengeveld: I don't think we should record the committer as the creator for non-local builds. It would make it harder to tell where the data originated (the commit or the git config) whilst not providing any benefit (we already record `committerEmail` so it would be duplicative).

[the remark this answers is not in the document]
ghengeveld: We should consider hashing the email and recording `creatorId` or something like that instead, in order to avoid recording (even more) PII. We don't need the actual plaintext email, right? We can always hash the `committerEmail` as well if we want to cross-reference a CI build with a local build. Gravatar's [implementation](https://en.gravatar.com/site/implement/hash) (md5) seems reasonable.

ghengeveld (commented):
There's some far-reaching consequences we should consider before merging this. Overall I'm not too keen on the implementation specifics right now.

node-src/tasks/gitInfo.ts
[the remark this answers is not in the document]
ghengeveld: What happens if (for whatever reason) we can't retrieve the Git user email when running a local build? I think it should not be allowed to do that because we wouldn't be able to track local baselines.

node-src/types.ts
```diff
@@ -141,8 +141,16 @@ export interface Context {
 
   git: {
     version: string;
+    /**
+     * The creator of the build is:
+     *   - the current user in local builds
+     *   - the creator of the commit in CI builds
+     *   - 'unknown' in CI situations where we cannot read the current commit info.
+     */
+    creatorEmail: string;
```
ghengeveld, line 150: Within the context of the `git` object, this should be `userEmail`. Our APIs of course should still reference `creatorEmail` (or e.g. `creatorId`).

tmeasday: I was using `userEmail` to refer to the `git config user.email`, where as `creatorEmail` is the slightly more complicated thing referenced in the comment above. I wanted to add the complicated thing to the `git` object because we need to pass it to various APIs along the way, so I didn't want to recalculate it each time (although I could).

node-src/tasks/gitInfo.ts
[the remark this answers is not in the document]
tmeasday: > What's the purpose of recording unknown? Usually we do that only when retrieving data, not when recording it.

I guess we don't need to *record* `'unknown'`, as really the `creatorEmail` doesn't have any effect on non-local builds anyway (at the moment).

The purpose for it on non-local builds is in baseline selection (ie. queries we do). In these cases we want to pass *something* as `onlyLocalBuildsFrom` on CI builds if we can't calculate the commit's author. That will have the effect of filtering out *all* local builds, which I think is better than including all local builds.

We probably also need to show a large warning when this happens.

> Let's just always record the configured user email (preferably hashed), even for CI builds. We already track the committerEmail, so it would be duplicative anyway.

Given the above, we might consider not even storing a `creatorEmail` on CI builds? I could go either way.

> We should consider hashing the email and recording creatorId or something like that instead, in order to avoid recording (even more) PII. 

OK, sure, let's start hashing (the new) email fields.

> What happens if (for whatever reason) we can't retrieve the Git user email when running a local build? I think it should abort the build (throw an error) because we wouldn't be able to track local baselines. Falling back to unknown is likely to cause weird issues, but the above logic isn't doing that (for local builds) anyway.

I think it would throw an error but I'll verify that.

node-src/lib/emailHash.ts
```diff
@@ -0,0 +1,7 @@
+// This is duplicated from https://github.com/chromaui/chromatic/blob/1ceabdd81936b883a2a9ddc804ed13bfa0919c47/services/index/model/lib/emailHash.ts#L1-L7
```
ghengeveld, line 1:
```suggestion
```
Doesn't seem appropriate to refer to private code from a public repo.

node-src/lib/localBuildsSpecifier.ts
```diff
@@ -0,0 +1,12 @@
+import { Context } from '../types';
+import { emailHash } from './emailHash';
+
+export function localBuildsSpecifier(ctx: Pick<Context, 'options' | 'git'>) {
+  if (ctx.options.isLocalBuild) return { localBuildEmailHash: emailHash(ctx.git.gitUserEmail) };
+
+  // For non local builds, if we have a committer hash, we want only local builds from that person
```
ghengeveld, line 7:
```suggestion
  // For global builds, we only want local builds from the committer (besides global builds)
```

node-src/tasks/initialize.ts
```diff
@@ -62,6 +64,7 @@ export const announceBuild = async (ctx: Context) => {
         patchBaseRef,
         patchHeadRef,
         preserveMissingSpecs,
+        gitUserEmailHash: emailHash(gitUserEmail),
```
ghengeveld, line 67:
```suggestion
        ...(gitUserEmail && { gitUserEmailHash: emailHash(gitUserEmail) }),
```
Otherwise we'll potentially get a hash of an empty string.

ghengeveld (approved):

node-src/tasks/gitInfo.ts
```diff
@@ -59,13 +60,21 @@ export const setGitInfo = async (ctx: Context, task: Task) => {
...
     interactive,
+    isLocalBuild,
   } = ctx.options;
 
+  const commitAndBranchInfo = await getCommitAndBranch(ctx, { branchName, patchBaseRef, ci });
+
   ctx.git = {
     version: await getVersion(),
-    ...(await getCommitAndBranch(ctx, { branchName, patchBaseRef, ci })),
+    gitUserEmail: await getUserEmail(),
```
ghengeveld, line 70: It seems this git command fails when there is no email configured, which is likely in a CI environment. 
```suggestion
    gitUserEmail: await getUserEmail().catch((e) => {
      ctx.log.debug('Failed to retrieve Git user email', e);
      return null;
    }),
```
I used this same pattern when [retrieving the uncommitted hash](https://github.com/chromaui/chromatic-cli/pull/798/files).

PR #798, opened 2023-08-11
Calculate and record `uncommittedHash` when creating a build

ghengeveld:
This is required for the "is outdated" concept in the Visual Tests addon, but may also be used in the future for other builds (e.g. during onboarding or even CI builds with an unclean working directory).

### DIFF
--- a/node-src/git/git.ts
+++ b/node-src/git/git.ts
@@ -90,6 +90,29 @@ export async function getBranch() {
   }
 }
 
+// Retrieve the hash of all uncommitted files, which includes staged, unstaged, and untracked files,
+// excluding deleted files (which can't be hashed) and ignored files. There is no one single Git
+// command to reliably get this information, so we use a combination of commands grouped together.
+export async function getUncommittedHash() {
+  const listStagedFiles = 'git diff --name-only --diff-filter=d --cached';
+  const listUnstagedFiles = 'git diff --name-only --diff-filter=d';
+  const listUntrackedFiles = 'git ls-files --others --exclude-standard';
+  const listUncommittedFiles = [listStagedFiles, listUnstagedFiles, listUntrackedFiles].join(';');
+
+  const uncommittedHash = (
+    await execGitCommand(
+      // Pass the combined list of filenames to hash-object to retrieve a list of hashes. Then pass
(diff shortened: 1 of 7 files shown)


node-src/git/git.ts
```diff
@@ -90,6 +90,29 @@ export async function getBranch() {
...
+  const listUnstagedFiles = 'git diff --name-only --diff-filter=d';
+  const listUntrackedFiles = 'git ls-files --others --exclude-standard';
+  const listUncommittedFiles = [listStagedFiles, listUnstagedFiles, listUntrackedFiles].join(';');
+
+  const uncommittedHash = (
+    await execGitCommand(
+      // Pass the combined list of filenames to hash-object to retrieve a list of hashes. Then pass
+      // the list of hashes to hash-object again to retrieve a single hash of all hashes. We use
+      // stdin to avoid the limit on command line arguments.
+      `(${listUncommittedFiles}) | git hash-object --stdin-paths | git hash-object --stdin`
```
tmeasday, line 107: As I mentioned, I don't think `x ; y ; z` does what you think it does in bash. Am I wrong?

tmeasday (changes requested):

node-src/git/git.ts
ghengeveld: I tried it and it gave me a concatenated list of the results of the three operations. So long as they're wrapped in parenthesis, the piping works. But maybe I missed something.

weeksling (approved):
lgtm!

weeksling (approved):
lgtm

weeksling:
It looks like this is failing tests because the CI for multiple reasons. Action-canary is a [git.email issue](https://github.com/chromaui/chromatic-cli/actions/runs/5834250340/job/15823377341?pr=798) others are 400 errors from git

ghengeveld:
> It looks like this is failing tests because the CI for multiple reasons. Action-canary is a [git.email issue](https://github.com/chromaui/chromatic-cli/actions/runs/5834250340/job/15823377341?pr=798) others are 400 errors from git

That's caused by [this PR](https://github.com/chromaui/chromatic-cli/pull/796) which is released as canary but not merged yet. It needs to deal with missing email config.

PR #801, opened 2023-08-23
Fix `Unexpected build status: PREPARED` error

ghengeveld:
Previously, the CLI could crash if the `waitForBuild` polling happened to return a build while it was `PUBLISHED` or `PREPARED`. These would be considered "complete" (not in progress) but the code that runs after the build is complete (rightfully) does not handle those build statuses. 

I could've updated the status check to take `PUBLISHED` and `PREPARED` into account but it seems more resilient to look at `completedAt` instead.

### DIFF
--- a/node-src/main.test.ts
+++ b/node-src/main.test.ts
@@ -145,7 +145,7 @@ jest.mock('node-fetch', () =>
       if (query.match('SnapshotBuildQuery')) {
         return {
           data: {
-            app: { build: { status: 'PENDING', changeCount: 1 } },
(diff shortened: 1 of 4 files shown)


yannbf (commented):
LGTM, as long as there are no gotchas with using `completedAt` as a check (for instance, a build could be completed but be in a broken state maybe)

ghengeveld:
> LGTM, as long as there are no gotchas with using `completedAt` as a check (for instance, a build could be completed but be in a broken state maybe)

Broken, failed and cancelled builds are already handled as valid completed builds.

tmeasday (approved):
LGTM

PR #802, opened 2023-08-23
Include all commit info in `GitInfo`

ghengeveld:
This adds `committedAt`, `committerName` and `committerEmail` to `GitInfo`, so that we can compare the last local commit against the `committedAt` of a build.


ghengeveld:
Released as `6.24.0-canary.0` (canary)

tmeasday (approved):
LGTM

PR #803, opened 2023-08-24
Add Modes Suffix to CLI Progress Indicator

thafryer:
With the addition of Modes, the prefix in the progress indicator should indicate whether this test has a valid mode or not.


andrewortwein (approved):
👍🏻

PR #805, opened 2023-08-28
Add a `onTaskProgress` option and report progress on it

tmeasday:
Currently only reported for tasks that can report meaningful progress. Open to changing the API.

### DIFF
--- a/node-src/tasks/snapshot.ts
+++ b/node-src/tasks/snapshot.ts
@@ -70,6 +70,10 @@ export const takeSnapshots = async (ctx: Context, task: Task) => {
   const updateProgress = throttle(
     ({ cursor, label }) => {
       task.output = pending(ctx, { cursor, label }).output;
+      ctx.options.onTaskProgress?.(
+        { ...ctx },
+        { progress: cursor, total: actualTestCount, unit: 'snapshots' }
+      );
     },
     // Avoid spamming the logs with progress updates in non-interactive mode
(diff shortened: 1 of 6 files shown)


node-src/tasks/snapshot.ts
```diff
@@ -70,6 +70,10 @@ export const takeSnapshots = async (ctx: Context, task: Task) => {
   const updateProgress = throttle(
     ({ cursor, label }) => {
       task.output = pending(ctx, { cursor, label }).output;
+      ctx.options.onTaskProgress?.(
+        { ...ctx },
+        { progress: cursor, total: actualTestCount, unit: 'snapshots' }
+      );
```
ghengeveld, line 76: I wonder if this should take skipped snapshots into account. Otherwise it's going to start at 90% progress for TurboSnap builds. Maybe an `initial` value would be able to cover that?

ghengeveld (approved):

node-src/tasks/snapshot.ts
tmeasday: I think I'll just subtract from the total, no need to over complicate, right?

tmeasday: Actually the `actualTestCount` is the right number: 

https://github.com/chromaui/chromatic-cli/blob/c4d4c8bb9bf2d209c09fd1fe6f141b05bdadac2c/node-src/tasks/verify.ts#L109

PR #808, opened 2023-08-30
Add `onTaskError` option to report errors to node consumers

weeksling:
Adds an option that receives an error when options or task fails. Consumes the same fatalError message that would be logged to CLI for a user error.

### DIFF
--- a/node-src/main.test.ts
+++ b/node-src/main.test.ts
@@ -356,7 +356,10 @@ it('passes options error to onTaskError', async () => {
 
   await expect(ctx.options.onTaskError).toHaveBeenCalledWith(
     expect.anything(), // Context
-    expect.stringContaining('Missing project token') // Long formatted error fatalError https://github.com/chromaui/chromatic-cli/blob/217e77671179748eb4ddb8becde78444db93d067/node-src/ui/messages/errors/fatalError.ts#L11
+    expect.objectContaining({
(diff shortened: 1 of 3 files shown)


weeksling:
@tmeasday I mentioned this the other day. Right now I just thought to push the formatted error we normally send to the CLI output to the user. 
This captures runtime errors I can emit to the client, but the output is pretty long for storybook notification. 

I'm curious what you think about adding another formatter to make it shorter or passing along the raw error or other info along with the formatted error.

tmeasday (approved):
LGTM but I agree, let's pass the raw error(s) as well as the formatted version.

node-src/main.test.ts
```diff
@@ -343,6 +343,26 @@ it('fails on missing project token', async () => {
...
+
+  ctx.options.onTaskError = jest.fn();
+  ctx.env.CHROMATIC_PROJECT_TOKEN = '';
+  await runBuild(ctx);
+
+  await expect(ctx.options.onTaskError).toHaveBeenCalledWith(
+    expect.anything(), // Context
+    expect.objectContaining({
+      formattedError: expect.stringContaining('Missing project token'), // Long formatted error fatalError https://github.com/chromaui/chromatic-cli/blob/217e77671179748eb4ddb8becde78444db93d067/node-src/ui/messages/errors/fatalError.ts#L11
+      originalError: expect.anything(), // No jest matcher for an error.
```
ghengeveld, line 361: How about `expect.any(Error)` or if that doesn't work, `expect.objectContaining({ message: expect.any(String) })` ?

ghengeveld (approved):

PR #810, opened 2023-09-04
Add `onTaskStart`, and a new typed field `ctx.task`

tmeasday:
Rather than have the addon try to infer the current task from what's set on the context (or worse by reading the titles), let's just make everything explicit.

### DIFF
--- a/node-src/lib/tasks.ts
+++ b/node-src/lib/tasks.ts
@@ -22,7 +22,7 @@ export const createTask = ({
     ctx.title = title;
     ctx.startedAt = Number.isInteger(ctx.now) ? ctx.now : new Date().getTime();
 
-    ctx.options.onTaskStart?.({ ...ctx });
+    ctx.options.experimental_onTaskStart?.({ ...ctx });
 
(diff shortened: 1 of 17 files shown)


node-src/lib/tasks.ts
```diff
@@ -1,15 +1,29 @@
...
+
+export const createTask = ({
+  name,
+  title,
+  steps,
+  ...config
+}: TaskInput): Listr.ListrTask<Context> => ({
   title,
   task: async (ctx: Context, task: Listr.ListrTaskWrapper<Context>) => {
+    ctx.task = name;
```
ghengeveld, line 21: I don't like exposing task names as an API. People will start to take the order of tasks for granted as part of the API, but I want to have the flexibility to change the order of tasks, or even add/remove tasks as we see fit.

tmeasday: Have discussed and decided this API is basically internal for now. We will mark it as such (`experimental_`) and not worry too much about it.

ghengeveld (approved):

PR #812, opened 2023-09-06
Allow running a build on a single commit repo w/ option

tmeasday:
We have this check because in CI it is likely a single commit repo means a misconfigured CI job.

In the addon, this is much less likely, so we may as well allow you to run such builds as part of experimenting with a new storybook. I didn't add it as a CLI flag as it seems like asking for trouble.

I'm not sure if there's a good way to write tests for this. The original PR that added the check didn't have tests for https://github.com/chromaui/chromatic-cli/pull/207 -- do you have any ideas @ghengeveld ?


node-src/git/getCommitAndBranch.ts
```diff
@@ -56,7 +56,7 @@ export default async function getCommitAndBranch(
   const isGitHubAction = GITHUB_ACTIONS === 'true';
   const isGitHubPrBuild = GITHUB_EVENT_NAME === 'pull_request';
 
-  if (!(await hasPreviousCommit())) {
+  if (!options.allowSingleCommitRepositories && !(await hasPreviousCommit())) {
```
ghengeveld, line 59: Perhaps we should just only run the check in CI?

We could even query for builds _not_ matching the commit and throw if we find something (which indicates other commits are known to exist). That's not completely foolproof though (they might've amended) and probably isn't worth it.

ghengeveld (approved):
I'm not sure we need a new flag at all, but OK if you want to go for it.

node-src/git/getCommitAndBranch.ts
ghengeveld: Either way I think we should still print a warning message.

ghengeveld (approved):

PR #814, opened 2023-09-11
Add support for a JSON configuration file

tmeasday:
Note that I made a judgement on which keys to support in the file (see https://github.com/chromaui/chromatic-cli/blob/72ab4cc542a21e9bacddd1a22daa38aec2a6fdbb/node-src/lib/getConfiguration.ts#L9-L34) - which may be wrong, I am very open to ideas here.

### DIFF
--- a/node-src/index.ts
+++ b/node-src/index.ts
@@ -16,7 +16,6 @@ import invalidPackageJson from './ui/messages/errors/invalidPackageJson';
 import noPackageJson from './ui/messages/errors/noPackageJson';
 import { getBranch, getCommit, getSlug, getUserEmail, getUncommittedHash } from './git/git';
 import { emailHash } from './lib/emailHash';
-
 /**
  Make keys of `T` outside of `R` optional.
 */
@@ -38,7 +37,7 @@ interface Output {
   inheritedCaptureCount: number;
 }
 
-export type { Flags, Options, TaskName, Context } from './types';
+export type { Flags, Options, TaskName, Context, Configuration } from './types';
 
 export async function run({
   argv = [],
@@ -81,8 +80,9 @@ export async function run({
   setExitCode(ctx, exitCodes.OK);
 
   ctx.http = (ctx.http as HTTPClient) || new HTTPClient(ctx);
+  ctx.extraOptions = options;
 
-  await runAll(ctx, options);
+  await runAll(ctx);
 
   return {
     // Keep this in sync with the configured outputs in action.yml
@@ -102,9 +102,9 @@ export async function run({
   };
 }
 
-export async function runAll(ctx, options?: Partial<Options>) {
+export async function runAll(ctx) {
   // Run these in parallel; neither should ever reject
-  await Promise.all([runBuild(ctx, options), checkForUpdates(ctx)]);
+  await Promise.all([runBuild(ctx), checkForUpdates(ctx)]);
 
   if (ctx.exitCode === 0 || ctx.exitCode === 1) {
     await checkPackageJson(ctx);
--- a/node-src/lib/getConfiguration.ts
+++ b/node-src/lib/getConfiguration.ts
@@ -1,0 +1,64 @@
+import { readFile } from 'jsonfile';
+import { ZodError, z } from 'zod';
+import { missingConfigurationFile } from '../ui/messages/errors/missingConfigurationFile';
+import { unparseableConfigurationFile } from '../ui/messages/errors/unparseableConfigurationFile';
+import { invalidConfigurationFile } from '../ui/messages/errors/invalidConfigurationFile';
+
+const configurationSchema = z
+  .object({
+    projectId: z.string(),
+    projectToken: z.string(),
+
+    onlyChanged: z.union([z.string(), z.boolean()]),
+    onlyStoryFiles: z.array(z.string()),
+    onlyStoryNames: z.array(z.string()),
+    untraced: z.array(z.string()),
+    externals: z.array(z.string()),
+    ci: z.boolean(),
+    debug: z.boolean(),
+    junitReport: z.union([z.string(), z.boolean()]),
+    autoAcceptChanges: z.union([z.string(), z.boolean()]),
+    exitZeroOnChanged: z.union([z.string(), z.boolean()]),
+    exitOnceUploaded: z.union([z.string(), z.boolean()]),
+    ignoreLastBuildOneBranch: z.string(),
+
+    buildScriptName: z.string(),
+    outputDir: z.string(),
+
+    storybookBuildDir: z.string(),
+    storybookBaseDir: z.string(),
+    storybookConfigDir: z.string(),
+
+    ownerName: z.string(),
+    repositorySlug: z.string(),
+  })
+  .partial();
+
+export type Configuration = z.infer<typeof configurationSchema>;
+
+export async function getConfiguration(configFile?: string) {
+  const usedConfigFile = configFile || '.chromatic.config.json';
+  try {
+    const rawJson = await readFile(usedConfigFile);
+
+    return configurationSchema.parse(rawJson);
+  } catch (err) {
+    // Config file does not exist
+    if (err.message.match(/ENOENT/)) {
+      // The user passed no configFile option so it's OK for the file not to exist
+      if (!configFile) {
+        return {};
+      }
+      if (configFile) {
+        throw new Error(missingConfigurationFile(configFile));
+      }
+    }
+    if (err.message.match('Unexpected string')) {
+      throw new Error(unparseableConfigurationFile(usedConfigFile, err));
+    }
+    if (err instanceof ZodError) {
+      throw new Error(invalidConfigurationFile(usedConfigFile, err));
+    }
+    throw err;
+  }
+}
--- a/node-src/lib/getOptions.ts
+++ b/node-src/lib/getOptions.ts
@@ -25,13 +25,21 @@ const undefinedIfEmpty = <T>(array: T[]) => {
   return filtered.length ? filtered : undefined;
 };
 
-export default function getOptions({ argv, env, flags, log, packageJson }: Context): Options {
+export default function getOptions({
+  argv,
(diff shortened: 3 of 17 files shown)


package.json
```diff
@@ -101,7 +101,9 @@
   "resolutions": {
     "any-observable": "^0.5.1"
   },
-  "dependencies": {},
+  "dependencies": {
+    "zod": "^3.22.2"
+  },
```
ghengeveld, line 106: I think this should be in devDependencies because we bundle everything.

node-src/ui/messages/errors/missingConfigurationFile.ts
```diff
@@ -0,0 +1,11 @@
+import chalk from 'chalk';
+import { dedent } from 'ts-dedent';
+
+import { error } from '../../components/icons';
+
+export const missingConfigurationFile = (configFile: string) =>
+  dedent(chalk`
+    ${error} Configuration file ${chalk.bold(configFile)} could not be found.
+
+    Check the ${chalk.bold(`--config-file`)} flag of the CLI.
```
ghengeveld, line 10:
```suggestion
    ${error} Configuration file {bold ${configFile}} could not be found.

    Check the {bold --config-file} flag of the CLI.
```
That's how chalk works.

node-src/ui/messages/errors/invalidConfigurationFile.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { error } from '../../components/icons';
+
+export const invalidConfigurationFile = (configFile: string, err: ZodError) => {
+  const { formErrors, fieldErrors } = err.flatten();
+
+  return dedent(chalk`
+    ${error} Configuration file ${chalk.bold(
+    configFile
+  )} was invalid, please check the allowed keys.
```
ghengeveld, line 13:
```suggestion
    ${error} Configuration file {bold ${configFile}} was invalid, please check the allowed keys.
```

node-src/lib/getConfiguration.ts
```diff
@@ -0,0 +1,64 @@
...
+    onlyStoryNames: z.array(z.string()),
+    untraced: z.array(z.string()),
+    externals: z.array(z.string()),
+    ci: z.boolean(),
+    debug: z.boolean(),
+    junitReport: z.union([z.string(), z.boolean()]),
+    autoAcceptChanges: z.union([z.string(), z.boolean()]),
+    exitZeroOnChanged: z.union([z.string(), z.boolean()]),
+    exitOnceUploaded: z.union([z.string(), z.boolean()]),
+    ignoreLastBuildOneBranch: z.string(),
```
ghengeveld, line 23:
```suggestion
    ignoreLastBuildOnBranch: z.string(),
```

node-src/lib/getConfiguration.ts
```diff
@@ -0,0 +1,64 @@
...
+    onlyChanged: z.union([z.string(), z.boolean()]),
+    onlyStoryFiles: z.array(z.string()),
+    onlyStoryNames: z.array(z.string()),
+    untraced: z.array(z.string()),
+    externals: z.array(z.string()),
+    ci: z.boolean(),
+    debug: z.boolean(),
+    junitReport: z.union([z.string(), z.boolean()]),
+    autoAcceptChanges: z.union([z.string(), z.boolean()]),
+    exitZeroOnChanged: z.union([z.string(), z.boolean()]),
```
ghengeveld, line 21:
```suggestion
    exitZeroOnChanges: z.union([z.string(), z.boolean()]),
```

node-src/lib/getConfiguration.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    ownerName: z.string(),
+    repositorySlug: z.string(),
+  })
+  .partial();
+
+export type Configuration = z.infer<typeof configurationSchema>;
+
+export async function getConfiguration(configFile?: string) {
+  const usedConfigFile = configFile || '.chromatic.config.json';
```
ghengeveld, line 40:
```suggestion
  const usedConfigFile = configFile || 'chromatic.config.json';
```
Very much a personal preference, but I don't like dotfiles that have an extension. It's weird. I see no reason to prefix it with a dot.

ghengeveld (approved):
I don't like the filename, but otherwise looks good.

node-src/lib/getConfiguration.ts
jonniebigodes: Agreed with your reasoning @ghengeveld .

node-src/lib/getConfiguration.ts
```diff
@@ -0,0 +1,64 @@
...
+    ignoreLastBuildOneBranch: z.string(),
+
+    buildScriptName: z.string(),
+    outputDir: z.string(),
+
+    storybookBuildDir: z.string(),
+    storybookBaseDir: z.string(),
+    storybookConfigDir: z.string(),
+
+    ownerName: z.string(),
```
jonniebigodes, line 32: @tmeasday can you provide some context on the `ownerName` and `repositorySlug` options as these seem more for us for internal use and if that's the case we shouldn't have them documented.

node-src/lib/getConfiguration.ts
```diff
@@ -0,0 +1,64 @@
...
+    projectToken: z.string(),
+
+    onlyChanged: z.union([z.string(), z.boolean()]),
+    onlyStoryFiles: z.array(z.string()),
+    onlyStoryNames: z.array(z.string()),
+    untraced: z.array(z.string()),
+    externals: z.array(z.string()),
+    ci: z.boolean(),
+    debug: z.boolean(),
+    junitReport: z.union([z.string(), z.boolean()]),
```
jonniebigodes, line 19: Looking at this option, wouldn't the `diagnostics` one be a better fit here for debugging purposes?  Also, wouldn't the`ci` flag impact the builds coming from the addon itself? Unless there's something more to to these that I'm not seeing here.

jonniebigodes (approved):
@tmeasday left some questions for you that we can follow async to address the documentation accordingly. I'm approving this to unblock you on this.

node-src/lib/getConfiguration.ts
tmeasday: The `junitReport` is used to generate a file consumed by services like circle CI, ala:

image

My aim here is to include all the options that you might want to configure chromatic "permanently" to (eventually) replace the need for CLI flags, outside of one-off or special case invocations of the CLI.

node-src/lib/getConfiguration.ts
tmeasday: Hmm, so it looks like the CLI has `--repository-slug` and the action has these two:

https://github.com/chromaui/chromatic-cli/blob/544464cdf53e8bfe73b7e1781e3834c4eeee3e48/CHANGELOG.md#L119

I'm not sure what makes more sense to be configured in a JSON file. WDYT? @ghengeveld?

tmeasday: Also -- I assume options provided to the action should take precedence over configuration settings (that's what I did with CLI flags). Do we agree?

tmeasday: Actually @ghengeveld I realise that I am not actually preferencing CLI flags over configuration options yet in this PR. It's a bit complicated, can we pair on it please?

ghengeveld: Yes, let's pair on it tomorrow?

ghengeveld: I would allow setting `repositorySlug` and `branchName` rather than `ownerName`. The ownerName can be inferred from either of them (repositorySlug `chromaui/chromatic` and branchName `chromaui:main` both yield ownerName `chromaui`). Usually both are retrieved from Git. I suppose there's value in being able to only override the `ownerName` part, but that would add complexity as we'd have to update the `slug` field on the announce mutation to take the override into account.

node-src/lib/getConfiguration.ts
jonniebigodes: @tmeasday I talked with @ghengeveld, and one thing that may have a ripple effect is when a user enables the `ci` flag in the configuration file. For example, the user enables the option in the file and runs a build from the CLI/CI. Then, it links it to the addon and runs another build. As the `ci` option is enabled, it can cause some issues as the second build will not be marked as a local build but as a public build (i.e., from the CLI or CI job) when it's not. I was wondering if we should avoid having it for now or have a detection mechanism for situations like this to prevent issues. Happy to discuss this in more detail.

tmeasday: I think you are right, it's not a good flag for the file, let's drop it.

weeksling (approved):
I had a question and a couple suggestions about using the flag to override configuration instead of the other way around.

node-src/lib/getOptions.ts
```diff
@@ -77,6 +85,7 @@ export default function getOptions({ argv, env, flags, log, packageJson }: Conte
     patchHeadRef,
     patchBaseRef,
   };
+  const options = { ...optionsFromFlags, ...configuration, ...extraOptions };
```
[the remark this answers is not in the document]
tmeasday: I agree, but unfortunately it's not as simple as that because the `optionsFromFlags` set a bunch of defaults, and it's not totally clear how we should handle it in all cases. I'll work through it with @ghengeveld tonight.

node-src/index.ts
```diff
@@ -38,7 +37,7 @@ interface Output {
   inheritedCaptureCount: number;
 }
 
-export type { Flags, Options, TaskName, Context } from './types';
+export type { Flags, Options, TaskName, Context, Configuration } from './types';
```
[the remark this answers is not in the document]
tmeasday: Oh, that's an export actually

node-src/lib/getConfiguration.ts
tmeasday: We decided it's best to drop both options for now because if someone forks the repo they'll keep the configuration, which might be bad.

PR #816, opened 2023-09-14
Add support for a JSON configuration file

tmeasday:


tmeasday:
I don't think the problem is this branch, it happens on `main` too.

PR #819, opened 2023-09-18
No longer log to loggly

tmeasday:
We don't look at it as far as I am aware and it was causing an issue with this error: https://app.circleci.com/pipelines/github/chromaui/chromatic/26578/workflows/81195e51-d036-4244-aa8b-b62ad5e3dd75/jobs/267077


ethriel3695 (approved):
Yeah, we don't use Loggly anymore.
Looks great!

thafryer (approved):

PR #820, opened 2023-09-20
Add `experimental_abortSignal` to Node API to allow canceling builds

ghengeveld:
Aborting the AbortSignal will gracefully cancel the build process. This will only work so long as the build isn't fully published yet, because at that point it's handed off to the capture cloud which doesn't yet support cancellation.

The build-storybook process is actually terminated. The upload should also be terminated (TODO).

### DIFF
--- a/node-src/tasks/build.ts
+++ b/node-src/tasks/build.ts
@@ -84,7 +84,6 @@ export const buildStorybook = async (ctx: Context) => {
     subprocess = execa.command(command, { stdio: [null, logFile, logFile] });
     await Promise.race([subprocess, timeoutAfter(ctx.env.STORYBOOK_BUILD_TIMEOUT)]);
   } catch (e) {
-    endActivity(ctx);
     abortSignal?.throwIfAborted();
 
     const buildLog = fs.readFileSync(ctx.buildLogFile, 'utf8');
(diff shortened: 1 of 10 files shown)


node-src/tasks/build.ts
```diff
@@ -69,14 +69,23 @@ export const buildStorybook = async (ctx: Context) => {
...
-      execa.command(command, { stdio: [null, logFile, logFile] }),
-      timeoutAfter(ctx.env.STORYBOOK_BUILD_TIMEOUT),
-    ]);
+
+    let subprocess;
+    if (abortSignal) {
+      abortSignal.onabort = () => {
+        subprocess?.kill('SIGTERM', { forceKillAfterTimeout: 2000 });
+      };
+    }
```
tmeasday, line 83: execa has [supported an abort signal since `6.1.0`](https://github.com/sindresorhus/execa/releases/tag/v6.1.0). Maybe we should upgrade?

tmeasday (commented):
Seems like the right approach.

node-src/tasks/build.ts
ghengeveld: Oh nice! I'll look into that.

ghengeveld:
Wth I didn't want to merge this yet.

PR #821, opened 2023-09-21
Migrate from Jest to Vitest, update ESLint config and upgrade Execa

ghengeveld:
All this to allow us to use AbortSignal with Execa...


tmeasday (approved):
Seems fine although I didn't review it all. Is there any thing in there I should look at?

ghengeveld:
@tmeasday The tests for `getParentCommits` use `process.chdir` which isn't supported in Vitest without passing `--no-threads`, so I ended up configuring it to run separately. I think that might be worth checking out but otherwise I feel pretty good about this change.

node-src/git/getParentCommits.test.ts
```diff
@@ -76,17 +72,18 @@ interface Repository {
...
       const runGit = makeRunGit(dirname);
       const commitMap = await generateGitRepository(runGit, descriptions[key]);
       repositories[key] = { dirname, runGit, commitMap };
     })
-  )
-);
+  );
+});
 
+// Skipped because vitest doesn't allo `chdir` in workers
```
tmeasday, line 86: Not any more?

tmeasday (approved):
Seems reasonable

PR #822, opened 2023-09-22
Add `experimental_abortSignal` to Node API to allow canceling builds

ghengeveld:
Supersedes #820 

Aborting the AbortSignal will gracefully cancel the build process. This will only work so long as the build isn't fully published yet, because at that point it's handed off to the capture cloud which doesn't yet support cancellation.

The build-storybook process is actually terminated. The upload should also be terminated (TODO).


ghengeveld:
Now that Execa's upgraded, I'll look into passing the abortSignal to it directly.

JonathanKolnik (approved):
✅

PR #832, opened 2023-10-05
Improvements to auto release workflow

ghengeveld:


tmeasday (commented):
LGTM! Where does it actually call out to `publish-action` though?

ghengeveld:
@tmeasday Right [here](https://github.com/chromaui/chromatic-cli/blob/27ea291cd24e0b716db49ce39d4df8b767f55ad6/%20auto.config.ts#L10)

 auto.config.ts
```diff
@@ -7,7 +7,7 @@ import type { AutoRc } from 'auto';
 export default function rc(): AutoRc {
   return {
     baseBranch: 'main',
-    plugins: ['npm', 'released', ['exec', { afterShipIt: 'yarn publish-action' }]],
```
ghengeveld, line 10: This turned out to be unpredictable and hard to debug, so I just updated the publish-action script to always run on release and figure out the proper tag from the version in package.json.

tmeasday (approved):
LGTM

PR #835, opened 2023-10-11
Fix TurboSnap fallback dependency tracing for monorepos and pnpm projects

ghengeveld:


tmeasday (approved):
LGTM! Thanks for sorting this out.

How to best QA this?

ghengeveld:
It's been properly tested by the customer, and we have test coverage for the change, so I'm already confident about it.

PR #839, opened 2023-10-18
Drop official support for Node 14/16

ghengeveld:
Fixes AP-3324

This makes our CI config in line with the latest Node LTS, and limits official CLI support to Node 18+ (currently in maintenance). The CLI compilation target is still Node 16 (unchanged) to avoid breaking changes. The GitHub Action is now compiled for and configured to run on Node 20.

<details>
  <summary>📦 Published PR as canary version: <code>8.0.0--canary.839.6812102149.0</code></summary>
  <br />

  ```bash
  npm install chromatic@8.0.0--canary.839.6812102149.0
  # or 
  yarn add chromatic@8.0.0--canary.839.6812102149.0
  ```
</details>


.github/workflows/chromatic-main-and-prs.yml
```diff
@@ -15,7 +15,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-node@v3
         with:
-          node-version: 16
+          node-version: lts/*
```
thafryer, line 18: What are pros and cons to using LTS tag vs pinning the version? My biggest concern is that this tag can change the major versions when we may not be ready to.

ghengeveld: That's the point. If it fails, we'll know early on (before customers run into it).

thafryer: But how do we know any any changes we made to support new version still work on existing node version?

thafryer: Well I guess that's mitigated by saying we are dropping support for the older versions. I guess we would need to be okay with the small windows where multiple versions are still supported

ghengeveld: I discussed with Paul that we should support whatever version is still being maintained (i.e. isn't end-of-life) according to the [Node.js release schedule](https://github.com/nodejs/release#release-schedule). Right now that's Node 18+. We'll test against each LTS version (even numbers) as well as the latest.

For most tests the exact Node version doesn't matter, and so I'm specifying `lts/*` for those so that we won't have to update them over time. For some tests it _does_ matter, and those specify something different. The LTS versions are covered in `smoke-test-node18.yml` and `smoke-test-node20.yml`, while the latest (currently Node 21) is covered in `smoke-test-node-api.yml`.

tsup.config.ts
```diff
@@ -17,7 +17,7 @@ export default defineConfig((options) => [
     sourcemap: false,
     clean: true,
     platform: 'node',
-    target: 'node16',
+    target: 'node18', // Minimum supported (LTS) version
```
tmeasday, line 20: I'm not sure we shouldn't compile down to the node versions supported by the versions of SB we support?

tmeasday (changes requested):
I'm inclined to say the cost benefit of a few extra lines in our bundle to support node 16 is worth the benefit of not having to unexpectedly telling users on older versions of SB to upgrade node version.

I'm OK with saying in our README that it's not officially supported though. WDTY?

tsup.config.ts
ghengeveld: Fair enough. The difference is 2615 bytes.

.github/workflows/chromatic-main-and-prs.yml
thafryer: That sounds good. As long as we mitigate the tag changing and catching us unaware, I am fine!

tmeasday (approved):
LGTM

ghengeveld:
:rocket: PR was released in [`v8.0.0`](https://github.com/chromaui/chromatic-cli/releases/tag/v8.0.0) :rocket:

PR #841, opened 2023-10-22
Add workflow to require certain PR labels before merging

ghengeveld:
This adds a GitHub workflow for [chromaui/pr-label-checker-action](https://github.com/chromaui/pr-label-checker-action), which I created for this. It's configured to require one of `patch` / `minor` / `major` as well as either `release` or `skip-release` and disallow the `DO NOT MERGE` label.

The purpose of this is to enforce having the necessary labels for `auto` to work properly.

### DIFF
--- a/.github/workflows/check-labels.yml
+++ b/.github/workflows/check-labels.yml
@@ -1,0 +1,18 @@
+name: Check PR Labels
+on:
+  pull_request:
+    types: [opened, labeled, unlabeled, synchronize]
+
+jobs:
+  check-labels:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 20
+      - uses: chromaui/pr-label-checker-action@main
+        with:
+          one-of: |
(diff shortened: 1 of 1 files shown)


tmeasday (approved):
Sure! Although does auto not already have a feature like this? Either way, WFM

ghengeveld:
I suppose maybe we should use `auto pr-check` instead. It only checks for semver release labels though, whereas this setup is a bit more strict about it. The only benefit I see with `auto pr-check` is that it supports the [pr-body-labels](https://intuit.github.io/auto/docs/generated/pr-body-labels) plugin, which would make it a little bit more convenient to specify the proper label (by ticking a box). Not something I feel strongly about.

skitterm (commented):
Thanks for looking into this, Gert.

I don't have a strong opinion of auto vs [chromaui/pr-label-checker-action](https://github.com/chromaui/pr-label-checker-action), but I'd hope we can use the same solution for all of our various npm packages (`addon-visual-tests`, `test-archiver`, etc).

Issue #785, opened 2023-07-10
Support detection of pnpm-lock.yaml files for TurboSnap


Niznikr:
**Feature request**

TurboSnap currently detects and evaluates [yarn and npm lockfiles](https://www.chromatic.com/docs/turbosnap#avoid-re-testing-on-changes-to-package-control-files) to avoid re-testing everything if possible. This should also be supported for projects using `pnpm`.

**Proposed solution**

Update [this bit of code](https://github.com/chromaui/chromatic-cli/blob/main/node-src/lib/getDependentStoryFiles.ts#L14) and related tests to ensure `pnpm-lock.yaml` are supported.

vladsovetov:
I believe also should be changed [here](https://github.com/chromaui/chromatic-cli/blob/main/node-src/lib/findChangedDependencies.ts#L31)

On real prod repos where pnpm is used this affects a lot the expenses unfortunately and would be great to solve it ASAP. Meantime I think we will try to generate a yarn/npm lock file to avoid full execution of the snapshots

lapa182:
Since Next.js is heavily using pnpm in the examples, I wonder when the pnpm-lock file will be included on TurboSnap?

kevinlandsberg:
Please fix this!

PR #853, opened 2023-11-10
add support pnpm-lock files

lapa182:
This pull request adds support to the pnpm-lock.yaml files.

I've updated the tests to also support this change, please let me know if there's anything further to apply this change to chromatic-cli!

Relates to issue #785

### DIFF
--- a/node-src/lib/findChangedDependencies.test.ts
+++ b/node-src/lib/findChangedDependencies.test.ts
@@ -220,6 +220,26 @@ describe('findChangedDependencies', () => {
     );
   });
 
+  it('uses pnpm-lock.yaml if yarn.lock is missing', async () => {
+    findFiles.mockImplementation((file) => {
+      if (file.endsWith('yarn.lock'))
+        return Promise.resolve([file.replace('yarn.lock', 'pnpm-lock.yaml')]);
+      return Promise.resolve(file.startsWith('**') ? [file.replace('**', 'subdir')] : [file]);
+    });
+
+    checkoutFile.mockImplementation((ctx, commit, file) => Promise.resolve(`${commit}.${file}`));
+    buildDepTree.mockResolvedValue({ dependencies: {} });
+
+    await expect(findChangedDependencies(getContext(['A']))).resolves.toEqual([]);
+
+    expect(buildDepTree).toHaveBeenCalledWith(
+      '/root',
+      'A.subdir/package.json',
+      'A.subdir/pnpm-lock.yaml',
+      true
+    );
+  });
+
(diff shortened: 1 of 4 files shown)


tevanoff:
Thanks for the PR @lapa182! We'll take a look at this ASAP.

Obi-Dann:
I am trying these changes in our setup. I don't think it will work because `snyk-nodejs-lockfile-parser` does not support pnpm. It will likely report an error from https://github.com/chromaui/chromatic-cli/blob/602d90cfc31827d7a0f07b4243c5631eda4fe894/node-src/lib/getDependencies.ts#L28
```
Unknown lockfile pnpm-lock.yaml. Please provide either package-lock.json or yarn.lock.
```

https://github.com/snyk/nodejs-lockfile-parser/issues/111

lapa182:
@Obi-Dann do you think you could create a specific condition just for PNPM and use https://www.npmjs.com/package/@pnpm/lockfile-file to parse the lock file just for pnpm? 🤔

tevanoff:
Thanks @Obi-Dann! You're right, this won't change the current behavior since it won't be able to parse the pnpm lock file. It will fall back to looking at `package.json` which is what happens currently when it can't find the lock file at all. 

@lapa182 that pnpm library looks promising, but we haven't yet dug into it enough to know if it will give us what we need. We have a ticket on our end to take a closer look, but I can't say for sure when that will be prioritized. I'll see if I can get some traction on that. Thanks!

kevinlandsberg:
Come on, pls fix this finally!

milahu:
> `const headTree = await buildDepTreeFromFiles(rootPath, manifestPath, lockfilePath, includeDev);`

related: my [parse-package-lock](https://github.com/milahu/parse-package-lock) is a generic lockfile parser for npm, yarn, pnpm
to provide a generic interface to `@npmcli/arborist`, `@yarnpkg/parsers`, `@pnpm/lockfile-file`

but it does not return a tree of dependencies
its a stream parser that calls

```js
eventHandlers.enterPackage(packageData);
// recurse into dependencies of this package
eventHandlers.leavePackage(packageData);
```

i started this 2 years ago because i was not happy with snyk-nodejs-lockfile-parser
also because it does not return the resolved and integrity values of the locked packages

downsides: not used, not tested, abandoned project
i just found this again via my [pnpm-install-only](https://github.com/milahu/pnpm-install-only)
which is a generic package installer for npm, yarn, (not yet pnpm)
using the filesystem layout of pnpm to create a deep tree

jwir3:
This is out of sync with the `main` branch, and hasn't been updated in almost a year. Please re-open after rebasing if you still want to merge this.

PR #870, opened 2023-12-06
Implement file hashing for to-be-uploaded files

ghengeveld:
Uses xxhash-wasm for fast hashing of Storybook file contents. Chromatic's Storybook took 21ms to hash on my machine (Apple M1 Pro).

<details>
  <summary>📦 Published PR as canary version: <code>10.2.0--canary.870.7197051615.0</code></summary>
  <br />

  ```bash
  npm install chromatic@10.2.0--canary.870.7197051615.0
  # or 
  yarn add chromatic@10.2.0--canary.870.7197051615.0
  ```
</details>


node-src/lib/getFileHashes.ts
```diff
@@ -0,0 +1,71 @@
...
+          hash.update(buffer);
+          readIncremental(fd, hash);
+        }
+      });
+    });
+  });
+};
+
+export const getFileHashes = async (files: string[], dir = process.cwd()) => {
+  const limit = pLimit(48);
```
thafryer, line 57: This is a cool lib, but it's unclear what it does and why we're using until you look it up. Can we alias it or add a comment that gives more info. Also, what's the reason behind the choice of 48 as the concurrency limit?

node-src/lib/getFileHashes.ts
```diff
@@ -0,0 +1,71 @@
+import { close, open, read } from 'fs';
+import pLimit from 'p-limit';
+import { join } from 'path';
+import xxHashWasm, { XXHash, XXHashAPI } from 'xxhash-wasm';
+
+const hashFile = (buffer: Buffer, path: string, xxhash: XXHashAPI): Promise<string> => {
+  const BUFFER_SIZE = buffer.length;
+
+  return new Promise((resolve, reject) => {
+    const done = (fd: number, getResult: () => bigint) => {
```
thafryer, line 10: What does `fd` stand for?

thafryer (changes requested):

node-src/lib/parseArgs.ts
```diff
@@ -48,6 +48,7 @@ export default function parseArgs(argv: string[]) {
       --junit-report [filepath]        Write build results to a JUnit XML file. {buildNumber} will be replaced with the actual build number. [chromatic-build-{buildNumber}.xml]
       --list                           List available stories. This requires running a full build.
       --log-file [filepath]            Write log output to a file. Disable via --no-log-file. [chromatic.log]
+      --no-file-hashing                Disable file hashing. This will cause all files to be uploaded on every build.
```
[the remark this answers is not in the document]
ghengeveld: We don't want people to disable this feature. It's just an escape hatch for when file hashing doesn't work on their system, for whatever reason. It's obscure and technical on purpose.

node-src/lib/getFileHashes.ts
ghengeveld: This is an alternative to the resource pool used in the [original implementation](https://github.com/tripadvisor/npm-packages/blob/main/packages/fast-hash-file/throttled.ts#L9-L12). However that library is hardly used/maintained so I swapped it out for [p-limit](https://www.npmjs.com/package/p-limit) which I've used before and is very popular and well maintained. I kept the 48 concurrency limit for no particular reason other than that I have no idea what would be a better value. We could make it an environment variable I suppose.

node-src/lib/getFileHashes.ts
ghengeveld: File descriptor. Pretty common in the Node.js world.

ghengeveld: Luckily we also have types to clarify things if needed. The `fd` variable is _so_ common that I don't think spelling it out will actually improve readability. The [official docs](https://nodejs.org/en/learn/manipulating-files/working-with-file-descriptors-in-nodejs) use `fd` everywhere and I've never seen anyone _not_ use `fd`.

node-src/lib/getFileHashes.ts
```diff
@@ -0,0 +1,72 @@
...
+          hash.update(buffer);
+          readIncremental(fd, hash);
+        } else {
+          hash.update(buffer.subarray(0, bytesRead));
+          done(fd, () => hash.digest());
+        }
+      });
+    };
+
+    open(path, 'r', (openErr, fd) => {
```
tmeasday, line 34: Do we really need to use this oldschool node callback style of programming? Can we not import from `fs/promises` and make things a bit easier to understand?

tmeasday (commented):
I think I'm less concerned because we'll probably rarely touch that file again, but I sort of agree with @thafryer's basic point which is that the file is a bit impenetrable and could be written in a simpler way.

node-src/lib/getFileHashes.ts
andrewortwein: Perhaps it would be useful to leave a comment explaining _why_ this style was chosen? (I think Gert told me someone proved it to be faster than the alternative)

andrewortwein (approved):
I'll defer to the other comments here, but otherwise approve.

#️⃣#️⃣#️⃣👍🏻

ghengeveld:
:rocket: PR was released in [`v10.2.0`](https://github.com/chromaui/chromatic-cli/releases/tag/v10.2.0) :rocket:

PR #876, opened 2023-12-13
Replace `getUploadUrls` with `uploadBuild` mutation

ghengeveld:
Depends on https://github.com/chromaui/chromatic/pull/8049

This uses the new `uploadBuild` mutation which uses S3 presigned POST requests for uploading and enforces strict file size limitations.

The mutation no longer returns the storybook base domain, instead we query for `storybookUrl` on the build. All references to `cachedUrl` and `isolatorUrl` have been replaced with `storybookUrl`.

<details>
  <summary>📦 Published PR as canary version: <code>10.1.1--canary.876.7220703629.0</code></summary>
  <br />

  ```bash
  npm install chromatic@10.1.1--canary.876.7220703629.0
  # or 
  yarn add chromatic@10.1.1--canary.876.7220703629.0
  ```
</details>

### DIFF
--- a/node-src/index.test.ts
+++ b/node-src/index.test.ts
@@ -98,14 +98,16 @@ vi.mock('node-fetch', () => ({
       }
 
       if (query?.match('PublishBuildMutation')) {
-        if (variables.input.isolatorUrl.startsWith('http://throw-an-error')) {
-          throw new Error('fetch error');
-        }
-        publishedBuild = { id: variables.id, ...variables.input };
+        publishedBuild = {
+          id: variables.id,
+          ...variables.input,
+          storybookUrl: 'https://5d67dc0374b2e300209c41e7-pfkaemtlit.chromatic.com/',
+        };
         return {
           data: {
             publishBuild: {
               status: 'PUBLISHED',
+              storybookUrl: 'https://5d67dc0374b2e300209c41e7-pfkaemtlit.chromatic.com/',
             },
           },
         };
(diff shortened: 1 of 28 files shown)


tmeasday (commented):
LGTM! I'm unsure exactly why we need two mutations however, why is the metadata separated from the build data (if it ends up in the same place)?

tmeasday (approved):

ghengeveld:
> LGTM! I'm unsure exactly why we need two mutations however, why is the metadata separated from the build data (if it ends up in the same place)?

Because the metadata (CLI logs and such) is collected at the very end of the CLI process (when the build might've already finished). I suppose we could send part of the metadata along with the build upload itself (e.g. config files) but that would complicate things.

andrewortwein:
I am not able to run a build using the suggested command in the description. I have tried multiple projects, both existing and new, but the build always fails at the Storybook verifying step, just specifying that there was a failure to extract the Storybook. I also noticed that all of the smoke tests in this PR are failing, so I decided not to dig too far into the problems I'm seeing locally.

ghengeveld:
@andrewortwein Sorry, I was using my local version rather than the canary, so I hadn't noticed the canary was very old and not updating anymore. That turned out to be due to the `skip-release` label on this PR. I changed it and reran the release action, and now we have an up-to-date canary that I confirmed works correctly.

andrewortwein (approved):
LGTM 🆕👍🏻

ghengeveld:
:rocket: PR was released in [`v10.2.0`](https://github.com/chromaui/chromatic-cli/releases/tag/v10.2.0) :rocket:

PR #880, opened 2023-12-20
Add new Init command for Project Setup

thafryer:
This PR adds an init command to help customers set up Chromatic for usage within their projects. The API looks as such:
`npx chromatic init -f cypress|playwright`. The init script does the following
1. installs needed dependencies based on the user's provided framework.
2. Add a Chromatic Config file with some initial options.
3. Add a Npm script for Chromatic.

To QA:

`npx chromatic@10.4.0--canary.880.7629584432.0 init`

Project Type to QA

- [x] New Storybook Project
- [x] New Playwright Project, with Storybook Already Installed
- [x] New Playwright Project, without Storybook Already Installed
- [x] New Cypress Project, with Storybook Already Installed
- [x] New Cypress Project, without Storybook Already Installed

<details>
  <summary>📦 Published PR as canary version: <code>10.4.0--canary.880.7629584432.0</code></summary>
  <br />

  ```bash
  npm install chromatic@10.4.0--canary.880.7629584432.0
  # or 
  yarn add chromatic@10.4.0--canary.880.7629584432.0
  ```
</details>


bin-src/init.ts
```diff
@@ -0,0 +1,144 @@
...
+        ...(buildScriptName && {
+            buildScriptName
+        }),
+        onlyChanged: true,
+        skip: "dependabot/**"
+    });
+}
+
+export const installArchiveDependencies = async (packageJson: PackageJson, testFramework: TestFrameworkType) => {
+    let installArgs = ['-D', 'chromatic',`chromatic-${testFramework}`, 'storybook@next', '@storybook/addon-essentials@next', '@storybook/server-webpack5@next']
```
thafryer, line 56: Right now, this is intentionally installing SB 8.0 alpha. We can update before release to ensure we are on the latest.

package.json
```diff
@@ -196,6 +203,14 @@
...
     "zod": "^3.22.2"
   },
+  "peerDependencies": {
+    "@chromaui/test-archiver": "0.0.53"
+  },
+  "peerDependenciesMeta": {
+    "@chromaui/test-archiver": {
+      "optional": true
+    }
+  },
```
thafryer, line 213: This was a pre-cursor to the new packages. We can tweak this to be those once ready.

node-src/lib/checkPackageJson.ts
```diff
@@ -22,7 +22,7 @@ export default async function checkPackageJson({
   if (!options.interactive) return;
 
   try {
-    const json = { ...packageJson };
+    const { readme, _id, ...json } = packageJson;
```
thafryer, line 25: These fields (`readme` & `_id`) are added to package.json if we don't remove them here.

bin-src/init.ts
```diff
@@ -0,0 +1,144 @@
...
+    return await findUp(['.storybook/main.ts', '.storybook/main.js', '.storybook/main.tsx', '.storybook/main.jsx', '.storybook/main.mjs', '.storybook/main.cjs'])
+}
+
+export const addChromaticScriptToPackageJson = async ({ packageJson, packagePath }) => {
+    try {
+        const json = {
+            ...packageJson,
+            scripts: {
+                ...packageJson?.scripts,
+                chromatic: `npx chromatic`
```
tmeasday, line 35: Don't we recommend `npx chromatic@latest` now?

bin-src/init.ts
```diff
@@ -0,0 +1,144 @@
...
+    }
+}
+
+export const createChromaticConfigFile = async ({configFile, buildScriptName = null}) => {
+    await writeFile(configFile, {
+        autoAcceptChanges: "main",
+        ...(buildScriptName && {
+            buildScriptName
+        }),
+        onlyChanged: true,
```
tmeasday, line 50: I guess we don't want this for E2E right?

bin-src/init.ts
```diff
@@ -0,0 +1,144 @@
...
+        }
+        await writeFile(packagePath, json, { spaces: 2 });
+    } catch (e) {
+        console.warn(e)
+    }
+}
+
+export const createChromaticConfigFile = async ({configFile, buildScriptName = null}) => {
+    await writeFile(configFile, {
+        autoAcceptChanges: "main",
```
tmeasday, line 46: I don't think this is recommended

bin-src/init.ts
tmeasday, line 50: I guess we don't want this for E2E right?

bin-src/init.ts
```diff
@@ -0,0 +1,144 @@
...
+const intializeChromatic = async ({testFramework, packageJson, packagePath}: {testFramework: TestFrameworkType, packageJson: PackageJson, packagePath: string }) => {
+    await addChromaticScriptToPackageJson({ packageJson, packagePath });
+    await createChromaticConfigFile({configFile: 'chromatic.config.json'})
+    switch (testFramework) {
+        case TestFramework.CYPRESS:
+            await installArchiveDependencies(packageJson, TestFramework.CYPRESS)
+            break;
+        case TestFramework.PLAYWRIGHT:
+            await installArchiveDependencies(packageJson, TestFramework.PLAYWRIGHT)
+            break;
```
tmeasday, line 75:
```suggestion
        case TestFramework.CYPRESS:
        case TestFramework.PLAYWRIGHT:
            await installArchiveDependencies(packageJson, TestFramework.PLAYWRIGHT)
            break;
```

bin-src/init.ts
tmeasday: We need to check if the user already has SB installed and use the same version they have specified in `package.json` if so.

thafryer: @tmeasday We do this step later on lines 57-60.

bin-src/init.ts
thafryer: installArchiveDependencies takes in a test framework that determines which dependency will get installed. So, that's why they are separate cases. Am I missing something from this suggestion?

skitterm:
@thafryer can you include the API change in the PR description (aka `npx chromatic init --playwright|cypress`) and any other ramifications this PR will have?

thafryer:
> @thafryer can you include the API change in the PR description (aka `npx chromatic init --playwright|cypress`) and any other ramifications this PR will have?

@skitterm I updated the description!

bin-src/init.ts
```diff
@@ -0,0 +1,142 @@
...
+    if(sbConfigPath && sbVersion) {
+        installArgs = ['-D', 'chromatic',`chromatic-${testFramework}`, `@storybook/server-webpack5@${sbVersion}`]
+    }
+    const installCommand = await getPackageManagerInstallCommand(installArgs)
+    await execaCommand(installCommand)
+}
+
+const intializeChromatic = async ({testFramework, packageJson, packagePath}: {testFramework: TestFrameworkType, packageJson: PackageJson, packagePath: string }) => {
+    await addChromaticScriptToPackageJson({ packageJson, packagePath });
+    await createChromaticConfigFile({configFile: 'chromatic.config.json'})
```
skitterm, line 66: This `chromatic.config.json` file isn't currently used/needed for E2E. Could we skip generating this file in those cases for now?

bin-src/init.ts
```diff
@@ -0,0 +1,142 @@
...
+
+export async function main(argv: string[]) {
+    const { flags } = meow(
+        `
+        Usage
+          $ chromatic init [-f|--framework]
+
+        Options
+          --framework, -f <framework     Test Framework that you are aiming to use with Chromatic. (default: 'storybook')
+        `,
```
skitterm, line 88: It'd sure be nice for brevity/DX if it was `npx chromatic init --playwright` instead of `npx chromatic init -f playwright`.  Hopefully people would copy-paste but if not it's one more thing to type. 

I guess conceptually the `-f` is nice since it's really a key-value pair, so take that for what it's worth.

bin-src/init.ts
```diff
@@ -0,0 +1,142 @@
...
+            {
+                type: flags.framework ? null : 'select',
+                name: 'testFramework',
+                message: 'What testing framework are you using?',
+                choices: [
+                    {title: 'Storybook', value: TestFramework.STORYBOOK},
+                    {title: 'Playwright', value: TestFramework.PLAYWRIGHT},
+                    {title: 'Cypress', value: TestFramework.CYPRESS},
+                ],
+                initial: 0
```
skitterm, line 130: @thafryer does this mean that if they don't pass the `-f` flag, we'll prompt them with this menu instead? That's pretty nice.

thafryer: Yes, that's correct!

bin-src/init.ts
thafryer: The Config file or CLI flags are needed for configuration of Chromatic CLI/Action.

We are trying to promote the config file moving forward as the way to configure Chromatic. In the future, this file would automatically be configured with setting that we recommend like the one here.

In this case, if we take this out, Chromatic would run on any dependabot PRs that open. Customers complain about this often enough that we codify this in our docs.

thafryer: TLDR: If you're running Chromatic CLI, you will need either CLI flags or the config.

bin-src/init.ts
thafryer: Yeah, this one I struggled with because we either are adding a new CLI flag for every framework or an option on a single CLI flag. I could go either way here. 

Moreover, having multiple flags means we need to add extra validation if they pass both —cypress and —playwright. We would also have to do the same for additional frameworks.

skitterm: Good point. @tevanoff @byebrianwong what are your thoughts on this?

skitterm (approved):

bin-src/init.ts
tevanoff: Good question! I don't feel strongly either way since it's a one-off command. One flag with different options seems fine to me.

bin-src/init.ts
```diff
@@ -0,0 +1,142 @@
...
+};
+type TestFrameworkType = typeof TestFramework[keyof typeof TestFramework];
+
+export const getPackageManagerInstallCommand = async (args: string[]) => {
+    return await getCliCommand(parseNi, args, { programmatic: true });
+};
+
+const getSBConfigFilePath = async () => {
+    // Walks up directory tree to find nearest Storybook config file.
+    return await findUp(['.storybook/main.ts', '.storybook/main.js', '.storybook/main.tsx', '.storybook/main.jsx', '.storybook/main.mjs', '.storybook/main.cjs'])
```
tevanoff, line 26: Do we need to support `storybookConfigDir` in case the dir is not at the default `.storybook` location?

bin-src/init.ts
```diff
@@ -0,0 +1,142 @@
...
+}
+
+export const addChromaticScriptToPackageJson = async ({ packageJson, packagePath }) => {
```
tevanoff, line 29: Do we have other tooling that edits `package.json` directly? This just feels like something that could be prone to errors.

storybook-out/preview-stats.trimmed.json
```diff
@@ -0,0 +1,18 @@
+{
```
tevanoff, line 1: This looks like a mistaken addition.

bin-src/init.ts
thafryer: Yes, we do [here](https://github.com/chromaui/chromatic-cli/blob/main/node-src/lib/checkPackageJson.ts). It's pretty simple and is just writing to file. If they already have a chromatic script, it's overwritten. If they do not want the change, it is easy to revert it to what it was before.

bin-src/init.ts
thafryer: Yeah, that's a valid point. This is meant to be a part of the logic check to determine if a customer has Storybook already configured. The check looks to see if the user has a storybook `main.js|ts` file and the package versions for the `storybook` package in their package.json. My instincts say checking for the config file might be more complex than we desire, as raised by your point here.

Checking if the `@storybook/addon-essentials` or `storybook` packages are in the package.json file could be satisfactory and get us there without the complexity.

bin-src/init.ts
thafryer: I ended up leaving as is for now.

bin-src/init.ts
thafryer: @tevanoff I decided to check whether `@storybook/addon-essentials` and `storybook` packages are included in package.json. It's a simpler heuristic and easily as strong indicator of whether the packages need to be installed.

tevanoff: Sounds good!

bin-src/init.ts
```diff
@@ -0,0 +1,134 @@
...
+            buildScriptName
+        })
+    });
+}
+
+export const installArchiveDependencies = async (packageJson: PackageJson, testFramework: TestFrameworkType) => {
+    let installArgs = ['-D', 'chromatic',`chromatic-${testFramework}`, 'storybook@latest', '@storybook/addon-essentials@latest', '@storybook/server-webpack5@latest']
+    const storybookVersion = packageJson?.devDependencies?.storybook || packageJson?.dependencies?.storybook
+    const essentialsVersion = packageJson?.devDependencies?.['@storybook/addon-essentials'] || packageJson?.dependencies?.['@storybook/addon-essentials']
+    if(storybookVersion && essentialsVersion) {
```
tevanoff, line 50: It would be possible for somebody to have `storybook` installed, but not `addon-essentials`, right? If that's the case, this will upgrade their `storybook` version rather than installing `addon-essentials` at the current sb version.

bin-src/init.ts
tmeasday: Oh, yeah right, but you could pass `testFramework`, but really its just a qn of style..

bin-src/init.ts
```diff
@@ -0,0 +1,145 @@
...
+    }
+}
+
+export const createChromaticConfigFile = async ({configFile, buildScriptName = null}) => {
+    await writeFile(configFile, {
+        ...(buildScriptName && {
+            buildScriptName
+        })
+    });
+}
```
tevanoff, line 44: Minor change
```suggestion
}

```

tevanoff (approved):
Looks good!

ghengeveld:
:rocket: PR was released in [`v10.4.0`](https://github.com/chromaui/chromatic-cli/releases/tag/v10.4.0) :rocket:

PR #882, opened 2024-01-09
Add `--playwright` & `--cypress` flags

tmeasday:
When set, build storybook by directly invoking the binary from `@chromaui/test-archiver`.

This is a POC, we will move this binary into a `chromatic-playwright` package soon.

TODO:
 - [x] Add cypress flag too
 - [x] Ensure the flag works in the action
 - [x] Throw errors if incompatible flags are used
 - [x] Tests

Some notes:
 
 - This won't work if you symlink the `chromatic` package into a test project (due to symlinking and resolution). Probably just use the canary to test it out (`chromatic@10.2.1--canary.882.7460727945.0`)
 
  - For this to work in it's current form, you need to add the following export to the `@chromaui/archive-storybook` `package.json:exports` field:

```
    "./package.json": {
      "require": "./package.json",
      "import": "./package.json"
    }
```

I did this manually inside `node_modules` of my test project for now.

For some reason I couldn't `require.resolve('@chromaui/archive-storybook')` directly. @ndelangen may know more about this.

- We'll want to resolve things in `chromatic-playwright` really, so we'd be adding the export to that package once we setup the monorepo.

- This calls `node path/to/@chromaui/archive-storybook/dist/bin/build-archive-storybook.js -o ...`, which then calls `node path/to/@storybook/cli/dist/bin/storybook.js -c path/to/config`. There might be a way to avoid the double node invocation. Then again, it's probably not a big deal.

<details>
  <summary>📦 Published PR as canary version: <code>10.5.0--canary.882.7634992282.0</code></summary>
  <br />

  ```bash
  npm install chromatic@10.5.0--canary.882.7634992282.0
  # or 
  yarn add chromatic@10.5.0--canary.882.7634992282.0
  ```
</details>

### DIFF
--- a/node-src/lib/getE2eBinPath.ts
+++ b/node-src/lib/getE2eBinPath.ts
@@ -1,0 +1,18 @@
+import { Context } from '../types';
+import missingDependency from '../ui/messages/errors/missingDependency';
+import { exitCodes, setExitCode } from './setExitCode';
+import { failed } from '../ui/tasks/build';
+
+export function getE2eBinPath(ctx: Context, flag: 'playwright' | 'cypress') {
+  const dependencyName = `chromatic-${flag}`;
+  try {
+    return require.resolve(`${dependencyName}/bin/build-archive-storybook`);
+  } catch (err) {
+    if (err.code === 'MODULE_NOT_FOUND') {
+      ctx.log.error(missingDependency({ dependencyName, flag }));
+      setExitCode(ctx, exitCodes.MISSING_DEPENDENCY, true);
+      throw new Error(failed(ctx).output);
+    }
+    throw err;
+  }
+}
--- a/package.json
+++ b/package.json
@@ -196,8 +196,8 @@
     "zod": "^3.22.2"
   },
   "peerDependencies": {
-    "chromatic-playwright": "^0.3.1",
-    "chromatic-cypress": "^0.3.2"
+    "chromatic-playwright": "^1.0.0",
+    "chromatic-cypress": "^1.0.0"
   },
   "peerDependenciesMeta": {
     "chromatic-playwright": {
--- a/package.json
+++ b/package.json
@@ -196,8 +196,8 @@
     "zod": "^3.22.2"
   },
   "peerDependencies": {
-    "chromatic-playwright": "^0.3.1 || ^1.0.0",
-    "chromatic-cypress": "^0.3.2 || ^1.0.0"
+    "chromatic-playwright": "^0.4.0 || ^1.0.0",
+    "chromatic-cypress": "^0.4.0 || ^1.0.0"
   },
   "peerDependenciesMeta": {
     "chromatic-playwright": {
--- a/package.json
+++ b/package.json
@@ -196,8 +196,8 @@
     "zod": "^3.22.2"
   },
   "peerDependencies": {
-    "chromatic-playwright": "^1.0.0",
-    "chromatic-cypress": "^1.0.0"
+    "chromatic-playwright": "^0.3.1 || ^1.0.0",
+    "chromatic-cypress": "^0.3.2 || ^1.0.0"
   },
   "peerDependenciesMeta": {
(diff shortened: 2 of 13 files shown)


ndelangen:
You can simplify:
```
    "./package.json": {
      "require": "./package.json",
      "import": "./package.json"
    }
```
to:
```
    "./package.json": "./package.json"
```

tmeasday:
@ndelangen do you know why I could `require.resolve()` `@chromaui/archive-storybook/package.json` but not `@chromaui/archive-storybook`, given there is a `exports` field already containing `'.'`?

https://github.com/chromaui/archive-storybook/blob/26a2269c17b10ecf693d74c296b8254b1ece807c/package.json#L14-L20

ndelangen:
No, it should resolve to the full `require` path from here:
https://github.com/chromaui/archive-storybook/blob/main/package.json#L16C8-L16C36

So the outcome would be:
`/path/to/node_modules/@chromaui//archive-storybook/dist/index.js`

What error did you get? "could not resolve"?

tmeasday:
Yes, the standard `MODULE_NOT_FOUND` that we check for in this PR.

node-src/lib/getE2eBinPath.ts
```diff
@@ -0,0 +1,18 @@
...
+    return require.resolve(`${dependencyName}/bin/build-archive-storybook`);
+  } catch (err) {
+    if (err.code === 'MODULE_NOT_FOUND') {
+      ctx.log.error(missingDependency({ dependencyName, flag }));
+      setExitCode(ctx, exitCodes.MISSING_DEPENDENCY, true);
+      throw new Error(failed(ctx).output);
+    }
+    throw err;
+  }
+}
```
thafryer, line 18: Are we adding the new packages as optional peer dependencies in this PR?

tmeasday: This is a good qn!

thafryer: The only piece that needs to be added here, based on scoped work, is the optional peer dependencies.  However, is it crucial for all of this to work?

thafryer (approved):

node-src/lib/getE2eBinPath.ts
tmeasday: Only in some very specific cases but we may as well just do it! I will get to it

package.json
```diff
@@ -195,6 +195,18 @@
     "zen-observable": "^0.8.15",
     "zod": "^3.22.2"
   },
+  "peerDependencies": {
+    "chromatic-playwright": "^0.3.1",
+    "chromatic-cypress": "^0.3.2"
```
tevanoff, line 200: If we're going to cut 1.0.0 versions of these when we go GA, we might want to include that here.

tevanoff (approved):
Left one suggestion on the peer deps, but otherwise looks good!

package.json
tmeasday: OK, we can do that. It'll still be possible to test (with incorrect versions) against the current versions if we do that, but it'll warn in some package managers, and error in npm: https://www.notion.so/chromatic-ui/Peer-Dependency-Investigations-ae77792bae6e420bbe860916bec10cbe?pvs=4#176332d500194cdfacadf1efd927c815

tmeasday: Oh, I misinterpreted, you mean to have both versions supported 👍

tmeasday:
QA-ed on my test app w/ `chromatic-playwright@0.4.0` and works well. The only issue I found was that because I had no `scripts` at all in my `package.json`, so I made this change: https://github.com/chromaui/chromatic-cli/pull/882/commits/42d2d2bdf5668fd48e18d22b76bb5b5a007923ba

ghengeveld:
:rocket: PR was released in [`v10.5.0`](https://github.com/chromaui/chromatic-cli/releases/tag/v10.5.0) :rocket:

PR #890, opened 2024-01-16
Add missing `skip` option to configuration schema

ghengeveld:
This option was documented and supposed to be available, but was omitted in the configuration schema. This adds the option so it can be used through `chromatic.config.json`.

### DIFF
--- a/node-src/lib/getConfiguration.ts
+++ b/node-src/lib/getConfiguration.ts
@@ -26,6 +26,7 @@ const configurationSchema = z
 
     buildScriptName: z.string(),
     outputDir: z.string(),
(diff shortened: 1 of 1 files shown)


cgbl-90 (approved):

tmeasday:
@ghengeveld this was an intentional omission. The rationale was why would you always want to skip? But perhaps folks are using specific config files for specific situations. In which case we might want to consider adding some other once-off flags.

ghengeveld:
Skip can take a branch name, in which case it makes sense to specify that in the config file. E.g. to skip dependabot branches.

tmeasday:
Ahh, great point!

PR #892, opened 2024-01-17
Fix potential zip upload error when deduping files on a very large Storybook

ghengeveld:
When publishing a Storybook with > 1000 files, the processing happens in batches. When using the `zip` option, each batch returns a `zipTarget`. However, if there are no files to upload for a batch, `zipTarget` will be `undefined`. Previously the CLI would use whatever `zipTarget` was returned last, even if that `zipTarget` is `undefined`. In this case, the CLI falls back to uploading individual files, but the `sentinelUrl` for unpacking the zip file will still be present and the CLI will wait for the unpack that will never happen, eventually timing out with a failed build.

This PR fixes the `zipTarget` by ignoring batches that have no `zipTarget`. Additionally, it deduplicates `sentinelUrls` because each batch returns the same basic `sentinelUrl`, with only a deviating query param (signature). We only really need to await one of them. Also, logging is improved to clarify which type of sentinel file we're dealing with (`files-copied` or `zip-unpacked`).

<details>
  <summary>📦 Published PR as canary version: <code>10.3.1--canary.892.7557680958.0</code></summary>
  <br />

  ```bash
  npm install chromatic@10.3.1--canary.892.7557680958.0
  # or 
  yarn add chromatic@10.3.1--canary.892.7557680958.0
  ```
</details>

### DIFF
--- a/node-src/tasks/upload.test.ts
+++ b/node-src/tasks/upload.test.ts
@@ -6,7 +6,13 @@ import { default as compress } from '../lib/compress';
 import { getDependentStoryFiles as getDepStoryFiles } from '../lib/getDependentStoryFiles';
 import { findChangedDependencies as findChangedDep } from '../lib/findChangedDependencies';
 import { findChangedPackageFiles as findChangedPkg } from '../lib/findChangedPackageFiles';
-import { calculateFileHashes, validateFiles, traceChangedFiles, uploadStorybook } from './upload';
+import {
+  calculateFileHashes,
+  validateFiles,
+  traceChangedFiles,
+  uploadStorybook,
+  waitForSentinels,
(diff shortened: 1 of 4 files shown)


node-src/tasks/upload.test.ts
```diff
@@ -335,92 +341,6 @@ describe('uploadStorybook', () => {
     expect(ctx.uploadedFiles).toBe(2);
   });
 
-  it('batches calls to uploadBuild mutation', async () => {
```
ghengeveld, line 338: This test appeared twice in this file, as a result of a bad merge. So I removed one of them.

andrewortwein (approved):
LGTM👍🏻

PR #910, opened 2024-02-02
Add Fallback getGitInfo if we cannot find the origin URL

thafryer:
This PR adds a fallback for situations where we cannot get an origin URL to create a project slug. Without this, there's a downstream issue in the Visual Test Addon, where it thinks the project does not have git initialized, but it does. The project does not have an origin set yet.

To QA:
1. Run the tests.
2. Install this addon version (`1.0.4--canary.172.95c7a87.0`) in a repository where Git has been initialized, but there's no origin set yet.
3. Confirm that you see the "Git not detected" message.
4. Then, initialize the repo and make your first commit.
5. Confirm that you **_do not_** see the "Git not detected" message.

<details>
  <summary>📦 Published PR as canary version: <code>10.7.1--canary.910.7751165821.0</code></summary>
  <br />

  ```bash
  npm install chromatic@10.7.1--canary.910.7751165821.0
  # or 
  yarn add chromatic@10.7.1--canary.910.7751165821.0
  ```
</details>

### DIFF
--- a/node-src/index.test.ts
+++ b/node-src/index.test.ts
@@ -6,7 +6,7 @@ import fetchDefault from 'node-fetch';
 import { Readable } from 'stream';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
-import { runAll } from '.';
+import { runAll, getGitInfo } from '.';
 import * as git from './git/git';
 import { DNSResolveAgent } from './io/getDNSResolveAgent';
 import getEnv from './lib/getEnv';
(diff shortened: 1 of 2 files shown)


tmeasday (approved):
Makes sense given the current code:

https://github.com/chromaui/chromatic-cli/blob/b5e5cc1124ef8e2d61872fce82fc6cf3cbdc8e72/node-src/index.ts#L274

Although I do wonder if we should be setting `slug` to `null` rather than `''` there (let's not change in this PR).

Did the test you added fail outright before?

thafryer:
> Makes sense given the current code:
> 
> https://github.com/chromaui/chromatic-cli/blob/b5e5cc1124ef8e2d61872fce82fc6cf3cbdc8e72/node-src/index.ts#L274
> 
> Although I do wonder if we should be setting `slug` to `null` rather than `''` there.
> 
> Did the test you added fail outright before?

That's correct. When I mock the getSlug function to throw an error, the test fails.

thafryer:
> Although I do wonder if we should be setting slug to null rather than '' there.

We could do this. We would just have to update the logic here and in the VTA addon that expects the slug to be a string.

ghengeveld:
:rocket: PR was released in [`v10.7.1`](https://github.com/chromaui/chromatic-cli/releases/tag/v10.7.1) :rocket:

PR #912, opened 2024-02-05
Be smarter about comparing lock files

tmeasday:
- Don't consider lockfiles at all if there are no traced changes to package metadata
- Otherwise, after figuring out what lockfile pairs exist, filter out the ones that don't have diffs.

In practice this makes a big difference:
 - With no changes on the Chromatic monorepo it goes from 10-12s to 0s.
 - With only `./package.json` changed, it goes to 1s.

<details>
  <summary>📦 Published PR as canary version: <code>10.9.0--canary.912.7810229911.0</code></summary>
  <br />

  ```bash
  npm install chromatic@10.9.0--canary.912.7810229911.0
  # or 
  yarn add chromatic@10.9.0--canary.912.7810229911.0
  ```
</details>

### DIFF
--- a/node-src/lib/findChangedDependencies.test.ts
+++ b/node-src/lib/findChangedDependencies.test.ts
@@ -4,6 +4,7 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import * as git from '../git/git';
 import { findChangedDependencies } from './findChangedDependencies';
 import TestLogger from './testLogger';
+import { Context } from '..';
 
 vi.mock('snyk-nodejs-lockfile-parser');
 vi.mock('yarn-or-npm');
@@ -25,25 +26,39 @@ afterEach(() => {
   buildDepTree.mockReset();
 });
 
-const getContext: any = (baselineCommits: string[], options = {}) => ({
-  log: new TestLogger(),
-  git: { baselineCommits },
-  options,
-});
+const getContext = (
+  input: Partial<Omit<Context, 'git' | 'options'>> & {
+    git: Partial<Context['git']>;
+    options?: Partial<Context['options']>;
+  }
+) =>
+  ({
+    log: new TestLogger(),
+    options: {},
+    ...input,
+  } as Context);
+
+const AMetadataChanges = [{ changedFiles: ['package.json'], commit: 'A' }];
 
 describe('findChangedDependencies', () => {
-  it('returns nothing given no baselines', async () => {
-    buildDepTree.mockResolvedValue({ dependencies: {} });
+  it('returns nothing given no changes', async () => {
+    const context = getContext({ git: { packageMetadataChanges: [] } });
 
-    await expect(findChangedDependencies(getContext([]))).resolves.toEqual([]);
+    await expect(findChangedDependencies(context)).resolves.toEqual([]);
+    expect(checkoutFile).not.toHaveBeenCalled();
+    expect(buildDepTreeFromFiles).not.toHaveBeenCalled();
   });
 
-  it('returns nothing when dependency tree is empty', async () => {
-    checkoutFile.mockResolvedValueOnce('A.package.json');
-    checkoutFile.mockResolvedValueOnce('A.yarn.lock');
-    buildDepTree.mockResolvedValue({ dependencies: {} });
+  it('returns nothing given no changes to found package metadata', async () => {
+    const context = getContext({
+      // Only detected metadata change is to a file that's not on disk
+      git: { packageMetadataChanges: [{ changedFiles: ['foo/package.json'], commit: 'A' }] },
+    });
 
-    await expect(findChangedDependencies(getContext(['A']))).resolves.toEqual([]);
+    await expect(findChangedDependencies(context)).resolves.toEqual([]);
+
+    expect(checkoutFile).not.toHaveBeenCalled();
+    expect(buildDepTreeFromFiles).not.toHaveBeenCalled();
   });
 
   it('returns nothing when dependency tree is unchanged', async () => {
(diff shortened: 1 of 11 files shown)


node-src/lib/findChangedDependencies.ts
```diff
@@ -60,29 +60,44 @@ export const findChangedDependencies = async (ctx: Context) => {
...
-  const untracedCount = pathPairs.length - tracedPairs.length;
-  if (untracedCount) {
-    ctx.log.debug(`Skipping ${untracedCount} manifest/lockfile pairs due to --untraced`);
+  // Now filter out any pairs that don't have git changes, or for which the manifest is untraced
+  const filteredPathPairs = pathPairs
+    .map(([manifestPath, lockfilePath]) => {
+      const commits = packageMetadataChanges
+        .filter(
+          ({ changedFiles }) =>
+            changedFiles.includes(manifestPath) || changedFiles.includes(lockfilePath)
```
ghengeveld, line 69: A single `some` operation would be more efficient. The lockfile is more likely to have changed than the manifest file.

node-src/lib/findChangedDependencies.ts
```diff
@@ -60,29 +60,44 @@ export const findChangedDependencies = async (ctx: Context) => {
...
+        )
+        .map(({ commit }) => commit);
+
+      return [manifestPath, lockfilePath, commits] as const;
+    })
+    .filter(([, , commits]) => commits.length > 0)
+    .filter(([manifestPath]) => !untraced.some((glob) => matchesFile(glob, manifestPath)));
```
ghengeveld, line 76: Why not a single filter?

node-src/lib/findChangedDependencies.ts
```diff
@@ -60,29 +60,44 @@ export const findChangedDependencies = async (ctx: Context) => {
...
-    ctx.log.debug(`Skipping ${untracedCount} manifest/lockfile pairs due to --untraced`);
+  // Now filter out any pairs that don't have git changes, or for which the manifest is untraced
+  const filteredPathPairs = pathPairs
+    .map(([manifestPath, lockfilePath]) => {
+      const commits = packageMetadataChanges
+        .filter(
+          ({ changedFiles }) =>
+            changedFiles.includes(manifestPath) || changedFiles.includes(lockfilePath)
+        )
+        .map(({ commit }) => commit);
```
ghengeveld, line 71: We should probably dedupe `commits`.

package.json
```diff
@@ -232,6 +232,7 @@
     ]
   },
   "docs": "https://www.chromatic.com/docs/cli",
+  "packageManager": "yarn@1.22.19+sha256.732620bac8b1690d507274f025f3c6cfdc3627a84d9642e38a07452cc00e0f2e",
```
ghengeveld, line 235: Is this intentional?

ghengeveld (approved):
Seems like the right thing to do.

node-src/lib/findChangedDependencies.test.ts
```diff
@@ -220,6 +290,28 @@ describe('findChangedDependencies', () => {
     );
   });
 
+  it('ignores lockfile changes if metadata file is untraced', async () => {
+    findFiles.mockImplementation((file) => {
```
skitterm, line 294: I'm glad you thought of this case.

node-src/lib/findChangedDependencies.ts
```diff
@@ -60,29 +60,44 @@ export const findChangedDependencies = async (ctx: Context) => {
...
-  const tracedPairs = pathPairs.filter(([manifestPath, lockfilePath]) => {
-    if (untraced.some((glob) => matchesFile(glob, manifestPath))) return false;
-    if (untraced.some((glob) => matchesFile(glob, lockfilePath))) return false;
-    return true;
-  });
-  const untracedCount = pathPairs.length - tracedPairs.length;
-  if (untracedCount) {
-    ctx.log.debug(`Skipping ${untracedCount} manifest/lockfile pairs due to --untraced`);
+  // Now filter out any pairs that don't have git changes, or for which the manifest is untraced
+  const filteredPathPairs = pathPairs
```
skitterm, line 64:
```suggestion
  const filteredPathPairs = metadataPathPairs
```

skitterm (approved):

node-src/lib/findChangedDependencies.ts
tmeasday: I believe the original list is one per commit?

package.json
tmeasday: Yeah, `lint-staged` was failing for me without it. Can pull out to another PR if you like.

node-src/lib/findChangedDependencies.ts
tmeasday: Sorry @skitterm I'm not quite sure what you mean by this comment - are you suggesting we change the name of the first `pathPairs` variable?

package.json
ghengeveld: Nah just checking.

node-src/lib/findChangedDependencies.ts
ghengeveld: Wouldn't it be possible to have multiple `packageMetadataChanges` for the same build (i.e. commit)? E.g. when there's two package.json files in the repo and they both changed.

tmeasday: No, because the changes are the list of *all* the metadata changes for that commit. So it'd be an entry `{ changedFiles: ['package.json', 'subdir/package.json'], commit: 'xyz' }`. 

See: https://github.com/chromaui/chromatic-cli/blob/5436e2db7c0a1227799c5b4cf2f7bb7c843751ba/node-src/tasks/gitInfo.ts#L197-L208

Although... maaaaybe it is possible to have >1 build for the same commit?

ghengeveld:
:rocket: PR was released in [`v10.9.0`](https://github.com/chromaui/chromatic-cli/releases/tag/v10.9.0) :rocket:

PR #917, opened 2024-02-08
Directly execute build-archive-storybook if we can't resolve it

tmeasday:
In the GitHub action, we cannot import peer dependencies (or anything really). However we should be able to `npm exec <bin-command>` inside the user's project dir, if the package is installed.

<details>
  <summary>📦 Published PR as canary version: <code>10.9.1--canary.917.7824503990.0</code></summary>
  <br />

  ```bash
  npm install chromatic@10.9.1--canary.917.7824503990.0
  # or 
  yarn add chromatic@10.9.1--canary.917.7824503990.0
  ```
</details>

### DIFF
--- a/action-src/main.ts
+++ b/action-src/main.ts
@@ -135,6 +135,9 @@ async function run() {
     process.chdir(path.join(process.cwd(), workingDir || ''));
 
     const output = await runNode({
+      options: {
+        inAction: true,
+      },
       flags: {
         allowConsoleErrors: maybe(allowConsoleErrors, false),
         autoAcceptChanges: maybe(autoAcceptChanges),
--- a/node-src/index.test.ts
+++ b/node-src/index.test.ts
@@ -38,8 +38,11 @@ beforeEach(() => {
 vi.mock('dns');
 vi.mock('execa');
 
-vi.mock('./lib/getE2eBinPath', () => ({
-  getE2eBinPath: () => 'path/to/bin',
+// NOTE: we'd prefer to mock the require.resolve() of `@chromatic-com/playwright/..` but
+// vitest doesn't allow you to do that.
+const mockedBuildCommand = 'mocked build command';
+vi.mock('./lib/getE2eBuildCommand', () => ({
+  getE2eBuildCommand: () => mockedBuildCommand,
 }));
 
 const execaCommand = vi.mocked(execaDefault);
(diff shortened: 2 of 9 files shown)


thafryer (approved):
As an eventual follow-up, I'll note a ticket to add some playwright and cypress tests to CI. So, we know if anything breaks as early as possible.

tmeasday:
@tevanoff I managed to get the action to run in the wrong dir and show the error [here](https://github.com/tmeasday/playwright-docs/actions/runs/7824418960/job/21346944702):

image

tevanoff (approved):
All worked in a few different test cases for me! 💯

shrink:
I think this change has caused some important errors to be swallowed. I was encountering an error that presented as follows:

```console
Building your Storybook
    → Running command: npm exec build-archive-storybook --output-dir /tmp/chromatic--1880-VEzc3CPix51U --webpack-stats-json /tmp/chromatic--1880-VEzc3CPix51U
    → [*                   ]
✖ Failed to import `@chromatic-com/playwright`, is it installed in `package.json`?
    → Command failed: npm exec build-archive-storybook --output-dir /tmp/chromatic--1880-VEzc3CPix51U --webpack-stats-json /tmp/chromatic--1880-VEzc3CPix51U

ℹ To run `chromatic --playwright` you must have `@chromatic-com/playwright` installed.
```

So I spent a while debugging my dependencies and trying to find out why `@chromatic-com/playwright` wasn't available. Anyway, after a while I decided that the error message must be wrong and dug in further by running the `npm exec build-archive-storybook` command in my Workflow myself which led me to the real error:

```console
Error: Chromatic archives directory cannot be found: /home/runner/work/_/_/test-results/chromatic-archives
Please make sure that you have run your E2E tests, or have set the CHROMATIC_ARCHIVE_LOCATION env var if the output directory for the tests is not in the standard location.
```

Looking at this change, there's this comment:

```javascript
// If we tried to run the E2E package's bin directly (due to being in the action)
// and it failed, that means we couldn't find it. This probably means they haven't
// installed the right dependency or run from the right directory
```

I think that's probably the wrong logic because `build-archive-storybook` could fail to execute for many different reasons. I think the easiest fix would be to dump out the underlying error (so that there's a clue for someone encountering a different error) or I think the ideal fix would be to limit the `missingDependency` error to only cases where the dependency was missing.

tmeasday:
Thanks for the investigation @shrink and that's a great point!

PR #920, opened 2024-02-12
Fix Storybook published messaging

tevanoff:
This fixes a couple of issues with the Storybook published messaging.

1. Logging this from the `publishBuild` step meant that sometimes the `ctx.build` used to construct the message was in an early stage (like `PASSED`) without component and story information, resulting in a message like:
```
✔ Storybook published
We found undefined components with undefined stories.
```

2. This is also logged from the `verifyBuild` step under certain conditions (like with `exit-once-uploaded`), resulting in duplicate messaging like:
```
✔ Storybook published
We found undefined components with undefined stories.

✔ Storybook published
We found 19 components with 221 stories.
```

### How to Test
* With the canary version, commit some changes and run builds with various flags including with and without `exit-once-uploaded` and `only-changed`.

<details>
  <summary>📦 Published PR as canary version: <code>10.9.5--canary.920.7904534893.0</code></summary>
  <br />

  ```bash
  npm install chromatic@10.9.5--canary.920.7904534893.0
  # or 
  yarn add chromatic@10.9.5--canary.920.7904534893.0
  ```
</details>

### DIFF
--- a/node-src/tasks/verify.ts
+++ b/node-src/tasks/verify.ts
@@ -75,8 +75,6 @@ export const publishBuild = async (ctx: Context) => {
     setExitCode(ctx, exitCodes.BUILD_FAILED, false);
     throw new Error(publishFailed().output);
   }
-
-  ctx.log.info(storybookPublished(ctx));
 };
 
 const StartedBuildQuery = `
@@ -244,12 +242,13 @@ export const verifyBuild = async (ctx: Context, task: Task) => {
     }
   }
 
+  ctx.log.info(storybookPublished(ctx));
+
   transitionTo(success, true)(ctx, task);
 
   if (list || ctx.isPublishOnly || matchesBranch(ctx.options.exitOnceUploaded)) {
     setExitCode(ctx, exitCodes.OK);
     ctx.skipSnapshots = true;
-    ctx.log.info(storybookPublished(ctx));
   }
 };
 
--- a/node-src/ui/messages/info/storybookPublished.ts
+++ b/node-src/ui/messages/info/storybookPublished.ts
@@ -7,11 +7,14 @@ import link from '../../components/link';
 import { stats } from '../../tasks/snapshot';
 
 export default ({ build, storybookUrl }: Pick<Context, 'build' | 'storybookUrl'>) => {
-  if (build) {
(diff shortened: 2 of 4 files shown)


node-src/tasks/verify.ts
```diff
@@ -244,12 +242,13 @@ export const verifyBuild = async (ctx: Context, task: Task) => {
     }
   }
 
+  ctx.log.info(storybookPublished(ctx));
```
tevanoff, line 245: @ghengeveld it looks like the addition of this on line 79 above was meant to print this all the time, so I moved it out of the conditional block down here since the build on the context at this point is in a correct state for logging component and story counts. But please let me know if I'm overlooking other reasons for it being added above.

ethriel3695:
@tevanoff 
How can I test this?

Can I reference the Canary build and then what?
How can I trigger this use case?

tevanoff:
> @tevanoff How can I test this?
> 
> Can I reference the Canary build and then what? How can I trigger this use case?

@ethriel3695 I added some info about testing to the PR message. Those two flags seemed to affect the behavior of the output in my testing.

node-src/ui/messages/info/storybookPublished.ts
```diff
@@ -7,7 +7,10 @@ import link from '../../components/link';
 import { stats } from '../../tasks/snapshot';
 
 export default ({ build, storybookUrl }: Pick<Context, 'build' | 'storybookUrl'>) => {
-  if (build) {
+  // `ctx.build` is initialized and overwritten in many ways, which means that
+  // this can be any kind of build without component and stories information,
+  // like PASSED builds, for example
```
ghengeveld, line 12: I don't see why `PASSED` builds would not have component and story information. They were tested after all.

node-src/tasks/verify.ts
ghengeveld: Yeah I think this is fine.

ghengeveld (approved):

ghengeveld:
:rocket: PR was released in [`v10.9.5`](https://github.com/chromaui/chromatic-cli/releases/tag/v10.9.5) :rocket:

PR #921, opened 2024-02-13
Add new invalid sb base dir error message and validator function

andrewortwein:
We will now check if the `storybookBaseDir` is valid before letting the build continue. If it is not defined but needs to be, or is defined but does not point to the correct location, the build will fail and an error will be displayed.

## QA
1. Use an app that has TurboSnap enabled
2. Commit a visual change
3. Using the canary build, run a build with `--only-changed` and an invalid `--storybook-base-dir`
   - e.g. `npx chromatic@<CANARY_VERSION> --only-changed --storybook-base-dir "src/foobar"`
4. Verify that the build fails with an error message indicating that the Storybook base dir is invalid
   - Prior to this change, the build would complete successfully, incorrectly detecting no visual changes

<details>
  <summary>📦 Published PR as canary version: <code>11.0.0--canary.921.8007040697.0</code></summary>
  <br />

  ```bash
  npm install chromatic@11.0.0--canary.921.8007040697.0
  # or 
  yarn add chromatic@11.0.0--canary.921.8007040697.0
  ```
</details>


node-src/tasks/upload.ts
```diff
@@ -142,6 +143,9 @@ export const traceChangedFiles = async (ctx: Context, task: Task) => {
     }
 
     const stats = await readStatsFile(statsPath);
+
+    checkStorybookBaseDir(ctx.options.storybookBaseDir, stats);
```
andrewortwein, line 147: I couldn't decide between doing the checking AND the error throwing in this function, or checking in this function and having it return a falsy value and then throwing the error here in upload. Which one is a more consistent pattern in this repo?

node-src/lib/checkStorybookBaseDir.test.ts
```diff
@@ -0,0 +1,82 @@
...
+import path from 'path';
+
+describe('checkStorybookBaseDir', () => {
+  it('should return if a js module in stats exists at the path prepended by the storybookBaseDir', () => {
+    const storybookBaseDir = path.join(__dirname, '../../');
+    const stats = {
+      modules: [
+        { name: './node_modules/@storybook/core-client/dist/esm/globals/polyfills.js' },
+        { name: './README.md' },
+        { name: './test-stories/A.js' },
```
andrewortwein, line 12: In the rest of the tests, I look for files in __mocks__. But this folder proved useful and _seems_ to be used in a similar way in other tests. Would it be better just to add a few extra files to `test-stories` and `subdir`?

andrewortwein:
@ghengeveld some unrelated tests are now breaking because of the check I put in `upload.ts`. Am I making assumptions about `stats.modules` that might not always be true?

node-src/ui/messages/errors/invalidStorybookBaseDir.ts
```diff
@@ -0,0 +1,10 @@
+import chalk from 'chalk';
+import { dedent } from 'ts-dedent';
+
+import { error } from '../../components/icons';
+
+export default () =>
```
ghengeveld, line 6: Please use a named export.

node-src/ui/messages/errors/missingStorybookBaseDir.ts
```diff
@@ -0,0 +1,10 @@
+import chalk from 'chalk';
+import { dedent } from 'ts-dedent';
+
+import { error } from '../../components/icons';
+
+export default () =>
```
ghengeveld, line 6: Please use a named export.

node-src/lib/checkStorybookBaseDir.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  // Check if any of the source module files exist in the storybookBaseDir
+  for (const file of sourceModuleFiles) {
+    const absolutePath = path.join(storybookBaseDir || '', file.name);
+    if (fs.existsSync(absolutePath)) {
+      return true;
+    }
+  }
+
+  throw new Error(storybookBaseDir ? invalidStorybookBaseDir() : missingStorybookBaseDir());
```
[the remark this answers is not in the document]
ghengeveld: I would've been okay with just `invalidStorybookBaseDir` even if it's missing. You could pass `storybookBaseDir` in as a parameter if you want to distinguish them.

ghengeveld (changes requested):
Nice. Needs a little work to safeguard performance but otherwise looks good.

node-src/lib/checkStorybookBaseDir.test.ts
ghengeveld: I don't feel strongly about either option. Mocks are meant to be used in tests so if you reuse them for that purpose I think it's fine.

node-src/tasks/upload.ts
ghengeveld: Yes, `INVALID_OPTIONS` (254) would work here.

node-src/lib/checkStorybookBaseDir.test.ts
```diff
@@ -0,0 +1,83 @@
...
+          name: './subdir/test.tsx',
+        },
+      ],
+    };
+
+    [statsWithJsModule, statsWithJsxModule, statsWithTsModule, statsWithTsxModule].forEach(
+      async (stats) => {
+        await expect(checkStorybookBaseDir(ctx, stats)).resolves.toBeUndefined();
+      }
+    );
```
ghengeveld, line 51: I think this needs to be a `Promise.all` with `.map`, a for loop or just four manual invocations (my preference, for simplicity). Otherwise this test isn't going to await the assertion.

ghengeveld (approved):

node-src/lib/checkStorybookBaseDir.ts
```diff
@@ -0,0 +1,44 @@
...
+        return limitConcurrency(() => {
+          const absolutePath = path.join(storybookBaseDir || '', file.name);
+          return new Promise((resolve, reject) =>
+            fs.access(absolutePath, (err) => {
+              if (err) {
+                reject();
+              } else {
+                ctx.log.debug('Found:', absolutePath);
+                resolve(true);
+              }
```
thafryer, line 33: Can we log the path we **did not** find instead of the ones that we did? It will make it easier to debug and keep the logs cleaner? This could end up being a lot of files.

node-src/lib/checkStorybookBaseDir.ts
```diff
@@ -0,0 +1,44 @@
...
+            })
+          );
+        });
+      })
+    );
+  } catch (err) {
+    ctx.log.debug('No modules found in:', storybookBaseDir);
```
thafryer, line 40: This message may need to be adjusted to say that Module from state file was not found.

thafryer (approved):
LGTM

node-src/lib/checkStorybookBaseDir.ts
andrewortwein: Hmm, you actually have me thinking that either scenario could end up being a lot of files and, thus, a lot of debug prints in the logs. Even if the very first file check succeeds and the promise resolves, the other file checks will still occur. Logging every one of these could be very noisy. On the other hand, the same number of checks could happen in the scenario where none of them succeed. So that could also be noisy.

If we succeed on a check, we really just need that one path logged. Maybe we could bubble that back up to the promise and log it there? If we fail on all checks, would we want a log for every path checked or just one as an example?

andrewortwein: So, to answer your question, yes we could log files that were **not** found instead of those that were. But that could still be very noisy. I suppose that would be better since we are only potentially logging a lot of files in the unhappy path of an invalid `storybookBaseDir`. That seems like a good improvement.

ghengeveld:
:rocket: PR was released in [`v11.0.0`](https://github.com/chromaui/chromatic-cli/releases/tag/v11.0.0) :rocket:

Issue #933, opened 2024-03-01
Allow specifying --storybook-build-dir when buildScriptName is configured in config file


woutervanvliet:
**Feature request**

I'd like to include buildScriptName in my config file, so we don't have to worry about passing that as a commandline argument. However, we sometimes use a prebuilt storybook. This triggers an error as those options are mutually exclusive.

**Proposed solution**

I'd like to be able to specify both options

**Alternative solutions**

It looks like passing `--build-script-name -` is possible to remove that option from the configuration, but this is cumbersome and not very well documented.

**Additional context**

I don't understand why they can't both be specified. buildScriptName will just never be used when using a prebuilt storybook but that doesn't really matter.

woutervanvliet:
PR merged, so closing this issue

PR #934, opened 2024-03-01
Allow passing both `buildScriptName` and `storybookBuildDir`

woutervanvliet:
Allows passing both --storybook-build-dir and  --build-script-name

Solves #933

### DIFF
--- a/node-src/lib/getOptions.test.ts
+++ b/node-src/lib/getOptions.test.ts
@@ -115,10 +115,10 @@ describe('getOptions', () => {
     });
   });
 
-  it('throws if you try to pass a build script and a directory', async () => {
+  it('allows you to pass both a build script and a directory', async () => {
     await expect(() =>
-      getOptions(getContext(['-b', '/tmp/dir', '--storybook-build-dir', '/tmp/dir']))
-    ).toThrow(/You can only use one of --build-script-name, --storybook-build-dir/);
+      getOptions(getContext(['-b', 'build:storybook', '--storybook-build-dir', '/tmp/dir']))
+    ).not.toThrow()
   });
 
   it('allows you to specify the branch name', async () => {
--- a/node-src/lib/getOptions.test.ts
+++ b/node-src/lib/getOptions.test.ts
@@ -115,27 +115,10 @@ describe('getOptions', () => {
     });
   });
 
-  it('allows you to pass both a build script and a directory', () => {
-    const context = getContext(['-b', 'build:storybook', '--storybook-build-dir', '/tmp/dir'])
-    const log = context.log as unknown as TestLogger
+  it('allows you to pass both a build script and a directory', async () => {
     expect(() =>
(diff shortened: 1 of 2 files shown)


ghengeveld:
Hi Wouter, thanks for working on a fix for this!

I'm okay with your proposed change to allow `--storybook-build-dir` whilst `buildScriptName` is set in your chromatic.config.json. However, the original reason we had this restriction was to avoid confusion over which one gets priority. I propose the following solution:

- Rather than checking `singularOpts` against the final (combined) `options` object, check it against `configuration` (JSON), `optionsFromFlags` (flags / GitHub Action args) and `extraOptions` (Node API args) individually. This would allow overriding configuration using a flag or API arg, while preventing any misconfiguration or bad combination of flags/args.
- Print a warning when this happens.

Are you comfortable / willing to make this change?

woutervanvliet:
Thanks for the input - I've reluctantly updated my PR, because I'm not sure I entirely agree. So here are a few followup questions:

 - Am I right to assume that the presence of the `buildScriptName` configuration, regardless of where it comes from, doesn't actually have any effect - until the chromatic cli decides that a build is necessary because no build dir was specified? 
 - Am I also right to assume that build script name, when not specified, defaults to "build-storybook", and as such always can be considered to be specified?

If both answers are yes, than I don't think assumption that combining buildScriptName and storybookBuildDir in configuration and/or command line arguments in any way should indicate a potential mistake, and I think my latest commit should be reverted.

And if my assumptions are not entirely correct, please check if my change reflects your intention.

ghengeveld:
> * Am I right to assume that the presence of the `buildScriptName` configuration, regardless of where it comes from, doesn't actually have any effect - until the chromatic cli decides that a build is necessary because no build dir was specified?

Almost. The `buildScriptName` is also used to infer `configDir` and `staticDir` options from its flags. This is relevant for proper functioning of TurboSnap, even if `storybookBuildDir` is also specified. In that case it won't build Storybook but it'll assume the prebuilt Storybook used the same `buildScriptName` flags.

> * Am I also right to assume that build script name, when not specified, defaults to "build-storybook", and as such always can be considered to be specified?

Yes, although we'll still check if that script actually exists in your package.json.

> If both answers are yes, than I don't think assumption that combining buildScriptName and storybookBuildDir in configuration and/or command line arguments in any way should indicate a potential mistake, and I think my latest commit should be reverted.

After giving this some more thought, I think that's valid. There _is_ actually a valid reason to have both a `buildScriptName` and a `storybookBuildDir` (for configDir/staticDir inference), so combining these may not necessarily be a mistake.

ghengeveld (approved):

ghengeveld:
:rocket: PR was released in [`v11.2.0`](https://github.com/chromaui/chromatic-cli/releases/tag/v11.2.0) :rocket:

woutervanvliet:
Thanks

PR #940, opened 2024-03-06
Expose E2E build errors

tevanoff:
Currently, any error thrown during the build step of an E2E build is treated as a missing dependency error, and the error message output is very specific to that. However, the build can fail for other reasons, such as archive directory not found. These other legit errors are swallowed and the resulting output is confusing.

This flips things around to try to determine if the error is due to a missing dependency, where the error usually has something to do about a command not being found. If the error message doesn't match, this will output a more generic error message that includes the original error message. 

**How to test**
* With a local project, install the canary CLI and try to run the chromatic command in both of the following scenarios and see that the errors are correct:
  * With the E2E dep installed, but with the archive directory removed
  * With the archive directory in place, but without the E2E dep installed
* Using the canary github action (`chromaui/action-canary@v1`), test a project with the above two scenarios and see that the errors are correct. 

<details>
  <summary>📦 Published PR as canary version: <code>11.0.5--canary.940.8192597877.0</code></summary>
  <br />

  ```bash
  npm install chromatic@11.0.5--canary.940.8192597877.0
  # or 
  yarn add chromatic@11.0.5--canary.940.8192597877.0
  ```
</details>

### DIFF
--- a/node-src/lib/setExitCode.ts
+++ b/node-src/lib/setExitCode.ts
@@ -25,6 +25,9 @@ export const exitCodes = {
   STORYBOOK_START_FAILED: 22,
   STORYBOOK_BROKEN: 23,
 
+  // E2E errors
+  E2E_BUILD_FAILED: 51,
+
   // Subprocess errors
   GIT_NOT_CLEAN: 101,
   GIT_OUT_OF_DATE: 102,
--- a/node-src/tasks/build.test.ts
+++ b/node-src/tasks/build.test.ts
@@ -168,7 +168,18 @@ describe('buildStorybook', () => {
 });
 
 describe('buildStorybook E2E', () => {
-  it('fails with missing dependency error when dependency not installed', async () => {
+  // Error messages that we expect to result in the missing dependency error
+  const missingDependencyErrorMessages = [
+    { name: 'not found 1', error: 'Command not found: build-archive-storybook' },
+    { name: 'not found 2', error: 'Command "build-archive-storybook" not found' },
+    { name: 'npm not found', error: 'NPM error code E404\n\nMore error info' },
+    { name: 'exit code not found', error: 'Command failed with exit code 127: some command\n\nsome error line\n\n' },
+    { name: 'single line command failure', error: 'Command failed with exit code 1: npm exec build-archive-storybook --output-dir /tmp/chromatic--4210-0cyodqfYZabe' },
+  ];
+
+  it.each(
+    missingDependencyErrorMessages
+  )('fails with missing dependency error when error message is $name', async ({ error }) => {
     const ctx = {
       buildCommand: 'npm exec build-archive-storybook',
       options: { buildScriptName: '', playwright: true },
@@ -176,20 +187,11 @@ describe('buildStorybook E2E', () => {
       log: { debug: vi.fn(), error: vi.fn() },
     } as any;
 
-    const missingDependencyErrorMessages = [
-      'Command not found: build-archive-storybook',
-      'Command "build-archive-storybook" not found',
-      'NPM error code E404\n\nMore error info',
-      'Command failed with exit code 1: npm exec build-archive-storybook --output-dir /tmp/chromatic--4210-0cyodqfYZabe'
-    ];
-
-    for(const errorMessage of missingDependencyErrorMessages) {
-      command.mockRejectedValueOnce(new Error(errorMessage));
-      await expect(buildStorybook(ctx)).rejects.toThrow('Command failed');
-      expect(ctx.log.error).toHaveBeenCalledWith(expect.stringContaining('Failed to import `@chromatic-com/playwright`'));
-      
-      ctx.log.error.mockClear();
-    };
+    command.mockRejectedValueOnce(new Error(error));
+    await expect(buildStorybook(ctx)).rejects.toThrow('Command failed');
+    expect(ctx.log.error).toHaveBeenCalledWith(expect.stringContaining('Failed to import `@chromatic-com/playwright`'));
+    
+    ctx.log.error.mockClear();
   });
 
   it('fails with generic error message when not missing dependency error', async () => {
--- a/node-src/tasks/build.test.ts
+++ b/node-src/tasks/build.test.ts
@@ -200,7 +200,7 @@ describe('buildStorybook E2E', () => {
       log: { debug: vi.fn(), error: vi.fn() },
     } as any;
 
-    const errorMessage = 'Cannot find archive directory\n\nOther info';
+    const errorMessage = 'Command failed with exit code 1: npm exec build-archive-storybook --output-dir /tmp/chromatic--4210-0cyodqfYZabe\n\nMore error message lines\n\nAnd more';
     command.mockRejectedValueOnce(new Error(errorMessage));
     await expect(buildStorybook(ctx)).rejects.toThrow('Command failed');
     expect(ctx.log.error).not.toHaveBeenCalledWith(expect.stringContaining('Failed to import `@chromatic-com/playwright`'));
--- a/node-src/tasks/build.test.ts
+++ b/node-src/tasks/build.test.ts
@@ -166,3 +166,45 @@ describe('buildStorybook', () => {
     );
   });
 });
+
+describe('buildStorybook E2E', () => {
+  it('fails with missing dependency error when dependency not installed', async () => {
+    const ctx = {
+      buildCommand: 'npm exec build-archive-storybook',
+      options: { buildScriptName: '', playwright: true },
+      env: { STORYBOOK_BUILD_TIMEOUT: 0 },
+      log: { debug: vi.fn(), error: vi.fn() },
+    } as any;
+
+    const missingDependencyErrorMessages = [
+      'Command not found: build-archive-storybook',
+      'Command "build-archive-storybook" not found',
+      'NPM error code E404\n\nMore error info',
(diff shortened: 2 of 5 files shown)


node-src/tasks/build.ts
```diff
@@ -77,7 +77,7 @@ export const buildStorybook = async (ctx: Context) => {
     ctx.log.debug('Runtime metadata:', JSON.stringify(ctx.runtimeMetadata, null, 2));
 
     const subprocess = execaCommand(ctx.buildCommand, {
-      stdio: [null, logFile, logFile],
+      stdio: [null, logFile, null],
```
tevanoff, line 80: This pipes stderr to the terminal rather than the storybook build log file, which seems better in general?

node-src/ui/messages/errors/e2eBuildFailed.ts
```diff
@@ -0,0 +1,17 @@
...
+  flag,
+  errorMessage,
+}: {
+  flag: string;
+  errorMessage: string;
+}) => {
+  return dedent(chalk`
+      ${error} Failed to run \`chromatic --${flag}\`:
+      
+      ${errorMessage}
```
tevanoff, line 15: We could add a list of causes for this error, or maybe better yet link to a troubleshooting page in our docs.

node-src/tasks/build.ts
```diff
@@ -87,15 +87,12 @@ export const buildStorybook = async (ctx: Context) => {
     // and it failed, that means we couldn't find it. This probably means they haven't
     // installed the right dependency or run from the right directory
     if (
-      ctx.options.inAction &&
-      (ctx.options.playwright || ctx.options.cypress) &&
-      e.message.match(e2EbuildBinName)
+      e.message.match(e2EbuildBinName) &&
```
tevanoff, line 90: Not sure if we even need this check on the bin name. The command that failed is in the error message, so I don't know when this wouldn't match for E2E projects.

node-src/tasks/build.ts
tmeasday: I'm not sure what SB outputs to stderr vs stdout?

node-src/tasks/build.ts
tmeasday: Right, I guess it's probably rendundant.

node-src/ui/messages/errors/e2eBuildFailed.ts
tmeasday: I guess the main one we are worried about is when an error like `no such binary exists "build-archive-storybook"`. For a user that's going to be a pretty opaque error if they have no idea about what archive SB even is.

node-src/tasks/build.ts
```diff
@@ -60,6 +61,34 @@ export const setBuildCommand = async (ctx: Context) => {
...
+  return errorRegexes.some((regex) => msg.match(new RegExp(regex, 'gi')));
+}
+
+function e2eBuildErrorMessage(err, workingDir, ctx: Context): { exitCode: number, message: string } {
+  const flag = ctx.options.playwright ? 'playwright' : 'cypress';
+
+  // If we tried to run the E2E package's bin directly (due to being in the action)
+  // and it failed, that means we couldn't find it. This probably means they haven't
+  // installed the right dependency or run from the right directory.
+  if (isE2EBuildCommandNotFoundError(err)) {
```
tevanoff, line 78: @tmeasday as per our slack thread, I fixed this to try to determine if it's the command not found error.

node-src/tasks/build.ts
```diff
@@ -60,6 +61,34 @@ export const setBuildCommand = async (ctx: Context) => {
 const timeoutAfter = (ms) =>
   new Promise((resolve, reject) => setTimeout(reject, ms, new Error(`Operation timed out`)));
 
+function isE2EBuildCommandNotFoundError(errorMessage: string) {
+  // It's hard to know if this is the case as each package manager has a different type of
+  // error for this, but we'll try to figure it out.
+  const errorRegexes = ['command not found', `[\\W]?${e2eBuildBinName}[\\W]? not found`, 'code E404', 'exit code 127', `command failed.*${e2eBuildBinName}.*$`];
+  return errorRegexes.some((regex) => errorMessage.match(new RegExp(regex, 'gi')));
```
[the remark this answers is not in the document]
tmeasday: This is decent advice (be careful about dynamic regexps) from Codacy but the solution doesn't make sense to me -- I could be off though. cc @paulelliott

node-src/tasks/build.ts
```diff
@@ -60,6 +61,34 @@
...
```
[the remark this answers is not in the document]
tmeasday: @paulelliott my feeling is the AI is getting really wordy here and I am glazing over and missing the key points:

1. Be careful about dynamic strings in regexps
2. You can comment away the warning
3. However, be super careful if you are going to do that.

node-src/tasks/build.test.ts
```diff
@@ -166,3 +166,45 @@ describe('buildStorybook', () => {
     );
   });
 });
+
+describe('buildStorybook E2E', () => {
+  it('fails with missing dependency error when dependency not installed', async () => {
```
tmeasday, line 171: you could use `test.each()` here?

tmeasday (approved):
I like it, thanks for working through it @tevanoff

node-src/tasks/build.ts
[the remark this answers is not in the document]
paulelliott: The other piece of advice it had was to use the RE2 library from google to parse user-supplied regex in a safe way. That feels like an easy change to make and right way to resolve these issues. The initial comment it put on the PR yesterday said that and I'm not sure why it shifted to this. You can see that advice on the second warning on the Codacy app.

https://app.codacy.com/gh/chromaui/chromatic-cli/pullRequest?prid=13786760

node-src/tasks/build.ts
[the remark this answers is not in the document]
paulelliott: When I enabled this I just blindly turned the whole thing on. It is quite configurable in the application so if there are rules we disagree with we can easily turn them off per repo. You can also dismiss this specific warning in their interface and it keeps a log of all the specific places it was ignored.

node-src/tasks/build.ts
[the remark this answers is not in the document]
tevanoff: This is good advice, however, the `e2eBuildBinName` that it is calling out is a simple string constant, not user-controlled input, so this is all moot.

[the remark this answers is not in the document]
tevanoff: @tmeasday I can't make sense of that suggestion to use `.replace()` either 🤔

node-src/lib/setExitCode.ts
```diff
@@ -25,6 +25,9 @@ export const exitCodes = {
   STORYBOOK_START_FAILED: 22,
   STORYBOOK_BROKEN: 23,
 
+  // E2E errors
+  E2E_BUILD_FAILED: 51,
+
```
skitterm, line 30: Should we add this to the CLI [exit codes](https://www.chromatic.com/docs/cli/#exit-codes) in the documentation?

node-src/tasks/build.ts
```diff
@@ -60,6 +61,40 @@ export const setBuildCommand = async (ctx: Context) => {
...
 
+function isE2EBuildCommandNotFoundError(errorMessage: string) {
+  // It's hard to know if this is the case as each package manager has a different type of
+  // error for this, but we'll try to figure it out.
+  const errorRegexes = [
+    'command not found', // `Command not found: build-archive-storybook`
+    `[\\W]?${e2eBuildBinName}[\\W]? not found`, // `Command "build-archive-storybook" not found`
+    'code E404', // npm not found error can include this code
+    'exit code 127', // Exit code 127 is a generic not found exit code
+    `command failed.*${e2eBuildBinName}.*$`]; // A single line error from execa like `Command failed: yarn build-archive-storybook ...`
```
skitterm, line 72: Thanks for doc-ing the edge-cases and where each comes from.

node-src/tasks/build.test.ts
```diff
@@ -166,3 +166,47 @@ describe('buildStorybook', () => {
...
+
+  it('fails with generic error message when not missing dependency error', async () => {
+    const ctx = {
+      buildCommand: 'npm exec build-archive-storybook',
+      options: { buildScriptName: '', playwright: true },
+      env: { STORYBOOK_BUILD_TIMEOUT: 0 },
+      log: { debug: vi.fn(), error: vi.fn() },
+    } as any;
+
+    const errorMessage = 'Command failed with exit code 1: npm exec build-archive-storybook --output-dir /tmp/chromatic--4210-0cyodqfYZabe\n\nMore error message lines\n\nAnd more';
```
skitterm, line 205: @tevanoff this generic error message includes (for example) that the archive directory isn't found? Just wanted to make sure we're still passing that part through.

tevanoff: @skitterm yes, this is testing that the generic error message (`Command failed with exit code 1: npm exec build-archive-storybook...`) that has multiple lines gets output using our the generic E2E build error template, which will out put the full error. That's how things like the archive directory not found errors are shaped.

node-src/lib/setExitCode.ts
tevanoff: @skitterm sure, I can open a PR for that.

skitterm (approved):

ghengeveld:
:rocket: PR was released in [`v11.0.6`](https://github.com/chromaui/chromatic-cli/releases/tag/v11.0.6) :rocket:

PR #943, opened 2024-03-08
Fix e2e peer dep versions

tevanoff:
The `^0.5.2` version specifier will only match `0.5.x` versions, but we're at `0.6.x`.

<details>
  <summary>📦 Published PR as canary version: <code>11.0.8--canary.943.8236202501.0</code></summary>
  <br />

  ```bash
  npm install chromatic@11.0.8--canary.943.8236202501.0
  # or 
  yarn add chromatic@11.0.8--canary.943.8236202501.0
  ```
</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -206,8 +206,8 @@
     "zod": "^3.22.2"
   },
   "peerDependencies": {
-    "@chromatic-com/cypress": "^0.5.2 || ^1.0.0",
-    "@chromatic-com/playwright": "^0.5.2 || ^1.0.0"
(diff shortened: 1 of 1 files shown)


thafryer (approved):

tevanoff:
> How common do we think this is going to be? A part of me thinks we can benefit from having CLI and E2E live in one singular monorepo.

@thafryer it shouldn't be common, it's just that semver matching is weird with `0.x` versions and so what we had specified only matched `0.5.x`. It should now match all `0.x.x` versions.

PR #962, opened 2024-03-28
Suppress issues caused by missing Git remote

ghengeveld:
A missing slug (caused by missing Git remote URL) is not a serious problem, so there is no need to log a warning for it. A debug message is more appropriate.

The patch build mechanism attempts to check for remote updates, which when failed should not prevent the patch build from running. In this case a warning _is_ appropriate because it might lead to unexpected results.

<details>
  <summary>📦 Published PR as canary version: <code>11.2.1--canary.962.8479861319.0</code></summary>
  <br />

  ```bash
  npm install chromatic@11.2.1--canary.962.8479861319.0
  # or 
  yarn add chromatic@11.2.1--canary.962.8479861319.0
  ```
</details>

### DIFF
--- a/node-src/git/git.ts
+++ b/node-src/git/git.ts
@@ -148,10 +148,15 @@ export async function getChangedFiles(baseCommit: string, headCommit = '') {
 
 /**
  * Returns a boolean indicating whether the workspace is up-to-date (neither ahead nor behind) with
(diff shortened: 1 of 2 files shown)


thafryer (approved):
LGTM

PR #988, opened 2024-05-17
Use replacement build when encountering baseline build with uncommitted changes

ghengeveld:
Addresses https://github.com/chromaui/addon-visual-tests/issues/231

Similar to how we use a replacement build as baseline when an ancestor build's commit cannot be found, this will use a replacement build as baseline when the original baseline build had uncommitted changes.

### DIFF
--- a/node-src/git/getBaselineBuilds.ts
+++ b/node-src/git/getBaselineBuilds.ts
@@ -16,6 +16,7 @@ const BaselineCommitsQuery = gql`
         status(legacy: false)
         commit
         committedAt
+        uncommittedHash
         changeCount
       }
     }
@@ -29,6 +30,7 @@ interface BaselineCommitsQueryResult {
       status: string;
       commit: string;
       committedAt: number;
+      uncommittedHash: string;
       changeCount: number;
     }[];
   };
--- a/node-src/git/getChangedFilesWithReplacement.ts
+++ b/node-src/git/getChangedFilesWithReplacement.ts
@@ -2,10 +2,11 @@ import { getChangedFiles } from './git';
 import { findAncestorBuildWithCommit } from './findAncestorBuildWithCommit';
 import { Context } from '../types';
 
-type BuildWithCommit = {
+type BuildWithCommitInfo = {
   id: string;
   number: number;
   commit: string;
+  uncommittedHash: string;
 };
 
 /**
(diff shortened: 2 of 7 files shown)


node-src/git/getBaselineBuilds.ts
```diff
@@ -16,6 +16,7 @@ const BaselineCommitsQuery = gql`
         status(legacy: false)
         commit
         committedAt
+        uncommittedHash
```
tmeasday, line 19: This will require auth changes on the index I expect?

node-src/git/getChangedFilesWithReplacement.ts
```diff
@@ -18,17 +19,18 @@ type BuildWithCommit = {
...
+    if (build.uncommittedHash) throw new Error('Build had uncommitted changes');
     const changedFiles = await getChangedFiles(build.commit);
     return { changedFiles };
   } catch (err) {
     context.log.debug(
       `Got error fetching commit for #${build.number}(${build.commit}): ${err.message}`
     );
 
-    if (err.message.match(/bad object/)) {
+    if (err.message.match(/(bad object|uncommitted changes)/)) {
```
tmeasday, line 33: Are we OK with throwing for flow control? 

I think maybe it would be less confusing to do:

```ts
let commitErr;
try {
  if (!build.uncommittedHash) {
     const changedFiles = await getChangedFiles(build.commit);
     return { changedFiles };
  }
} catch (err) {
   if (!err.message.match(/bad object/)) {
     throw err;
   }
   commitErr = err;
}

// now do replacement logic

// if fails
throw commitErr;
```

I'm not super fussed about it though.

tmeasday (commented):
LGTM, a question and a nit.

tmeasday (approved):

node-src/git/getBaselineBuilds.ts
tmeasday: Oh I see you already did that, I was looking at out of date code. Carry on.

node-src/git/getChangedFilesWithReplacement.ts
ghengeveld: I would've done that if not for the preexisting catch logic. It seemed more sensible to extend that rather than pile on a 2nd kind of "exception" handling.

ghengeveld:
:rocket: PR was released in [`v11.4.0`](https://github.com/chromaui/chromatic-cli/releases/tag/v11.4.0) :rocket:

PR #1032, opened 2024-09-09
Support storybook-rsbuild

### LINKED ISSUE #1017: Chromatic TurboSnap builds don't work with Rsbuild
_[I've created a codesandbox to demonstrate this issue](https://codesandbox.io/p/devbox/rsbuild-storybook-iframe-html-compile-error-forked-l5nd2t), using a throwaway Chromatic project with TurboSnap enabled._

When Storybook is built using [Rsbuild](https://rsbuild.dev/)/Rspack (via the [storybook-rsbuild](https://github.com/rspack-contrib/storybook-rsbuild) package), Chromatic fails to detect any CSF globs in the preview-stats.json file, and so TurboSnap fails. Full builds work fine.

E.g. running `npx chromatic --only-changed` would give an error like this

```
  Error: ✖ Failed to publish your built Storybook
  Could not retrieve dependent story files.
  Did not find any CSF globs in preview-stats.json
```

I'm not super familiar with Webpack or the stats files in general, but this seems to stem from the "reason" files having module names from [this list](https://github.com/chromaui/chromatic-cli/blob/c91d3010049ed7803cc369d61c6c0161530f4f0c/node-src/lib/getDependentStoryFiles.ts#L100-L111) when doing a Webpack or Vite build. Rsbuild, instead, has very different names.

For example, a Vite module name would look like this:

```json
{
  "moduleName": "/virtual:/@storybook/builder-vite/vite-app.js"
}
```

and an Rsbuild one like this:

```json
{
  "moduleName": "/Users/phil/work/ts-wholesale/ui/src|lazy|/^\\.\\/.*$/|include: /(?!.*node_modules)(?:\\/src(?:\\/(?!\\.)(?:(?:(?!(?:^|\\/)\\.).)*?)\\/|\\/|$)(?!\\.)(?=.)[^/]*?\\.stories\\.(js|jsx|mjs|ts|tsx))$/|chunkName: [request]|groupOptions: {}|strict namespace object"
}
```

Because the second module name (in modules[0].reasons[0].moduleName) is not one of the "approved" items, the CSF story module is skipped.

I _think_ that's it. Hopefully you can advise how we can use Rsbuild & Chromatic together!

joshwooding:
Closes #1017

### DIFF
--- a/node-src/lib/getDependentStoryFiles.ts
+++ b/node-src/lib/getDependentStoryFiles.ts
@@ -82,7 +82,7 @@ export async function getDependentStoryFiles(
   // Convert a "webpack path" (relative to storybookBaseDir) to a "git path" (relative to repository root)
   // e.g. `./src/file.js` => `path/to/storybook/src/file.js`
   const normalize = (posixPath: FilePath): NormalizedName => {
-    const CSF_REGEX = /\s+sync\s+/g;
+    const CSF_REGEX = /\s+(sync|lazy)\s+/g;
     const URL_PARAM_REGEX = /(\?.*)/g;
     const newPath = normalizePath(posixPath, rootPath, baseDir);
     // Trim query params such as `?ngResource` which are sometimes present
@@ -110,6 +110,8 @@ export async function getDependentStoryFiles(
     `./storybook-stories.js`,
     // vite builder
     `/virtual:/@storybook/builder-vite/vite-app.js`,
+    // rspack builder
+    `./node_modules/.cache/storybook/default/dev-server/storybook-stories.js`
   ].map(normalize);
 
   const modulesByName = new Map<NormalizedName, Module>();
--- a/node-src/lib/getDependentStoryFiles.ts
+++ b/node-src/lib/getDependentStoryFiles.ts
@@ -111,7 +111,7 @@ export async function getDependentStoryFiles(
     // vite builder
     `/virtual:/@storybook/builder-vite/vite-app.js`,
     // rspack builder
-    `./node_modules/.cache/storybook/default/dev-server/storybook-stories.js`
+    `./node_modules/.cache/storybook/default/dev-server/storybook-stories.js`,
   ].map(normalize);
 
   const modulesByName = new Map<NormalizedName, Module>();
--- a/node-src/lib/getDependentStoryFiles.test.ts
+++ b/node-src/lib/getDependentStoryFiles.test.ts
@@ -162,6 +162,35 @@ describe('getDependentStoryFiles', () => {
     });
   });
 
+  it('detects direct changes to CSF files, rspack', async () => {
+    const changedFiles = ['src/foo.stories.js'];
+    const modules = [
+      {
+        "name": "./src/foo.stories.js",
+        "id": "./src/foo.stories.js",
+        "reasons": [
+          {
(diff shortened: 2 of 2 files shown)


node-src/lib/getDependentStoryFiles.ts
```diff
@@ -80,7 +80,7 @@ export async function getDependentStoryFiles(
   // Convert a "webpack path" (relative to storybookBaseDir) to a "git path" (relative to repository root)
   // e.g. `./src/file.js` => `path/to/storybook/src/file.js`
   const normalize = (posixPath: FilePath): NormalizedName => {
-    const CSF_REGEX = /\s+sync\s+/g;
+    const CSF_REGEX = /\s+(sync|lazy)\s+/g;
```
joshwooding, line 83: Technically, this is not needed, but since Weboack 5 and Rspack both have lazy paths, should this ensure those are ignored?

fi3ework:
Hi @joshwooding, I'm the maintainer of https://github.com/rspack-contrib/storybook-rsbuild. Thanks for addressing this issue for Rsbuild, if this issue is found to be due to Storybook Rsbuild not being in line with some official behaviors. Feel free to let me know to fix that. Thanks.

joshwooding:
> Hi @joshwooding, I'm the maintainer of [rspack-contrib/storybook-rsbuild](https://github.com/rspack-contrib/storybook-rsbuild). Thanks for addressing this issue for Rsbuild, if this issue is found to be due to Storybook Rsbuild not being in line with some official behaviors. Feel free to let me know to fix that. Thanks.

No worries, I'm excited to start using it! I'm not sure if it's an issue or even that it's a problem with Storybook Rsbuild.

The main difference I can see is the path that storybook-stories is imported from 

```
`./node_modules/.cache/storybook/default/dev-server/storybook-stories.js`
```
vs
```
`./storybook-stories.js`,
```

But this could easily just be a difference in webpack vs rspack

fi3ework:
> But this could easily just be a difference in webpack vs rspack

`./storybook-stories.js` should be loaded through virtual modules. However, Rspack currently does not fully support virtual modules. We have opened an RFC at https://github.com/web-infra-dev/rspack/issues/7478 and may add support in the near future. Currently, `./node_modules/.cache/storybook/default/dev-server/storybook-stories.js` is generated using temporary files as a substitute for virtual modules. Until real virtual modules are implemented, our options are limited.

codykaup:
Thanks for the contribution @joshwooding, it's working great over here! I just pushed a new commit to apply all the linting changes I've been making the past couple days (sorry about that!).

In addition, thanks for the extra context and a link to that RFC @fi3ework!

joshwooding:
> > But this could easily just be a difference in webpack vs rspack
> 
> 
> 
> `./storybook-stories.js` should be loaded through virtual modules. However, Rspack currently does not fully support virtual modules. We have opened an RFC at https://github.com/web-infra-dev/rspack/issues/7478 and may add support in the near future. Currently, `./node_modules/.cache/storybook/default/dev-server/storybook-stories.js` is generated using temporary files as a substitute for virtual modules. Until real virtual modules are implemented, our options are limited.

Ah that's what I thought when I was reading about it :) thanks for confirming

node-src/lib/getDependentStoryFiles.ts
codykaup: Yeah I think that makes sense to me!

codykaup (approved):
Thanks for putting this together!

ghengeveld:
:rocket: PR was released in [`v11.10.0`](https://github.com/chromaui/chromatic-cli/releases/tag/v11.10.0) :rocket:

PR #1037, opened 2024-09-11
Add initial ESLint config based on internal repo

codykaup:
We recently went through and decided on a bunch of ESLint rules for one of our internal repos. This simply brings part of that config over and fixes any linting errors based on the new config!

**Note: This does not change any CLI behavior!**

- Updated ESLint to v9
- Added initial ESLint config
  - This does not include `unicorn` and `jsdoc` due to the number of errors from those configs (they will be added in separate PRs)
- Applied autofixes
- Applied manual fixes/disables where necessary

<details>
  <summary>📦 Published PR as canary version: <code>11.9.1--canary.1037.10819729204.0</code></summary>
  <br />

  ```bash
  npm install chromatic@11.9.1--canary.1037.10819729204.0
  # or 
  yarn add chromatic@11.9.1--canary.1037.10819729204.0
  ```
</details>


codykaup:
I forgot to disable ESLint on Codacy since they don't support ESLint 9 yet. 🙈

node-src/io/getDNSResolveAgent.ts
```diff
@@ -7,7 +7,7 @@ export class DNSResolveAgent extends Agent {
       ...options,
       lookup(
         hostname: string,
-        lookupOptions: dns.LookupOneOptions,
+        _lookupOptions: dns.LookupOneOptions,
```
jmhobbs, line 10: `LookupFunction` does not have `lookupOptions`, but [does have `options`](https://nodejs.org/api/dns.html#dnslookuphostname-options-callback), this should probably be that?

test-stories/timing.stories-disabled.js
```diff
@@ -19,7 +18,7 @@ const WaitFor = ({ seconds }) => {
 
   return (
     <div>
-      {Array.from(new Array(seconds - count)).map((x, index) => (
+      {Array.from(new Array(seconds - count)).map((_x, index) => (
```
jmhobbs, line 21: Do we want to fully elide these things?

```suggestion
      {Array.from(new Array(seconds - count)).map((_, index) => (
```

node-src/ui/tasks/verify.ts
```diff
@@ -43,8 +43,8 @@ export const runOnlyNames = (ctx: Context) => ({
     .join(', ')}`,
 });
 
-export const awaitingUpgrades = (ctx: Context, upgradeBuilds: { completedAt?: number }[]) => {
-  const pending = upgradeBuilds!.filter((upgrade) => !upgrade.completedAt).length;
+export const awaitingUpgrades = (_ctx: Context, upgradeBuilds: { completedAt?: number }[]) => {
+  const pending = upgradeBuilds.filter((upgrade) => !upgrade.completedAt).length;
```
jmhobbs, line 47: What happened to the `!` in here and is that a problem?  Also `_ctx` -> `_` maybe?

jmhobbs (commented):
Couple questions and tweaks in here, overall looks great.

Could be a patch instead of a minor I think, since no new features were added.

test-stories/timing.stories-disabled.js
codykaup: Yeah I was on the fence with that one. We can go with `_`!

node-src/ui/tasks/verify.ts
codykaup: I'm honestly not sure why `!` was there unless the `upgradeBuilds` arg was optional at one point. Since it's a required field, we don't need the non-null assertion there.

node-src/io/getDNSResolveAgent.ts
codykaup: Yeah we can update that! They're positional arguments so it's not a huge deal what they're called.

jmhobbs (approved):

ghengeveld:
:rocket: PR was released in [`v11.10.0`](https://github.com/chromaui/chromatic-cli/releases/tag/v11.10.0) :rocket:

PR #1062, opened 2024-09-27
Enable `unicorn/filename-case` ESLint rule

codykaup:
Simply enables our `unicorn/filename-case` ESLint rule and fixes any errors that appear.

<details>
  <summary>📦 Published PR as canary version: <code>11.11.1--canary.1062.11128106915.0</code></summary>
  <br />

  ```bash
  npm install chromatic@11.11.1--canary.1062.11128106915.0
  # or 
  yarn add chromatic@11.11.1--canary.1062.11128106915.0
  ```
</details>


bin-src/__mocks__/previewStatsJson/preview-stats.trimmed.json
```diff
@@ -643,11 +643,11 @@
       ]
     },
     {
-      "id": "./node-src/ui/messages/errors/noCSFGlobs.stories.ts",
-      "name": "./node-src/ui/messages/errors/noCSFGlobs.stories.ts + 1 modules",
+      "id": "./node-src/ui/messages/errors/noCsfGlobs.stories.ts",
+      "name": "./node-src/ui/messages/errors/noCsfGlobs.stories.ts + 1 modules",
```
jmhobbs, line 647: Is this a valid rename in our rules?  `CSF` is an initialism, not a word.   Would we do `Tcp` instead of `TCP`?

jmhobbs (approved):
I'm a little on the fence about the subcommand files.  There is something nice about the symmetry of `trim-stats-file` being found in `trim-stats-file.ts` instead of `trimStatsFile.ts`

bin-src/__mocks__/previewStatsJson/preview-stats.trimmed.json
jmhobbs: We also do `Dns` instead of `DNS` which feels weird.

jmhobbs: This is not a deal breaker for me.

codykaup: > Would we do `Tcp` instead of `TCP?`

Yes we would. 😞 

I honestly wasn't sure how I felt about it other than creating better separation in the filenames. I'd be down to flip them back though, WDYT?

codykaup: I think that's probably fine. I'll look into adding those exceptions and moving those files back!

codykaup:
> I'm a little on the fence about the subcommand files. There is something nice about the symmetry of `trim-stats-file` being found in `trim-stats-file.ts` instead of `trimStatsFile.ts`

@jmhobbs So here's a funny thing... I'm actually the opposite 😅. To me, the filenames should all match and my brain translates the command name into the camel case for the filename when I go to search for it.

PR #1064, opened 2024-09-30
Enable `unicorn/prevent-abbreviations` ESLint rule

codykaup:
Simply enables our `unicorn/prevent-abbreviations` ESLint rule and fixes any errors that appear. I also allowed/ignored certain errors if it made sense in the context of the project. Let me know if you'd like me to tweak those!

<details>
  <summary>📦 Published PR as canary version: <code>11.10.5--canary.1064.11110716020.0</code></summary>
  <br />

  ```bash
  npm install chromatic@11.10.5--canary.1064.11110716020.0
  # or 
  yarn add chromatic@11.10.5--canary.1064.11110716020.0
  ```
</details>


node-src/index.ts
```diff
@@ -302,7 +306,7 @@ export async function getGitInfo(): Promise<GitInfo> {
     uncommittedHash,
     userEmail,
     userEmailHash,
-    repositoryRootDir,
+    repositoryRootDir: repositoryRootDirectory,
```
jmhobbs, line 309: Is this return something we can change, or is it in public scope?

node-src/lib/utils.ts
```diff
@@ -3,19 +3,19 @@ import picomatch, { Matcher } from 'picomatch';
 export const lcfirst = (str: string) => `${str.charAt(0).toLowerCase()}${str.slice(1)}`;
 
 export const delay = (ms: number) => new Promise<void>((resolve) => setTimeout(resolve, ms));
-export const tick = async (times: number, interval: number, fn: (i: number) => any) => {
-  for (let i = 0; i < times; i += 1) {
+export const tick = async (times: number, interval: number, function_: (index: number) => any) => {
```
jmhobbs, line 6: Could be more descriptive than `function_` I think, maybe `callback`?

jmhobbs (approved):

node-src/lib/utils.ts
codykaup: Ope! Good find, not sure how I missed that one.

node-src/index.ts
codykaup: I think we'd be able to change it but it'll be a pain since the output is used in [chromaui/addon-visual-tests](https://github.com/chromaui/addon-visual-tests). Perhaps I should leave a comment about that here. I had to go hunting to figure out why this exists but isn't called anywhere.

PR #1070, opened 2024-10-01
Add TypeScript config from internal repo and fix some errors

codykaup:
This adds _most_ of the `tsconfig.json` from our internal repo with one exception: `strict: true`. Therefore, this is mostly fixing random errors that appear when enabling strict mode. There were 273 errors in 60 files when I first turned this on. I fixed a little more than half with this PR and figured that was a good place to stop. Unfortunately, there's 109 errors in 32 files remaining after this.

_Note: I focused on errors that should have zero impact on the flow of the CLI. There are some other changes we'll have to make that are a bit riskier so I left those for future me._


node-src/tasks/build.ts
```diff
@@ -30,7 +30,7 @@ export const setSourceDirectory = async (ctx: Context) => {
 export const setBuildCommand = async (ctx: Context) => {
   const webpackStatsSupported =
     ctx.storybook && ctx.storybook.version
-      ? semver.gte(semver.coerce(ctx.storybook.version), '6.2.0')
+      ? semver.gte(semver.coerce(ctx.storybook.version) || '', '6.2.0')
```
jmhobbs, line 33: Would it be better to have a "valid" fallback version string here for clarity?
```suggestion
      ? semver.gte(semver.coerce(ctx.storybook.version) || '0.0.0', '6.2.0')
```

node-src/tasks/upload.ts
```diff
@@ -119,10 +119,10 @@ export const validateFiles = async (ctx: Context) => {
 export const traceChangedFiles = async (ctx: Context, task: Task) => {
   if (!ctx.turboSnap || ctx.turboSnap.unavailable) return;
   if (!ctx.git.changedFiles) return;
-  if (!ctx.fileInfo.statsPath) {
+  if (!ctx.fileInfo?.statsPath) {
     // If we don't know the SB version, we should assume we don't support `--stats-json`
     const nonLegacyStatsSupported =
-      ctx.storybook?.version && semver.gte(semver.coerce(ctx.storybook.version), '8.0.0');
+      ctx.storybook?.version && semver.gte(semver.coerce(ctx.storybook.version) || '', '8.0.0');
```
jmhobbs, line 125: Same here
```suggestion
      ctx.storybook?.version && semver.gte(semver.coerce(ctx.storybook.version) || '0.0.0', '8.0.0');
```

node-src/ui/messages/info/storybookPublished.ts
```diff
@@ -15,12 +15,12 @@ export default ({ build, storybookUrl }: Pick<Context, 'build' | 'storybookUrl'>
     return dedent(chalk`
       ${success} {bold Storybook published}
       We found ${components} with ${stories}.
-      ${info} View your Storybook at ${link(storybookUrl)}
+      ${info} View your Storybook at ${link(storybookUrl || '')}
```
jmhobbs, line 18: Is this is not present, shouldn't we have an error?  Can we make `storybookUrl` required?

node-src/tasks/upload.ts
codykaup: Yeah that's a good call. I'll make that change!

node-src/ui/messages/info/storybookPublished.ts
codykaup: > Can we make storybookUrl required?

Yes and we should! I'll fix that up.

jmhobbs (approved):

PR #1072, opened 2024-10-02
Fix remaining TypeScript errors and unable `strict: true`

codykaup:
This is a continuation from #1070 by cleaning up the remaining errors and finally enabling strictness in our `tsconfig.json`! 🎉 

I tried to make these changes as small as possible to not change the flow but some of the errors made that difficult. In theory, a good chunk of these rely on values existing despite the `Context` type having them as optional so throwing new errors when they don't exist shouldn't do anything in practice. 

<details>
  <summary>📦 Published PR as canary version: <code>11.11.1--canary.1072.11163487586.0</code></summary>
  <br />

  ```bash
  npm install chromatic@11.11.1--canary.1072.11163487586.0
  # or 
  yarn add chromatic@11.11.1--canary.1072.11163487586.0
  ```
</details>

### DIFF
--- a/node-src/lib/getDependentStoryFiles.ts
+++ b/node-src/lib/getDependentStoryFiles.ts
@@ -9,7 +9,6 @@ import { isPackageManifestFile, matchesFile } from './utils';
 
 type FilePath = string;
 type NormalizedName = string;
-// type TraceToCheck = (string | number)[];
 type TraceToCheck = (string | number | string[])[];
 
 // Bail whenever one of these was changed
--- a/node-src/lib/getDependentStoryFiles.ts
+++ b/node-src/lib/getDependentStoryFiles.ts
@@ -9,6 +9,8 @@ import { isPackageManifestFile, matchesFile } from './utils';
 
 type FilePath = string;
 type NormalizedName = string;
+// type TraceToCheck = (string | number)[];
+type TraceToCheck = (string | number | string[])[];
 
 // Bail whenever one of these was changed
 const LOCKFILES = [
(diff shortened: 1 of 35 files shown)


node-src/lib/getDependentStoryFiles.ts
```diff
@@ -9,6 +9,8 @@ import { isPackageManifestFile, matchesFile } from './utils';
 
 type FilePath = string;
 type NormalizedName = string;
+// type TraceToCheck = (string | number)[];
```
jmhobbs, line 12: Accidental leftover here?

node-src/lib/getPackageManager.ts
```diff
@@ -13,7 +13,7 @@ export const getPackageManagerRunCommand = async (args: string[]) => {
 
 // e.g. `8.19.2`
 export const getPackageManagerVersion = async (packageManager: string) => {
-  const { stdout } = await execa(packageManager || (await getPackageManagerName()), ['--version']);
+  const { stdout } = await execa(packageManager, ['--version']);
```
jmhobbs, line 16: What does the caller look like on this?  Can we be sure `packageManager` will actually be a non-empty `string`?

jmhobbs (approved):

node-src/lib/getPackageManager.ts
codykaup: The caller grabs the package manager before calling `getPackageManagerVersion` [here](https://github.com/chromaui/chromatic-cli/blob/c7a3f96ae59431e3edde1b757b0cb5cf0d45ee0d/node-src/tasks/initialize.ts#L55-L59) which feels like the right flow rather than magically grabbing the package manager if it's not passed in.

I think you make a good point though, we should handle `packageManager: ''` directly instead of letting `execa` throw for us. I'll add that layer quick!

PR #1096, opened 2024-10-16
Update task output to account for E2E

codykaup:
When using our E2E integration ([Playwright](https://www.chromatic.com/docs/playwright/) or [Cypress](https://www.chromatic.com/docs/cypress/)), it can be rather confusing to see "Storybook" and "stories" everywhere. This changes those to reference "Test suite" and "tests" through the UI.

The biggest change I had to make for this was to pass the main `Context` around to set the task title and output based on if it's an E2E build or not.

<details>
  <summary>📦 Published PR as canary version: <code>11.14.0--canary.1096.11448057249.0</code></summary>
  <br />

  ```bash
  npm install chromatic@11.14.0--canary.1096.11448057249.0
  # or 
  yarn add chromatic@11.14.0--canary.1096.11448057249.0
  ```
</details>

### DIFF
--- a/node-src/ui/tasks/storybookInfo.ts
+++ b/node-src/ui/tasks/storybookInfo.ts
@@ -1,3 +1,4 @@
+import { isE2EBuild } from '../../lib/e2eUtils';
 import { Context } from '../../types';
 
 const capitalize = (string: string) =>
@@ -6,7 +7,12 @@ const capitalize = (string: string) =>
     .map((str) => str.charAt(0).toUpperCase() + str.slice(1))
     .join(' ');
 
-const infoMessage = ({ addons, version, viewLayer, builder }: Context['storybook']) => {
+const infoMessage = (ctx: Context) => {
+  if (isE2EBuild(ctx.options)) {
+    return ctx.options.playwright ? 'Playwright for E2E' : 'Cypress for E2E';
+  }
+
+  const { addons, version, viewLayer, builder } = ctx.storybook;
   const info = version && viewLayer ? `Storybook ${version} for ${capitalize(viewLayer)}` : '';
   const builderInfo = builder
     ? `${info}; using the ${builder.name} builder (${builder.packageVersion})`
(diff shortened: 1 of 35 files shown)


.github/workflows/smoke-test-node-api.yml
```diff
@@ -13,7 +13,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-node@v3
         with:
-          node-version: latest
+          node-version: 'lts/*'
```
codykaup, line 16: I was getting [some errors](https://github.com/chromaui/chromatic-cli/actions/runs/11373960183/job/31642317236) in this action related to

```
YN0018: resolve@patch:resolve@npm%3A1.22.8#optional!builtin<compat/resolve>::version=1.22.8&hash=c3c19d: The remote archive doesn't match the expected checksum
```

Turns out, it was because the `latest` version of Node was bumped to v23.0.0. We can set this to `lts` to fix that issue. I don't think we need to test against `latest` but I'm open to changing this back.

node-src/ui/tasks/storybookInfo.ts
```diff
@@ -1,12 +1,19 @@
...
 const capitalize = (string: string) =>
   string
     .split('-')
     .map((str) => str.charAt(0).toUpperCase() + str.slice(1))
     .join(' ');
 
-const infoMessage = ({ addons, version, viewLayer, builder }: Context['storybook']) => {
+const infoMessage = (ctx: Context) => {
+  if (isE2EBuild(ctx.options)) {
+    return ctx.options.playwright ? 'Playwright for E2E' : 'Cypress for E2E';
```
jmhobbs, line 13: Should we maybe do a switch case on these?  Just thinking, if we ever add another E2E type we would catch it with exhaustive case checks.  I guess that would require some type changes too, so maybe not worth the squeeze.

jmhobbs (approved):
LGTM, just one question in there.

node-src/ui/tasks/storybookInfo.ts
codykaup: I thought about that but I wasn't sure what the `default` should be. `[unknown]` or something?

codykaup: I'm good with `E2E` since the E2E checks are a bit all over the place. I'd hate to throw here if you forget to update the UI message after changing `isE2EBuild()`.

codykaup: @jmhobbs Moved it around a bit. Let me know what you think!

jmhobbs: Dig it, thank you!

PR #1102, opened 2024-10-18
Setup workflow for PR checks

codykaup:
We've been running our status checks on every `push` but we want to shift to only run them when there's a pull request. This changes that _and_ moves to a single workflow to manage everything (since we don't need to smoke test if lint and/or tests fail)!

image

<details>
  <summary>📦 Published PR as canary version: <code>11.12.7--canary.1102.11410064469.0</code></summary>
  <br />

  ```bash
  npm install chromatic@11.12.7--canary.1102.11410064469.0
  # or 
  yarn add chromatic@11.12.7--canary.1102.11410064469.0
  ```
</details>


node-src/git/getCommitAndBranch.test.ts
```diff
@@ -62,6 +71,10 @@ describe('getCommitAndBranch', () => {
   });
 
   it('retrieves CI context', async () => {
+    process.env.GITHUB_EVENT_NAME = 'push';
```
codykaup, line 74: All these tests were accidentally working since the event name was `push` in the past. This forces the behavior so they should work no matter how we trigger these workflows.

jmhobbs (approved):
I love this!

PR #1107, opened 2024-10-21
Add support for `logLevel` and `logPrefix` options, use local timestamp as default prefix

ghengeveld:
This is to allow the VTA to lower the log level and add a custom prefix to each log line in order to distinguish them from Storybook's own logs. The default prefix is a timestamp (e.g. `14:30:00.123`) in the local timezone.

Previously we would prefix each line with a UTC-based timestamp, but that only really worked when explicitly using the `--debug` or `--no-interactive` flag, or when writing to a log file. We now properly support `debug` and `interactive` as options to the logger as well. The logger no longer directly reads from `process.argv` but we pass in parsed arguments as flags instead. This doesn't affect existing setups except for GitHub Actions when using the `debug: true` or `interactive: false` option. Not a breaking change though.

Also we now preserve colors when using non-interactive or debug mode.

<details>
  <summary>📦 Published PR as canary version: <code>11.15.0--canary.1107.11482589542.0</code></summary>
  <br />

  ```bash
  npm install chromatic@11.15.0--canary.1107.11482589542.0
  # or 
  yarn add chromatic@11.15.0--canary.1107.11482589542.0
  ```
</details>

### DIFF
--- a/action-src/main.ts
+++ b/action-src/main.ts
@@ -116,6 +116,8 @@ async function run() {
     const forceRebuild = getInput('forceRebuild');
     const ignoreLastBuildOnBranch = getInput('ignoreLastBuildOnBranch');
     const logFile = getInput('logFile');
+    const logLevel = getInput('logLevel');
+    const logPrefix = getInput('logPrefix');
     const only = getInput('only');
     const onlyChanged = getInput('onlyChanged');
     const onlyStoryFiles = getMultilineInput('onlyStoryFiles');
@@ -169,6 +171,8 @@ async function run() {
         ignoreLastBuildOnBranch: maybe(ignoreLastBuildOnBranch),
         interactive: false,
         logFile: maybe(logFile),
+        logLevel: maybe(logLevel),
+        logPrefix: maybe(logPrefix),
         only: maybe(only),
         onlyChanged: maybe(onlyChanged),
         onlyStoryFiles: maybe(onlyStoryFiles),
--- a/node-src/lib/log.ts
+++ b/node-src/lib/log.ts
@@ -5,13 +5,14 @@ import stripAnsi from 'strip-ansi';
 import { format } from 'util';
 
 import { errorSerializer } from './logSerializers';
+import { Flags, Options } from '../types';
 
 interface QueueMessage {
   type: LogType;
   messages: string[];
 }
 
-const { DISABLE_LOGGING, LOG_LEVEL = '' } = process.env;
+const { DISABLE_LOGGING, LOG_LEVEL = '', LOG_PREFIX } = process.env;
 const LOG_LEVELS = { silent: 0, error: 1, warn: 2, info: 3, debug: 4 };
 const DEFAULT_LEVEL = 'info';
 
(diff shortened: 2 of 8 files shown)


node-src/lib/log.ts
```diff
@@ -4,47 +4,69 @@ import { createWriteStream, rm } from 'fs';
...
 
-const { DISABLE_LOGGING, LOG_LEVEL = '' } = process.env;
 const LOG_LEVELS = { silent: 0, error: 1, warn: 2, info: 3, debug: 4 };
 const DEFAULT_LEVEL = 'info';
 
 // Top-level promise rejection handler to deal with initialization errors
 const handleRejection = (reason: string) => console.error('Unhandled promise rejection:', reason);
 process.on('unhandledRejection', handleRejection);
 
+const isPrintable = (value: any, type = typeof value) =>
```
codykaup, line 22: Do we need the `type` argument here? Should we just grab the type in the function itself?

node-src/lib/log.ts
```diff
@@ -93,25 +115,36 @@ const fileLogger = {
   },
 };
 
-export const createLogger = () => {
-  let level = (LOG_LEVEL.toLowerCase() as keyof typeof LOG_LEVELS) || DEFAULT_LEVEL;
-  if (DISABLE_LOGGING === 'true') level = 'silent';
+/* eslint-disable-next-line complexity */
```
codykaup, line 118: Should we refactor this function to lower the complexity instead of disabling the rule?

node-src/lib/log.ts
```diff
@@ -93,25 +115,36 @@ const fileLogger = {
...
+export const createLogger = (flags: Flags, options?: Partial<Options>) => {
```
codykaup, line 119: Out of curiosity, why didn't we use a log package instead of writing it ourselves?

package.json
```diff
@@ -69,7 +69,7 @@
     "isChromatic.d.ts"
   ],
   "scripts": {
-    "build": "clean-package && tsup && clean-package restore",
+    "build": "clean-package ; tsup ; clean-package restore",
```
codykaup, line 72: 👏

codykaup (commented):
Is there a ticket associated with these changes somewhere (E2E repo or Linear)?

package.json
jmhobbs: When we do a build in production, don't we want this to fail if `clean-package` breaks?

ghengeveld: Ah yes, I'll revert the first one back to `&&` so tsup won't run if clean-package fails, but if tsup fails we'll still restore.

action-src/main.ts
```diff
@@ -116,6 +116,8 @@ async function run() {
     const forceRebuild = getInput('forceRebuild');
     const ignoreLastBuildOnBranch = getInput('ignoreLastBuildOnBranch');
     const logFile = getInput('logFile');
+    const logLevel = getInput('logLevel');
+    const logPrefix = getInput('logPrefix');
```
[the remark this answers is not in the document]
ghengeveld: Yes, good catch. Updated.

node-src/lib/log.ts
ghengeveld: I tried, and it wasn't getting any simpler. Pulling chunks of code out into separate functions only adds extra layers of indirection, it doesn't actually simplify anything.

node-src/lib/log.ts
ghengeveld: Because I tried to have the least amount of dependencies, and I needed one that could queue up log messages to print them later (due to how interactive mode works, we can't print anything while Listr is running).

node-src/lib/log.ts
codykaup: Gotcha. We might have to rethink how logging is structured then. That's completely outside of the scope of this change though so don't worry about that. Thanks for trying!

codykaup (commented):
Code looks good but we should figure out the `yarn build` script. I wouldn't be against creating a ticket for fixing that and pulling it out of this PR.

codykaup (approved):
:shipit:

PR #1110, opened 2024-10-28
Account for `rsbuild` stats JSON output and multiple locations

codykaup:
At some point the `rsbuild` cache location changed which broke Turbosnap. Simply adding the new location to our list fixes that issue.

I also noticed the stats output is slightly different than Webpack/Vite so we lean on `resolvedModule` (which only exists in `rsbuild`) but fallback to `moduleName` when it doesn't exist.

Both of these changes allow `rsbuild` to work on the newer versions!

_Note: We've been doing some external testing with the canary and things are working great!_

<details>
  <summary>📦 Published PR as canary version: <code>11.16.2--canary.1110.11566394830.0</code></summary>
  <br />

  ```bash
  npm install chromatic@11.16.2--canary.1110.11566394830.0
  # or 
  yarn add chromatic@11.16.2--canary.1110.11566394830.0
  ```
</details>

### DIFF
--- a/node-src/lib/getDependentStoryFiles.test.ts
+++ b/node-src/lib/getDependentStoryFiles.test.ts
@@ -179,7 +179,10 @@ describe('getDependentStoryFiles', () => {
         id: String.raw`./src lazy recursive ^\.\/.*$`,
         reasons: [
           {
-            moduleName: './node_modules/.cache/storybook/default/dev-server/storybook-stories.js',
+            resolvedModule:
+              './node_modules/.cache/storybook/default/dev-server/storybook-stories.js',
+            moduleName:
+              './node_modules/.cache/storybook/default/dev-server/storybook-stories.js + 2 modules',
           },
         ],
       },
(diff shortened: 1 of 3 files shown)


node-src/lib/getDependentStoryFiles.test.ts
```diff
@@ -179,7 +179,10 @@ describe('getDependentStoryFiles', () => {
         id: String.raw`./src lazy recursive ^\.\/.*$`,
         reasons: [
           {
-            moduleName: './node_modules/.cache/storybook/default/dev-server/storybook-stories.js',
+            resolvedModule:
```
jmhobbs, line 182: If this is `rsbuild` specific, should we have a comment in the tests about that?

jmhobbs (approved):

node-src/lib/getDependentStoryFiles.test.ts
codykaup: This test name mentions `rspack` already. Does that cover it or should I add a note quick?

jmhobbs: Nope, that sounds good!  I didn't expand up to see that, my bad! 👍

PR #1113, opened 2024-10-30
Add steps for how to run builds against local CLI

codykaup:
Just as the title says - this adds docs on how to run a separate project against your local copy of the CLI.

<details>
  <summary>📦 Published PR as canary version: <code>11.16.4--canary.1113.11600333794.0</code></summary>
  <br />

  ```bash
  npm install chromatic@11.16.4--canary.1113.11600333794.0
  # or 
  yarn add chromatic@11.16.4--canary.1113.11600333794.0
  ```
</details>

### DIFF
--- a/scripts/release.mjs
+++ b/scripts/release.mjs
@@ -60,7 +60,7 @@ async function build() {
   await $({
     stdout: 'inherit',
     stderr: 'inherit',
-  })`sentry-cli releases set-commits --auto ${nextVersion}`;
+  })`sentry-cli releases set-commits --auto ${nextVersion} --ignore-missing`;
 
   console.info('🧹 Removing sourcemaps from build');
   await $`yarn clean:sourcemaps`;
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Other platforms/versions may work, but are not officially supported. Certain fea
 
 ## Contributing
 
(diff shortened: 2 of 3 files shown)


jmhobbs (approved):
Do we want to mention the `node <path/to/dist>` option as well?  Or call out the downsides of that method?

codykaup:
> Do we want to mention the `node <path/to/dist>` option as well? Or call out the downsides of that method?

Good idea! I can call out the downsides here too.

scripts/release.mjs
```diff
@@ -60,7 +60,7 @@ async function build() {
   await $({
     stdout: 'inherit',
     stderr: 'inherit',
-  })`sentry-cli releases set-commits --auto ${nextVersion}`;
+  })`sentry-cli releases set-commits --auto ${nextVersion} --ignore-missing`;
```
codykaup, line 63: I rebased at some point which broke the `sentry-cli` because it couldn't find the commits. Passing `--ignore-missing` allows the CLI to fallback to the default behavior if it hits this error. Based on their [docs](https://docs.sentry.io/cli/releases/#dealing-with-missing-commits), passing this every time shouldn't have a downside.

jmhobbs: Send it!

PR #1120, opened 2024-11-12
Send `storybookBaseDir` as part of project metadata

tmeasday:
And refactor things a little so we calculate it upfront rather than as part of TS.

This means we will need to re-test some TS stuff as part of this PR, to make sure I haven't messed it up. @ghengeveld let us know if you have any good playbooks there.

<details>
  <summary>📦 Published PR as canary version: <code>11.19.0--canary.1120.11790916078.0</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install chromatic@11.19.0--canary.1120.11790916078.0
  # or 
  yarn add chromatic@11.19.0--canary.1120.11790916078.0
  ```
</details>

PR #1112, opened 2024-10-30
Detect `context.projectMetadata.hasRouter` and send to the index

tmeasday:
This PR is blocked on an index change (to come).

To QA:

- Create a routing based project (e.g. SB Next Sandbox)
- Run a build, check it sets `hasRouter=true`
- Create a non-routed project (e.g. `react-vite` Sandbox)
- Run a build, check it sets `hasRouter=false`

Use this review app: https://github.com/chromaui/chromatic/pull/9061

<details>
  <summary>📦 Published PR as canary version: <code>11.17.0--canary.1112.11702183022.0</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install chromatic@11.17.0--canary.1112.11702183022.0
  # or 
  yarn add chromatic@11.17.0--canary.1112.11702183022.0
  ```
</details>

PR #1122, opened 2024-11-13
Send project metadata to the index

tmeasday:
Reverts chromaui/chromatic-cli#1121 (re-implements https://github.com/chromaui/chromatic-cli/pull/1112) and incorporates https://github.com/chromaui/chromatic-cli/pull/1120.

Also:

 - added some error/timeout handling so we don't get stuck if there is something wrong with git.
 - shorted timeouts to 5s as it's not critical this works (we could shorten further?)
 - don't use `wc` or `head` as they may not be there. Instead:
   - simulate `head -n 1` via streams
   - just count the number of lines for `wc`. 
      - This is inefficient as it requires loading all the lines into memory. 
      - I'm not sure if this will be a problem or not if a repo has a heap of contributors or "app files".  
      - I could implement another `execGitCommandCountLines` but I figured it was better to keep the complexity down.

Things to check in QA:

- environments without access to `head` or `wc`.
- environments where these commands timeout [not sure how to simulate]

<details>
  <summary>📦 Published PR as canary version: <code>11.20.0--canary.1122.12129895074.0</code></summary>
  <br />

  ```bash
  npm install chromatic@11.20.0--canary.1122.12129895074.0
  # or 
  yarn add chromatic@11.20.0--canary.1122.12129895074.0
  ```
</details>

### DIFF
--- a/node-src/lib/getDependentStoryFiles.test.ts
+++ b/node-src/lib/getDependentStoryFiles.test.ts
@@ -2,26 +2,30 @@
 import chalk from 'chalk';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 
-import * as git from '../git/git';
 import { getDependentStoryFiles, normalizePath } from './getDependentStoryFiles';
-
-vi.mock('../git/git');
+import { Context } from '../../dist/node';
 
 const CSF_GLOB = String.raw`./src sync ^\.\/(?:(?!\.)(?=.)[^/]*?\.stories\.js)$`;
 const VITE_ENTRY = '/virtual:/@storybook/builder-vite/storybook-stories.js';
 const statsPath = 'preview-stats.json';
 
 const log = { info: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn() };
-const getContext: any = ({
-  configDir,
-  staticDir,
-  ...options
-}: { configDir?: string; staticDir?: string } = {}) => ({
+const getContext: any = (
+  {
+    configDir,
+    staticDir,
+    ...options
+  }: { configDir?: string; staticDir?: string } & Context['options'] = {} as any
+) => ({
   log,
-  options: { storybookBaseDir: '.', ...options },
+  options,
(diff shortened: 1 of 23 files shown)


jmhobbs:
> just count the number of lines for wc.
>  This is inefficient as it requires loading all the lines into memory.

I think it's a good trade off to add a `execGitCommandCountLines`.  I'd rather read newlines out of the stream and discard with one additional function than keep unbounded values in memory, twice after the `.split('\n')`

tmeasday:
@jmhobbs done! Looks like it still works (2 and 0 are the correct values for this repo):

image

node-src/lib/getDependentStoryFiles.test.ts
```diff
@@ -2,26 +2,30 @@
...
+  }: { configDir?: string; staticDir?: string } & Context['options'] = {} as any
+) => ({
   log,
-  options: { storybookBaseDir: '.', ...options },
+  options,
   turboSnap: {},
-  storybook: { configDir, staticDir },
-  git: {},
+  storybook: {
+    baseDir: options.storybookBaseDir ? options.storybookBaseDir.replace(/^\.[/\\]?/, '') : '',
```
jmhobbs, line 24: Could you throw a comment on here about that regex?

jmhobbs (approved):
Haven't QA'd, and there's some CI issues, but overall looks right to me.

tmeasday:
@jmhobbs FYI the CI issues (outside of Codacy) is due to the production index not yet having `projectMetadata.storybookBaseDir` supported yet (need to merge the relevant PR on the other end). This is a good thing I guess (we can't merge this until that's done!)

node-src/lib/getDependentStoryFiles.test.ts
tmeasday: I got rid of this. I think it's simpler to just pass up the string the user has configured and do any re-mapping on the backend.

PR #1143, opened 2025-01-16
Don't normalize package.json fields

### LINKED ISSUE #1123: CLI exits silently when package.json version field does not contain patch number
"version": "0.2-dev"

npx chromatic exits instantly with no output.

"version": "0.2.0-dev"

Chromatic starts as expected.

codykaup:
Fixes #1123 

This fixes a weird bug where an invalid semver string in the `version` field of a user's `package.json` causes the CLI to silently exit. Turns out, it comes from normalizing the `package.json` file from `read-package-up` and `meow`. This disables that behavior to prevent this issue.


jmhobbs (approved):
What else does normalization do?  Are we using any of it?

codykaup:
> What else does normalization do? Are we using any of it?

It does a [bunch of stuff](https://github.com/npm/normalize-package-data#what-normalization-currently-entails) but it doesn't appear we use any of it (I could be wrong though).

PR #1157, opened 2025-02-19
Fix moduleName resolution for Storybook files with import cycles.

jmhobbs:
Newer versions of our webpack builder is now emitting `resolvedModule`, or perhaps always did in an undocumented manner.

This is generally not a problem, as we used `resolvedModule` for rspack and it is generally `null` from webpack, but when you have import cycles you can get a value which is not what we desire from it.

Removing it works with current rspack builders as well.


codykaup (approved):
LGTM!

PR #1166, opened 2025-03-20
Preserve git history when publishing new GH action version

justin-thurman:
This PR updates our `publishAction.mjs` script to push updates to the action repository, rather than clobbering its git history on every update. This allows users to pin their action dependency to a particular commit SHA rather than a (mutable) tag.


scripts/publishAction.mjs
```diff
@@ -17,33 +17,36 @@ const publishAction = async ({ major, version, repo }) => {
...
   await $$`git add .`;
   await $$`git commit -m v${version}`;
-  await $$`git tag -a v${version} -m ${`v${version} without automatic upgrades (pinned)`}`;
-  await $$`git tag -a v${major} -m ${`v${version} with automatic upgrades to v${major}.x.x`}`;
-  await $$`git tag -a v1 -m ${`Deprecated, use 'latest' tag instead`}`;
-  await $$`git tag -a latest -m ${`v${version} with automatic upgrades to all versions`}`;
+  await $$`git tag -f -a v${version} -m ${`v${version} without automatic upgrades (pinned)`}`;
+  await $$`git tag -f -a v${major} -m ${`v${version} with automatic upgrades to v${major}.x.x`}`;
+  await $$`git tag -f -a v1 -m ${`Deprecated, use 'latest' tag instead`}`;
+  await $$`git tag -f -a latest -m ${`v${version} with automatic upgrades to all versions`}`;
```
justin-thurman, line 44: Potentially updating tags now, since they might already exist.

codykaup (approved):

justin-thurman:
I removed the `-f` flag from `git tag`. I also moved the `git tag` executions into the non-dry-run branch since I'd imagine, when messing around with `--dry-run`, devs might not have updated the version in the `package.json` yet, so `git tag` would error.

scripts/publishAction.mjs
```diff
@@ -17,33 +17,35 @@ const publishAction = async ({ major, version, repo }) => {
...
-  await $$`git tag -a v1 -m ${`Deprecated, use 'latest' tag instead`}`;
-  await $$`git tag -a latest -m ${`v${version} with automatic upgrades to all versions`}`;
 
   if (dryRun) {
     console.info('✅ Skipping git push due to --dry-run');
   } else {
-    await $$`git push origin HEAD:main --force`;
+    await $$`git tag -a v${version} -m ${`v${version} without automatic upgrades (pinned)`}`;
+    await $$`git tag -a v${major} -m ${`v${version} with automatic upgrades to v${major}.x.x`}`;
+    await $$`git tag -a latest -m ${`v${version} with automatic upgrades to all versions`}`;
```
codykaup, line 47: Would these error without `-f` since they're likely to exist?

justin-thurman: Oh, right. 🤦 We only want to error for the exact `version` tags. Fixing...

codykaup (approved):
:shipit:

PR #1169, opened 2025-04-02
:pencil2: Update help text to reflect actual state of exitOnceUploaded.

jwir3:
We previously suggested that users use the flag --exit-once-uploaded, but now that we have a workflow parameter for this, our github action can actually take advantage of reporting the correct way to pass this option.

Fixes CAP-2374.

<details>
  <summary>📦 Published PR as canary version: <code>11.28.1--canary.1169.14247692752.0</code></summary>
  <br />

  ```bash
  npm install chromatic@11.28.1--canary.1169.14247692752.0
  # or 
  yarn add chromatic@11.28.1--canary.1169.14247692752.0
  ```
</details>


justin-thurman (approved):
Looks good. Do you think it's worth keeping the reference to the `--exit-once-uploaded` flag, since it is still a valid choice? Though it would make this help message a bit overloaded. I'm fine either way.

PR #1178, opened 2025-04-30
Update the replacement build logic to search for multiple replacement builds

ethriel3695:
We can run into a case at times where the first replacement build times out, and it immediately throws an error and doesn't attempt to resolve the second or third parent commits for the replacement build.

This logic accounts for that and will move on to the next replacement build until a suitable one is found.

<details>
  <summary>📦 Published PR as canary version: <code>11.29.0--canary.1178.15008534431.0</code></summary>
  <br />

  ```bash
  npm install chromatic@11.29.0--canary.1178.15008534431.0
  # or 
  yarn add chromatic@11.29.0--canary.1178.15008534431.0
  ```
</details>

### DIFF
--- a/node-src/git/getChangedFilesWithReplacement.test.ts
+++ b/node-src/git/getChangedFilesWithReplacement.test.ts
@@ -2,11 +2,13 @@ import { beforeEach, describe, expect, it, vi } from 'vitest';
 
 import TestLogger from '../lib/testLogger';
 import { getChangedFilesWithReplacement } from './getChangedFilesWithReplacement';
-import { getChangedFiles } from './git';
 
 vi.mock('./git', () => ({
   getChangedFiles: (hash) => {
     if (/exists/.test(hash)) return ['changed', 'files'];
+    if (/exists/.test('timeout')) {
+      throw new Error('Command timed out after 20000ms');
+    }
     throw new Error(`fatal: bad object ${hash}`);
   },
   commitExists: (hash) => hash.match(/exists/),
@@ -127,41 +129,6 @@ describe('getChangedFilesWithReplacements', () => {
     ).rejects.toThrow(/fatal: bad object missing/);
   });
 
-  it('tries a replacement builds until it finds a working one', async () => {
-    const firstWorkingBuild = {
-      id: 'working',
-      number: 2,
-      commit: 'exists',
-      uncommittedHash: '',
-      isLocalBuild: false,
-    };
-    const secondWorkingBuild = {
-      id: 'working',
-      number: 1,
-      commit: 'exists',
-      uncommittedHash: '',
-      isLocalBuild: false,
-    };
-    client.runQuery
-      .mockReturnValueOnce({ app: { build: { ancestorBuilds: [firstWorkingBuild] } } })
-      .mockReturnValueOnce({ app: { build: { ancestorBuilds: [secondWorkingBuild] } } });
-
-    const result = await getChangedFilesWithReplacement(context, {
-      id: 'id',
-      number: 3,
-      commit: 'missing',
-      uncommittedHash: '',
-      isLocalBuild: false,
-    });
-
-    expect(result).toEqual({
-      changedFiles: ['changed', 'files'],
-      replacementBuild: firstWorkingBuild,
-    });
-
-    expect(client.runQuery).toHaveBeenCalledTimes(1);
-  });
-
   it('tries multiple replacement builds in the same batch', async () => {
     const failingBuild1 = {
       id: 'failing1',
@@ -255,15 +222,10 @@ describe('getChangedFilesWithReplacements', () => {
       app: { build: { ancestorBuilds: [timeoutBuild, workingBuild] } },
     });
 
-    // Simulate a timeout by making the first build's commit not exist
-    vi.fn(getChangedFiles).mockImplementation(() => {
-      throw new Error('timeout');
-    });
-
     const result = await getChangedFilesWithReplacement(context, {
       id: 'id',
       number: 3,
-      commit: 'missing',
+      commit: 'timeout',
       uncommittedHash: '',
       isLocalBuild: false,
     });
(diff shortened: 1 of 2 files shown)


node-src/git/getChangedFilesWithReplacement.ts
```diff
@@ -35,24 +35,46 @@ export async function getChangedFilesWithReplacement(
     const changedFiles = (await getChangedFiles(build.commit)) || [];
     return { changedFiles };
   } catch (err) {
+    console.log(err);
```
codykaup, line 38: Is this still needed?

node-src/git/getChangedFilesWithReplacement.ts
```diff
@@ -35,24 +35,46 @@ export async function getChangedFilesWithReplacement(
...
+    }
+
+    // Try to find a replacement build by checking multiple ancestor builds
+    let skip = 0;
+    const page = 10;
+    const limit = 80;
 
-      if (replacementBuild) {
+    while (skip < limit) {
+      const replacementBuild = await findAncestorBuildWithCommit(ctx, build.number, {
```
codykaup, line 53: Is this the function that's throwing?

ethriel3695:
https://github.com/chromaui/chromatic-cli/pull/1178/files/a1691a44b792b67614dbff169b4f494797869358#diff-47f22cd546abc461f23ebe20fe9fc05c8b82bfdb7b002772e73931054fae7e85R35

This is the attempt to bypass if the commit doesn't exist or times out and moves on to the next parent commit.

@codykaup

node-src/git/getChangedFilesWithReplacement.test.ts
```diff
@@ -125,4 +126,153 @@ describe('getChangedFilesWithReplacements', () => {
       })
     ).rejects.toThrow(/fatal: bad object missing/);
   });
+
+  it('tries a replacement builds until it finds a working one', async () => {
```
codykaup, line 130: These tests pass even when `node-src/git/getChangedFilesWithReplacement.ts` is on `main`. Is there a valid test that fails on `main` but passes on this branch?

node-src/git/getChangedFilesWithReplacement.ts
```diff
@@ -39,20 +39,41 @@ export async function getChangedFilesWithReplacement(
...
+
+      if (!replacementBuild) {
+        break;
+      }
+
+      try {
         ctx.log.debug(
           `Found replacement build for #${build.number}(${build.commit}): #${replacementBuild.number}(${replacementBuild.commit})`
         );
         const changedFiles = (await getChangedFiles(replacementBuild.commit)) || [];
```
codykaup, line 65: So if I understand correctly, this is the function that times out. Why would it time out? Would each ancestor build would have the same issue?

node-src/git/getChangedFilesWithReplacement.ts
```diff
@@ -39,20 +39,41 @@ export async function getChangedFilesWithReplacement(
...
     );
 
-    if (/(bad object|uncommitted changes)/.test(err.message)) {
-      const replacementBuild = await findAncestorBuildWithCommit(ctx, build.number);
+    if (!/(bad object|uncommitted changes)/.test(err.message)) {
+      throw err;
+    }
+
+    // Try to find a replacement build by checking multiple ancestor builds
+    let skip = 0;
```
codykaup, line 47: This loop is essentially what `findAncestorBuildWithCommit` does. Should `findAncestorBuildWithCommit` be responsible for returning a valid build? Again, it seems like it does but the `git` command fails. I'm just not sure trying another ancestor build would yield better results.

codykaup (commented):
Does the canary from this branch work for the case in question?

ethriel3695:
@jmhobbs 
This is the timeout getting triggered.
https://github.com/chromaui/chromatic-cli/blob/reuben/cap-3076-wealthsimple-is-reporting-issues-with-git-commit-not-found/node-src/git/execGit.ts#L11

I am not able to figure out why, though. I've run all sorts of combinations of `git --no-pager diff --name-only ${baseCommit} ${headCommit}`

I added a one-line change that might do something, but I'm unsure how to test it and was hoping you had an idea.

node-src/git/getChangedFilesWithReplacement.ts
```diff
@@ -39,7 +39,8 @@ export async function getChangedFilesWithReplacement(
       `Got error fetching commit for #${build.number}(${build.commit}): ${err.message}`
     );
 
-    if (/(bad object|uncommitted changes)/.test(err.message)) {
+    console.error(err.message);
```
jmhobbs, line 42: We'll want to pull this line before merge.

node-src/git/getChangedFilesWithReplacement.ts
```diff
@@ -39,7 +39,8 @@ export async function getChangedFilesWithReplacement(
...
+    if (/(bad object|uncommitted changes|timeout)/.test(err.message)) {
```
jmhobbs, line 43: We'd also need to trace the propagation, but it looks like `execa` sets `error.timedOut` on a timeout, so we could check that instead of string matching for `timeout`

https://github.com/sindresorhus/execa/blob/main/docs/termination.md#timeout

node-src/git/getChangedFilesWithReplacement.ts
ethriel3695: Done!

node-src/git/getChangedFilesWithReplacement.ts
ethriel3695: @jmhobbs, do you have recommendations on how I would do that?
Would it be something similar to this?
`if (/(bad object|uncommitted changes)/.test(err.message) || err.timedOut) {`

jmhobbs: What is the test for `bad object` and `uncommitted changes` looking for?  Have we seen issues with other errors from `git` as well as the time out?

ethriel3695: @jmhobbs The `bad object` is quite common when we can't find a commit and is the basis for our replacement build logic. The `uncommitted changes` are for the VTA.

I've never seen this timeout issue before.

ethriel3695:
This is not a viable fix, and is no longer needed.

PR #1181, opened 2025-05-21
Add git command logging and pass `--no-relative` to `git diff`.

tmeasday:
Ensure that we always `log.debug()` every git command we run -- this will make debugging weird git behaviour much easier.

Also pass `--no-relative` to git diff, in case the user has `diff.relative` set in their git config.

I can split these up into two PRs if that is preferred.


codykaup (approved):
Nice find!!

ghengeveld:
Reverting in #1183 because this should've been a major release.

PR #1188, opened 2025-06-06
Disable some deprecated flags

jmhobbs:
Cleans up some options we have had deprecated for a while.

|Flag|Reason|
|--|--|
|`--allow-console-errors`|No longer supported|
|`--app-code`|Moved to `--project-token`|
|`--diagnostics`|Moved to `--diagnostics-file`|
|`--only`|Moved to `--only-story-names`|


jmhobbs:
> Don't forget to update the docs!
> 
> https://www.chromatic.com/docs/configure/#deprecated-options

PR is already up!

PR #1189, opened 2025-06-11
Handle spaces in branch strings

jmhobbs:
In some detached HEAD situations it appears `git branch --show-current` may emit a string with spaces.  This PR is to add logs and test that assumption.


justin-thurman (approved):
Looks good. 🤞

codykaup (approved):

PR #1204, opened 2025-09-04
Handle framework names that specify full viewlayer packages

justin-thurman:
# Description

This PR fixes an issue in our logic mapping view layer packages and Storybook framework names that seems (as best I can tell, anyway) to be based on outdated `storybook build` output from SB6. See inline comment for more details.

# Manual QA

I created a project that uses a prebuilt storybook in its `chromatic` call, using React, Vite, and `storybook@10.0.0-beta.2`, and verified 1) that before this fix, we see the erroneous error message referencing the `--webpack-stats-json` flag, and 2) after this fix, we get the appropriate error message:
CleanShot 2025-09-15 at 14 49 36@2x

I also tested this on the customer's reproduction repo (it's private, so I can't link it), and confirmed the correct error message is shown.

<details>
  <summary>📦 Published PR as canary version: <code>13.1.5--canary.1204.17742984995.0</code></summary>
  <br />

  ```bash
  npm install chromatic@13.1.5--canary.1204.17742984995.0
  # or 
  yarn add chromatic@13.1.5--canary.1204.17742984995.0
  ```
</details>

### DIFF
--- a/node-src/lib/getPrebuiltStorybookMetadata.test.ts
+++ b/node-src/lib/getPrebuiltStorybookMetadata.test.ts
@@ -16,6 +16,19 @@ describe('getStorybookMetadataFromProjectJson', () => {
     });
   });
 
+  it('should return the metadata from a project with @storybook framework name', async () => {
+    const projectJsonPath = 'bin-src/__mocks__/atFrameworkNameProjectJson/project.json';
+    const metadata = await getStorybookMetadataFromProjectJson(projectJsonPath);
+
+    expect(metadata).toEqual({
+      version: '10.0.0-beta.2',
+      builder: {
+        name: '@storybook/builder-vite',
+        packageVersion: undefined,
+      },
+    });
+  });
+
   it('should return the metadata from a Storybook 6 project.json file', async () => {
     const projectJsonPath = 'bin-src/__mocks__/sb6ProjectJson/project.json';
     const metadata = await getStorybookMetadataFromProjectJson(projectJsonPath);
--- a/node-src/lib/getPrebuiltStorybookMetadata.ts
+++ b/node-src/lib/getPrebuiltStorybookMetadata.ts
@@ -33,7 +33,9 @@ export const getStorybookMetadataFromProjectJson = async (
 ): Promise<Partial<Context['storybook']>> => {
   const sbProjectJson = (await readFile(projectJsonPath)) as SBProjectJson;
   const viewLayerPackage = Object.keys(viewLayers).find(
-    (viewLayer) => viewLayers[viewLayer] === sbProjectJson.framework.name
+    (viewLayer) =>
+      viewLayers[viewLayer] === sbProjectJson.framework.name ||
+      viewLayer === sbProjectJson.framework.name
   );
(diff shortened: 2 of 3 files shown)


node-src/lib/getPrebuiltStorybookMetadata.ts
```diff
@@ -33,7 +33,9 @@ export const getStorybookMetadataFromProjectJson = async (
 ): Promise<Partial<Context['storybook']>> => {
   const sbProjectJson = (await readFile(projectJsonPath)) as SBProjectJson;
   const viewLayerPackage = Object.keys(viewLayers).find(
-    (viewLayer) => viewLayers[viewLayer] === sbProjectJson.framework.name
+    (viewLayer) =>
+      viewLayers[viewLayer] === sbProjectJson.framework.name ||
+      viewLayer === sbProjectJson.framework.name
```
justin-thurman, line 38: This change is a bit strange, but I think it's correct. From what I can tell, the existing usage of `viewLayers` is to map from SB6 framework names (e.g., `'react'`) to the actual package name (`'@storybook/react-vite'`, in this case). You can see an example of this in the `getPrebuiltStorybookMetadata.test.ts` tests around Storybook 6 projects. But it wasn't working properly for the framework name that we're getting in the customer's project (and my minimal reproduction). You can see in the test `project.json` that I added (which is a direct copy-paste of my minimal reproduction `project.json`) that we're getting a `framework.name` of `'@storybook/react-vite'`. In other words, we don't need the mapping layer from `viewLayers` and we can just compare against the key directly. That is, we use `viewLayer` instead of `viewLayers[viewLayer]`. According to the customer, this bug goes back to storybook 9 at least. Not sure when it popped up, but I'm guessing it's due to a change in how storybook build works sometime after SB6. I believe the reason this wasn't caught is that our existing unit test uses `'@storybook/react-webpack5'` as its framework, and for whatever reason, in the `viewLayers` mapping, `'@storybook/react-webpack5'` is an identity mapping. All other package names map to a more general name, e.g., `@storybook/vue -> vue`, etc. I have no idea why.

node-src/lib/getPrebuiltStorybookMetadata.test.ts
```diff
@@ -16,6 +16,19 @@ describe('getStorybookMetadataFromProjectJson', () => {
...
 
+  it('should return the metadata from a project with @storybook framework name', async () => {
+    const projectJsonPath = 'bin-src/__mocks__/atFrameworkNameProjectJson/project.json';
+    const metadata = await getStorybookMetadataFromProjectJson(projectJsonPath);
+
+    expect(metadata).toEqual({
+      version: '10.0.0-beta.2',
+      builder: {
+        name: '@storybook/builder-vite',
+        packageVersion: undefined,
```
justin-thurman, line 27: Note that this is a valid return based on `getStorybookMetadataFromProjectJson` return signature, and it matches the actual `project.json` from my reproduction project. It just doesn't have a separate builder dependency, and so has no version.

jwir3 (approved):
Thanks for the detailed explanation on the code changes - it really helped to speed up my review.

PR #1206, opened 2025-09-19
Updated the mapping for the storybook entry with a recent change in storybook-rsbuild

ethriel3695:
Recently, the dependency graph being built in the Storybook-rsbuild community package has changed, as has the dependency graph entry point.
https://github.com/rspack-contrib/storybook-rsbuild/issues/332

This results in users getting the Chromatic CLI error message `Did not find any CSF globs in **`.

This is the new format for the entry point:
```
{
      "id": "./storybook-config-entry.js",
      "name": "./storybook-config-entry.js + 1 modules",
      "modules": [
        { "name": "./storybook-config-entry.js" },
        { "name": "./storybook-stories.js" }
      ],
      "reasons": []
    },
```

To accommodate this, I've added the `storybook-config-entry.js` to the mapping in our getDependentStoryFiles.
I also added the file to the test to ensure it's passing.

I tested it out locally on the chromatic-cli repo with the stats file I have and I get the correct response now:
```
ℹ Traced 1 changed file to 1 affected story file:

— cdn/src/components/HostContent.stories.tsx + 3 modules [changed]
  ∟ [story index]

Set --trace-changed to 'expanded' to reveal underlying modules.
```

<details>
  <summary>📦 Published PR as canary version: <code>13.2.0--canary.1206.17862049123.0</code></summary>
  <br />

  ```bash
  npm install chromatic@13.2.0--canary.1206.17862049123.0
  # or 
  yarn add chromatic@13.2.0--canary.1206.17862049123.0
  ```
</details>

### DIFF
--- a/node-src/lib/turbosnap/getDependentStoryFiles.test.ts
+++ b/node-src/lib/turbosnap/getDependentStoryFiles.test.ts
@@ -170,6 +170,7 @@ describe('getDependentStoryFiles', () => {
     ['./node_modules/.cache/storybook-rsbuild-builder/storybook-stories.js'],
     [`./node_modules/.cache/storybook/storybook-rsbuild-builder/storybook-config-entry.js`],
     [`./node_modules/.cache/storybook-rsbuild-builder/storybook-config-entry.js`],
+    [`./storybook-config-entry.js`],
   ])('detects direct changes to CSF files, rspack (%s)', async (resolvedModule) => {
     const changedFiles = ['src/foo.stories.js'];
     const modules = [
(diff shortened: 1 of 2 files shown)


ethriel3695:
> This makes sense to me. I'm guessing the shape of this file hasn't changed, since you're able to test and it's working correctly? I'm not very familiar with rsbuild, so I just want to double check.

@justin-thurman you are correct. The overall structure of the file hasn't changed, only the entry point file. As you can see with the files above, it was previously in the rsbuilder cache in node_modules. We need to keep both for now for backwards compatibility.

PR #1212, opened 2025-10-01
Update storybook-addon.d.ts

jasikpark:
This adds the `disableSnapshot` parameter to the chromatic parameter types, which is referenced in chromatic docs: https://www.chromatic.com/docs/config-with-story-params/


storybook-addon.d.ts
jasikpark, line 2: BTW, I'm not sure this is actually effective? I definitely have to manually declare chromatic parameters in my own storybook interface augmentation.

PR #1217, opened 2025-10-23
Strip ' + N modules' when matching against untraced files

justin-thurman:
# Description

When tracing changes through imported files for purposes of TurboSnap, we sometimes call our `untrace` function with the `reasons` for a given file change, which takes the form of a string something like `src/foo.js + 2 modules`. This is the string that gets compared against user-provided untrace globs, e.g., `**/foo.js`. Currently, this doesn't work as expected because `**/foo.js` does not match `src/foo.js + 2 modules`. Users can work around this by passing `**/foo.js**` as the untraced value, but this PR enables the expected behavior by removing any ` + N module(s)` from the string that we compare against the values that the user has passed to the untraced flag.

# Manual QA

Reuben tested this, see details in Linear.

<details>
  <summary>📦 Published PR as canary version: <code>13.3.4--canary.1217.18974255621.0</code></summary>
  <br />

  ```bash
  npm install chromatic@13.3.4--canary.1217.18974255621.0
  # or 
  yarn add chromatic@13.3.4--canary.1217.18974255621.0
  ```
</details>

### DIFF
--- a/node-src/lib/turbosnap/getDependentStoryFiles.test.ts
+++ b/node-src/lib/turbosnap/getDependentStoryFiles.test.ts
@@ -708,37 +708,41 @@ describe('getDependentStoryFiles', () => {
     expect(result).toEqual({});
   });
 
-  it('ignores untraced files', async () => {
-    const changedFiles = ['src/utils.js'];
-    const modules = [
-      {
-        id: 1,
-        name: './src/utils.js', // changed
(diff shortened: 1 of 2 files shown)


ethriel3695 (approved):
@justin-thurman The change worked. I added the results in Linear since it's not a public project stats file.

jmhobbs (approved):

PR #1170, opened 2025-04-03
Update storybook to 8.6

kasperpeulen:

PR #1228, opened 2026-01-28
Update storybook to `10.x` & upgrade dependencies

ndelangen:
Potential replacement for: #1170

- I upgraded storybook to the latest
- But this forced me to upgrade vite and thus vitest
- I was forced to update the tsconfig due to the SB upgrade
- That forced me to update the eslint-typescript
- Doing that alone caused out-of-memory issues for eslint
- So I upgraded eslint to latest
- Then dealt with the side-effects of that inside our files.
- The webpack use and react was an issue, and so I migrated the internally used storybook to vite + html
- The local testing still needs react stories, so I added a second storybook config dir that has react framework. (but vite)
- storybook v10 no longer supports node18, so we can't run tests with node18 anymore

<details>
  <summary>📦 Published PR as canary version: <code>15.0.0--canary.1228.21595851672.0</code></summary>
  <br />

  ```bash
  npm install chromatic@15.0.0--canary.1228.21595851672.0
  # or 
  yarn add chromatic@15.0.0--canary.1228.21595851672.0
  ```
</details>


eslint.config.mjs
```diff
@@ -6,114 +7,132 @@ import prettier from 'eslint-plugin-prettier';
...
       'storybook-out/**',
       'coverage/**',
       'subdir/**',
     ],
   },
-  // eslint core + overrides
   eslint.configs.recommended,
   {
     rules: {
-      // check cyclomatic complexity
```
jwir3: I kind of like having these annotations on what each check is doing. I'll restore them in a separate commit.

PR #1249, opened 2026-03-17
feat: skip multiple Chromatic projects in a single CLI invocation

hamidrezahanafi:
## Summary
Adds the ability to skip multiple Chromatic projects in one CLI invocation by passing multiple `--project-token` values alongside `--skip`. Previously, skipping N projects required N separate CLI calls — now it can be done with a single command.
## Motivation
In monorepo setups with many Chromatic projects, skipping builds (e.g. on certain branches) required invoking the CLI once per project. This was slow, verbose, and hard to maintain in CI configurations. This change collapses all those invocations into one.
## How it works
When `--skip` is set and multiple `--project-token` values are provided, the CLI:
1. Gathers git info (commit, branch, slug) once.
2. For each token: authenticates and calls the `SkipBuild` mutation.
3. Reports how many projects were skipped and exits with code 0.
If some tokens fail, the CLI logs warnings but continues with the rest. It only throws an error if *all* tokens fail.
When a single token is provided, existing behavior is completely unchanged.
### CLI
```bash
npx chromatic \
  --project-token token-app-a \
  --project-token token-app-b \
  --project-token token-app-c \
  --skip

### DIFF
--- a/action-src/main.ts
+++ b/action-src/main.ts
@@ -124,7 +124,9 @@ async function run() {
     const outputDir = getInput('outputDir');
     const playwright = getInput('playwright');
     const preserveMissing = getInput('preserveMissing');
-    const projectToken = getInput('projectToken');
+    const projectTokenInput = getInput('projectToken');
+    const projectTokens = getMultilineInput('projectToken');
+    const projectToken = projectTokens.length > 1 ? projectTokens : projectTokenInput;
     const repositorySlug = getInput('repositorySlug');
     const shaInput = getInput('chromaticSha');
     const skip = getInput('skip');
--- a/action.yml
+++ b/action.yml
@@ -81,13 +81,13 @@ inputs:
     description: 'Deprecated, use onlyChanged, onlyStoryNames or onlyStoryFiles instead'
     required: false
   projectToken:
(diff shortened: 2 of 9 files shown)


codykaup:
Thanks for submitting this @hamidrezahanafi! We discussed this and decided we don't want to support batching `--skip` in this way and would prefer a single project token per run to avoid confusion (both as a user and a maintainer).

Are there some other improvements we should make to `--skip` to speed up the process?

hamidrezahanafi:
> Thanks for submitting this @hamidrezahanafi! We discussed this and decided we don't want to support batching `--skip` in this way and would prefer a single project token per run to avoid confusion (both as a user and a maintainer).
> 
> Are there some other improvements we should make to `--skip` to speed up the process?

@codykaup In our monorepo we have to skip all or most of the projects ( we have 28 of them ) on most PRs since the effected changes are in single package in the monorepo and whenever we run the skips in parallel, we get sigkill because your CLI use a lots of memory, so now we do the skips in serial which takes a long time to run. I tried to reduce the mem usage of the CLI on another PR and that one also a no go from other reviews. will try to do a patch on the CLI and call it a day. Thanks for looking into this :)

PR #1274, opened 2026-04-07
React Native build command

jmhobbs:
Experimental entrypoint for building React Native storybooks.

`npx chromatic react-native-build`

Screenshot 2026-04-07 at 5 06 41 PM

<details>
  <summary>📦 Published PR as canary version: <code>16.4.1--canary.1274.24731837513.0</code></summary>
  <br />

  ```bash
  npm install chromatic@16.4.1--canary.1274.24731837513.0
  # or 
  yarn add chromatic@16.4.1--canary.1274.24731837513.0
  ```
</details>

### DIFF
--- a/bin-src/reactNativeBuild.ts
+++ b/bin-src/reactNativeBuild.ts
@@ -5,14 +5,14 @@ import meow from 'meow';
 import os from 'os';
 import path from 'path';
 
+
 interface ExpoConfig {
   platforms?: string[];
   scheme?: string;
 }
 
 /**
- * Run a shell command, streaming output to the terminal and teeing it into a log file.
- * Returns the log file path on failure for user reference.
+ * Run a shell command, streaming output to the terminal.
  *
  * @param command
  * @param args
@@ -24,21 +24,14 @@ async function runBuildCommand(
   args: string[],
   options: { cwd?: string } = {}
 ): Promise<void> {
-  const logFile = path.join(fs.mkdtempSync(path.join(os.tmpdir(), 'chromatic-rn-')), 'build.log');
-  const logStream = fs.createWriteStream(logFile);
-
   try {
-    const subprocess = execa(command, args, {
+    await execa(command, args, {
       cwd: options.cwd,
-      stdout: ['inherit', logStream],
-      stderr: ['inherit', logStream],
+      stdout: 'inherit',
+      stderr: 'inherit',
     });
-
-    await subprocess;
   } catch {
-    throw new Error(`Build command failed: ${command} ${args.join(' ')}\nSee log: ${logFile}`);
-  } finally {
-    logStream.close();
+    throw new Error(`Build command failed: ${command} ${args.join(' ')}`);
   }
 }
 
--- a/bin-src/reactNativeBuild.ts
+++ b/bin-src/reactNativeBuild.ts
@@ -227,7 +227,20 @@ Currently, only builds using Expo are supported.`,
     process.exit(1);
   }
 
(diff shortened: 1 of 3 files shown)


jmhobbs:
Added a `--platform` flag instead of splitting into two scripts.  By default it will build whatever native platforms expo has in it's config.

```
 npx chromatic@16.2.0--canary.1274.24158018702.0 react-native-build --help

  Build React Native Storybook for Chromatic

  Usage
    $ chromatic react-native-build [options]

  Options
    --platform  Platform to build (android, ios). Can be specified multiple times. Defaults to all platforms in Expo config.
```

Screenshot 2026-04-08 at 4 56 56 PM

bin-src/reactNativeBuild.ts
```diff
@@ -0,0 +1,344 @@
...
+  );
+
+  try {
+    info(`xcodebuild ${xcodebuildArguments.join(' ')}`, 'Build iOS');
+    await execa('xcodebuild', xcodebuildArguments, { cwd: path.resolve('ios') });
+    if (!fs.existsSync(appPath)) {
+      throw new Error(`Expected .app bundle not found at ${appPath}`);
+    }
+  } finally {
+    fs.rmSync(derivedDataPath, { recursive: true, force: true });
```
codykaup: Maybe I'm reading this wrong but won't this delete the build artifacts before we have a chance to use them?

codykaup (commented):
Although I like the UI of this, it feels really different than what we already have. Is that okay? Do we make it match?

bin-src/reactNativeBuild.ts
jmhobbs: No, you're right.  That slipped in when I broke up the `main` function and reordered things to guarantee cleanup.  I'll get that fixed.

bin-src/reactNativeBuild.ts
```diff
@@ -0,0 +1,344 @@
...
+    '-workspace',
+    `${scheme}.xcworkspace`,
+    '-scheme',
+    scheme,
+    '-configuration',
+    'Release',
+    '-sdk',
+    'iphonesimulator',
+    '-derivedDataPath',
+    derivedDataPath,
```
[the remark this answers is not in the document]
jmhobbs: https://linear.app/chromaui/issue/CAP-4311/find-a-way-to-use-default-expo-deriveddata-path-for-ios-builds

jmhobbs:
> Although I like the UI of this, it feels really different than what we already have. Is that okay? Do we make it match?

@codykaup I figured a non-interactive one would be better since we will swallow and log the output to a file in a normal run (like we do with build-storybook). We can always integrate it more like our existing workflow.

codykaup:
> > Although I like the UI of this, it feels really different than what we already have. Is that okay? Do we make it match?
> 
> @codykaup I figured a non-interactive one would be better since we will swallow and log the output to a file in a normal run (like we do with build-storybook). We can always integrate it more like our existing workflow.

Oh, I meant the boxes that highlight the commands that are running which look great. It just looks very different from the rest of the CLI. 😂

bin-src/reactNativeBuild.ts
jmhobbs: Should be fixed!

jmhobbs:
Made it fit in a little better with our existing UI;

Screenshot 2026-04-13 at 3 56 07 PM

codykaup (approved):
🚀

jmhobbs:
Rebased off `main`, new UI since I didn't post it yet.  It's just the headings since build tool output is interleaved.

Screenshot 2026-04-13 at 3 56 07 PM

PR #1283, opened 2026-04-14
[WFL-208] Extract `refs` field from Storybook and pass when announcing build

ax-vasquez:
## Summary

Extract the `refs` configuration when announcing a build and pass it to Chromatic.

This is necessary for supporting composed Storybooks' MCP servers hosted on Chromatic.

<details>
  <summary>📦 Published PR as canary version: <code>16.6.0--canary.1283.24742543830.0</code></summary>
  <br />

  ```bash
  npm install chromatic@16.6.0--canary.1283.24742543830.0
  # or 
  yarn add chromatic@16.6.0--canary.1283.24742543830.0
  ```
</details>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,7 +27,3 @@ storybook-out
 coverage
 package.json.backup
 tests-file.xml
-
-# Compound Engineering plugin artifacts - not needed in repo
-docs/plans
-docs/brainstorms
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,7 @@ storybook-out
 coverage
 package.json.backup
 tests-file.xml
+
+# Compound Engineering plugin artifacts - not needed in repo
+docs/plans
+docs/brainstorms
--- a/node-src/lib/getStorybookMetadata.ts
+++ b/node-src/lib/getStorybookMetadata.ts
@@ -166,7 +166,10 @@ export const findBuilder = async (mainConfig, v7) => {
   ]);
 };
 
-export const findReferences = async (mainConfig, v7) => {
+// Only used by Chromatic - surfaces Storybook refs and is used when announcing a build.
+// The refs consumed by the MCP Addon for hosted Storybooks with composition on Chromatic.
+const findReferences = async (mainConfig, v7) => {
+  // The MCP Addon was first added within version 9; there is no need to check for older versions
   if (!mainConfig || !v7) {
     return {};
   }
--- a/node-src/lib/getStorybookMetadata.ts
+++ b/node-src/lib/getStorybookMetadata.ts
@@ -166,7 +166,6 @@ export const findBuilder = async (mainConfig, v7) => {
   ]);
 };
 
-// TODO: Remove this when we start tracking refs within the project.json file (will be redundant)
 // Only used by Chromatic - surfaces Storybook refs and is used when announcing a build.
 // The refs are consumed by the MCP Addon for hosted Storybooks with composition on Chromatic.
 const findReferences = async (mainConfig, v7) => {
--- a/node-src/lib/getStorybookMetadata.ts
+++ b/node-src/lib/getStorybookMetadata.ts
@@ -166,13 +166,13 @@ export const findBuilder = async (mainConfig, v7) => {
   ]);
 };
 
-export const findRefs = async (mainConfig, v7) => {
+export const findReferences = async (mainConfig, v7) => {
   if (!mainConfig || !v7) {
     return {};
   }
 
-  const refs = mainConfig.getSafeFieldValue(['refs']);
-  return refs ? { refs } : {};
+  const references = mainConfig.getSafeFieldValue(['refs']);
+  return references ? { refs: references } : {};
 };
 
 export const findStorybookConfigFile = async (ctx: Context, pattern: RegExp) => {
@@ -212,7 +212,7 @@ export const getStorybookMetadata = async (ctx: Context) => {
     findConfigFlags(ctx),
     findStorybookVersion(ctx),
     findBuilder(mainConfig, v7),
-    findRefs(mainConfig, v7),
+    findReferences(mainConfig, v7),
   ]);
 
   ctx.log.debug(info);
--- a/node-src/lib/getStorybookMetadata.ts
+++ b/node-src/lib/getStorybookMetadata.ts
@@ -166,6 +166,15 @@ export const findBuilder = async (mainConfig, v7) => {
   ]);
 };
 
+export const findRefs = async (mainConfig, v7) => {
+  if (!mainConfig || !v7) {
+    return {};
+  }
+
+  const refs = mainConfig.getSafeFieldValue(['refs']);
+  return refs ? { refs } : {};
+};
+
 export const findStorybookConfigFile = async (ctx: Context, pattern: RegExp) => {
   const configDirectory = ctx.options.storybookConfigDir ?? '.storybook';
   const files = await readdir(configDirectory);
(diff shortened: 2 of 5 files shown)


justin-thurman:
@ax-vasquez, can you rebase onto main? We just merged a PR swapping from `tsup` to `tsdown`, and I think we'd have more confidence testing this with that swap in.

ax-vasquez:
@justin-thurman Done!

node-src/lib/getStorybookMetadata.ts
```diff
@@ -166,6 +166,15 @@ export const findBuilder = async (mainConfig, v7) => {
   ]);
 };
 
+export const findReferences = async (mainConfig, v7) => {
```
justin-thurman: Does this need to be exported? I'm not seeing it used anywhere else.

node-src/lib/getStorybookMetadata.ts
```diff
@@ -166,6 +166,15 @@ export const findBuilder = async (mainConfig, v7) => {
...
+  if (!mainConfig || !v7) {
```
justin-thurman, line 174: Why are we not supporting v6 storybooks here? Something in SB preventing us?

node-src/tasks/initialize.test.ts
```diff
@@ -141,6 +141,7 @@ describe('announceBuild', () => {
           packageVersion: ctx.pkg.version,
           rebuildForBuildId: undefined,
           storybookAddons: ctx.storybook.addons,
+          storybookRefs: ctx.storybook.refs,
```
justin-thurman, line 151: This test feels a bit flimsy. Since there are no refs on the default context, this is basically asserting `undefined === undefined`. Can we populate some refs to verify plumbing works as expected?

node-src/types.ts
```diff
@@ -250,6 +250,7 @@ export interface Context {
       packageVersion?: string;
     };
     mainConfigFilePath?: string;
+    refs?: Record<string, unknown>;
```
justin-thurman: Can we tighten the schema at all here? I'd imagine the index knows more about the shape of the data here than `unknown`?

.gitignore
```diff
@@ -27,3 +27,7 @@ storybook-out
 coverage
 package.json.backup
 tests-file.xml
+
+# Compound Engineering plugin artifacts - not needed in repo
```
justin-thurman: I'm not sure we should clutter up our `.gitignore` with this. Can you just do this for your local setup?

justin-thurman (commented):
This seems like it works, and I tested the canary, no problems there. But I have a couple questions, and I feel like there are a few tweaks we could do to clean this up a bit.

.gitignore
codykaup: Justin beat me to it! For tools that are more specific to an individual's setup, I recommend using a global ignore. 

https://stackoverflow.com/a/7335487

node-src/lib/getStorybookMetadata.ts
ax-vasquez: This feature is intended only for the MCP Addon which was added at some point in version 9 - I don't recall the exact version, but it seemed like we didn't need to bother with handling the logic for older versions.

In short, no reason not to, but we also don't have a specific need for this file on anything below the versions that support the MCP Addon.

node-src/lib/getStorybookMetadata.ts
ax-vasquez: Ah my bad! This was a holdover from an earlier implementation I had been testing with. I'll fix this.

node-src/types.ts
ax-vasquez: Oh my bad, I thought I had gotten this already but I think I got sidetracked. Will fix this!

.gitignore
ax-vasquez: Got it, will do!

node-src/lib/getStorybookMetadata.ts
ax-vasquez: I could also add a check that is more specific for the version that supports the Addon MCP, but I had been trying to reuse what was there in my first pass. Think it's worth me adding that?

.gitignore
justin-thurman: Can also do, IIRC, `.git/info/excludes` or something like that.

node-src/lib/getStorybookMetadata.ts
justin-thurman: Gotcha. Can you add a comment explaining the "why" here?

justin-thurman: > I could also add a check that is more specific for the version that supports the Addon MCP, but I had been trying to reuse what was there in my first pass. Think it's worth me adding that?

I think keeping this simple is better than adding a check.

node-src/tasks/initialize.test.ts
ax-vasquez: 🤦 I just outright overlooked this... I'll make a change

node-src/lib/getStorybookMetadata.ts
ax-vasquez: Added a comment here! :)

node-src/tasks/initialize.test.ts
ax-vasquez: I've added a basic `refs` config to the test data - lemme know if you think it'd be worth adding some configs in different states!

justin-thurman: This is good enough, imo. Thanks!

justin-thurman (approved):
Looks good to me! Thanks for the tweaks.

node-src/lib/getStorybookMetadata.ts
```diff
@@ -166,6 +166,19 @@ export const findBuilder = async (mainConfig, v7) => {
   ]);
 };
 
+// TODO: Remove this when we start tracking refs within the project.json file (will be redundant)
```
codykaup: Is Storybook planning to add `refs` to the output of `project.json`?

node-src/types.ts
```diff
@@ -150,6 +150,18 @@ export interface Options extends Configuration {
...
 }
 
+interface StorybookReferenceConfig {
+  title: string;
+  url: string;
+  expanded?: boolean;
+  sourceUrl?: string;
+}
+
+type StorybookReference =
```
codykaup, line 160: Does Storybook export this type anywhere so that we could import it?

node-src/lib/getStorybookMetadata.ts
ax-vasquez: Yeah, but it's tentative so I have no idea when it'll actually happen. I just wanted to call out that this code won't be necessary when that happens.

node-src/types.ts
ax-vasquez: Okay, it doesn't look like, it unfortunately. There is a type, but it's not exported and also isn't the full possibility of what the docs state this value can be:
* https://storybook.js.org/docs/api/main-config/main-config-refs

node-src/lib/getStorybookMetadata.ts
codykaup: We'll have to support older versions of Storybook so instead of this TODO comment, do we check the existence of `refs` in `project.json` and:

- If it exists, use it! Skip parsing ourselves.
- If it doesn't exist, parse it ourselves (what this PR does).

Then it'll be valid even if we do make that change in the future.

node-src/types.ts
codykaup: Bummer! Thanks for checking.

ax-vasquez:
Got the approval, but holding off on merging after speaking with Cody - since it's EOD on a Friday, we want to wait to merge until Monday.

codykaup:
Adding "DO NOT MERGE" until we add support on the backed!

ax-vasquez:
Just an update! The backend changes have been merged and tested 👍

PR #1297, opened 2026-04-27
Add React Native build stage

jmhobbs:
This PR adds steps to the build task which can build the Android and/or iOS Storybook for React Native.

Additionally, it replaces the `scheme` usage from Expo for `name`, which matches what the actual tool does to determine file and scheme names for iOS.

Apologies for the size of this one, a lot of things moved around and conditionally showing UI steps in the build task added some overhead.

## QA Steps

- Run a web build
- Run a web build with `-d`
- Run an RN build
- Run an RN build with `-d`
- Run an RN build with `-d` but no `manifest.json`
- Run an RN build with [custom build commands](https://www.chromatic.com/docs/configure/#reactnative-iosbuildcommand

## Canary

<details>
  <summary>📦 Published PR as canary version: <code>16.8.0--canary.1297.25332153280.0</code></summary>
  <br />

  ```bash
  npm install chromatic@16.8.0--canary.1297.25332153280.0
  # or 
  yarn add chromatic@16.8.0--canary.1297.25332153280.0
  ```
</details>


bin-src/reactNativeBuild.test.ts
```diff
@@ -119,25 +123,13 @@ describe('react-native-build', () => {
     );
   });
 
-  it('exits with error when ios platform has no scheme', async () => {
```
jmhobbs, line 122: `name` is required for all expo apps, so we don't need to test and fail here anymore.

bin-src/reactNativeBuild.ts
```diff
@@ -1,164 +1,22 @@
...
+const platformNames: Record<string, string> = {
+  android: 'Android',
+  ios: 'iOS',
+};
 
 function error(message: string, title?: string) {
   console.error(chalk.bold.red('✖') + chalk.bold(title || 'Error') + '\n' + message);
 }
 
-// execa wrapper for convenience
```
jmhobbs, line 21: Most of this file moved into `node-src/lib/react-native/(build|expoConfig)`

node-src/tasks/build.ts
```diff
@@ -234,16 +260,96 @@ export const buildStorybook = async (ctx: Context) => {
...
+};
+
+const resolvePlatforms = (ctx: Context) => {
+  return (ctx.announcedBuild?.browsers ?? []).filter(
+    (b): b is 'ios' | 'android' => b === 'ios' || b === 'android'
+  );
+};
+
+/* eslint-disable-next-line complexity, max-statements */
+export const buildReactNativeArtifacts = async (ctx: Context, task: Task) => {
```
jmhobbs: This is a big function, but I didn't want to break it up much more than this.  Willing to change it if anyone disagrees, but I think it reads well and works as a single unit.

node-src/tasks/build.ts
```diff
@@ -234,16 +260,96 @@ export const buildStorybook = async (ctx: Context) => {
...
+      } else {
+        const config = await readExpoConfig();
+        const { artifactPath } = await buildIos(config.name, logStream);
+        renameSync(artifactPath, path.join(ctx.sourceDir, 'storybook.app'));
+      }
+    }
+  } catch (buildError) {
+    setExitCode(ctx, exitCodes.NPM_BUILD_STORYBOOK_FAILED, true);
+    await new Promise<void>((resolve) => logStream.end(resolve));
+    const tail = await readLastLines(ctx.reactNativeBuildLogFile, MAX_REACT_NATIVE_LOG_LINES);
```
jmhobbs: Unlike a web storybook build, I'm not emitting the whole log here.  RN builds are _very_ verbose, so a whole log will flood the terminal.

node-src/types.ts
```diff
@@ -364,6 +364,7 @@ export interface Context {
   sourceDir: string;
   buildCommand?: string;
   buildLogFile?: string;
+  reactNativeBuildLogFile?: string;
```
jmhobbs, line 367: Sorry.

node-src/lib/react-native/expoConfig.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Sanitize a project name for use as an iOS xcodebuild scheme and workspace name.
+ * Replicates the behavior of `@expo/config-plugins` sanitizedName from
+ * `@expo/config-plugins/build/ios/utils/Xcodeproj.js`.
+ *
+ * @param name The raw project name from the Expo config.
+ *
+ * @returns The sanitized name safe for use as an xcodebuild scheme.
+ */
+export function sanitizedName(name: string): string {
```
jmhobbs: I inlined this since it is internals on that Expo package.  I can import instead since it will bake in, we would just need to keep an eye on that package for refactors moving it.

node-src/ui/tasks/build.ts
```diff
@@ -9,31 +7,41 @@ export const initial = (ctx: Context) => ({
...
+  output: ctx.options?.reactNative?.iosBuildCommand
+    ? `Running command: ${ctx.options.reactNative.iosBuildCommand}`
+    : 'Building iOS',
+});
+
+export const success = (ctx: Context) => ({
+  status: 'success',
+  title: `${capitalize(buildType(ctx))} built in ${getDuration(ctx)}`,
+  output: ctx.isReactNativeApp
+    ? `View build log at ${ctx.reactNativeBuildLogFile}`
```
justin-thurman: I think there's a gap here, due to the early return from `buildReactNativeArtifacts`:
```ts
  // we skip this build step if not React Native, or we have prebuilt assets.
  if (!ctx.isReactNativeApp || !!ctx.options.storybookBuildDir) return;
```
It looks like we could have `ctx.isReactNativeApp === true`, but if `storybookBuildDir` is set, then `reactNativeBuildLogFile === undefined`. Am I reading that right?

node-src/tasks/build.ts
```diff
@@ -234,16 +260,96 @@ export const buildStorybook = async (ctx: Context) => {
...
+    ctx.log.error(reactNativeBuildFailed(ctx, buildError, tail));
+    throw new Error(failed(ctx).output);
+  } finally {
+    logStream.end();
```
justin-thurman: Need to await this like you do in the `catch` path?

node-src/tasks/build.ts
```diff
@@ -257,15 +363,8 @@ export default function main(ctx: Context) {
...
-      if (ctx.isReactNativeApp) {
-        if (!ctx.options.storybookBuildDir) {
-          ctx.log.error(missingStorybookBuildDirectory(ctx.announcedBuild?.browsers));
-          setExitCode(ctx, exitCodes.INVALID_OPTIONS, true);
-          throw new Error(missingBuildDirectoryForReactNative(ctx).output);
-        }
-
+      if (ctx.isReactNativeApp && ctx.options.storybookBuildDir) {
         ctx.sourceDir = ctx.options.storybookBuildDir;
-        ctx.options.outputDir = ctx.options.storybookBuildDir;
```
justin-thurman, line 268: What's up with this removal? Isn't this value referenced in downstream tasks? Is this safe?

node-src/tasks/build.ts
```diff
@@ -234,16 +260,96 @@ export const buildStorybook = async (ctx: Context) => {
...
+/* eslint-disable-next-line complexity, max-statements */
+export const buildReactNativeArtifacts = async (ctx: Context, task: Task) => {
+  // we skip this build step if not React Native, or we have prebuilt assets.
+  if (!ctx.isReactNativeApp || !!ctx.options.storybookBuildDir) return;
+
+  const platforms = resolvePlatforms(ctx);
+  const needsAndroid = platforms.includes('android');
+  const needsIos = platforms.includes('ios');
+
+  if (!needsAndroid && !needsIos) return;
```
justin-thurman: Seems like a potential point of confusion here. If user tries a RN build but has a misconfigured app on Chromatic that doesn't have any RN-compatible browsers enabled, this will just silently return, right? Worth adding a log or even making this an error state? I lean towards the latter: if I'm trying to run a RN build on an app that doesn't support it, I'd expect that to fail loudly.

node-src/tasks/build.ts
```diff
@@ -10,27 +18,49 @@ import { buildBinName as e2eBuildBinName, getE2EBuildCommand } from '../lib/e2e'
...
+      buffer.push(line);
+      if (buffer.length > lineCount) buffer.shift();
+    });
+    rl.on('close', () => resolve(buffer.join('\n')));
+    rl.on('error', reject);
+  });
+}
+
 export const setSourceDirectory = async (ctx: Context) => {
+  // if it has been set upstream of us, do not overwrite it
```
justin-thurman: Worth clarifying what "upstream of us" refers to? Where exactly would this be set?

node-src/lib/react-native/build.test.ts
```diff
@@ -0,0 +1,146 @@
...
+    } finally {
+      Object.defineProperty(process, 'platform', {
+        value: originalPlatform,
+        configurable: true,
+      });
+    }
+  });
+
+  it('calls expo prebuild and xcodebuild with correct args on macOS', async () => {
+    if (process.platform !== 'darwin') return;
```
justin-thurman: Doesn't this mean these tests don't run in CI? Can we mock platform like you do for the `throws when not on macOS` test, or can these tests actually not be run on Linux?

node-src/tasks/build.test.ts
```diff
@@ -559,6 +584,149 @@ describe('buildStorybook analytics', () => {
...
+      isReactNativeApp: true,
+      announcedBuild: { browsers: ['android'] },
+      log: new TestLogger(),
+      sourceDir: '/path/to/build',
+    } as any;
+    await buildReactNativeArtifacts(ctx, task);
+    expect(mockLogStream.end).toHaveBeenCalled();
+  });
+
+  it('closes the log stream and includes a truncated tail on build error', async () => {
```
justin-thurman: This doesn't actually test that the truncated tail shows up on the error. Would it be simple enough to add an assertion, or are we comfortable leaving it off?

bin-src/reactNativeBuild.test.ts
```diff
@@ -119,42 +138,75 @@ describe('react-native-build', () => {
...
-    const originalPlatform = process.platform;
-    Object.defineProperty(process, 'platform', { value: 'linux' });
+  it('creates output-dir if it does not exist when --output-dir is specified', async () => {
+    mockedExeca.mockImplementation((command: any, args: any) => {
+      if (command === 'npx' && args?.[0] === 'expo' && args?.[1] === 'config') {
+        return { stdout: JSON.stringify({ platforms: ['android'], name: 'MyApp' }) } as any;
+      }
+      return Promise.resolve({}) as any;
+    });
+    mockedExistsSync.mockReturnValue(true);
```
justin-thurman, line 167: This one is confusing. Given the test title, shouldn't this be `false`?

node-src/lib/react-native/expoConfig.ts
```diff
@@ -0,0 +1,24 @@
...
+    return JSON.parse(result.stdout) as ExpoConfig;
+  } catch {
+    throw new Error(
+      'Failed to read Expo config. Ensure Expo is installed and you are in an Expo project directory.'
+    );
+  }
+}
+
+// Re-export the sanitizedName function from expo/config-plugins to ensure iOS filename consistency.
+export const sanitizedName = IOSConfig.XcodeUtils.sanitizedName;
```
justin-thurman: It looks like we're adding a whole lot of dependencies (whatever comes with `@expo/config-plugins`, just to get a sanitization function. I haven't looked at the function itself, but based on the name, I'm guessing it's not large. Could we just inline this or copy its logic or something? Not mandatory, obviously, just a thought.

justin-thurman (commented):
Looking good! I have a few smaller inline questions, and one larger thought. In the `build` task, there's a lot of `if RN` branching threaded throughout. The task steps are basically a list of functions, only half of which we might expect to run on any given build (whether it's RN or web). Would it be simpler to create an entirely new RN-based build task? Then we could either 1) skip the entire build/build-RN task, rather than skipping individual steps, or 2) conditionally construct the task list at the very beginning of the run to include RN-build for RN or regular web build for web. What do you think?

node-src/tasks/build.ts
jmhobbs: I believe this line was added as amistake originally. `ctx.options` should be immutable input flags, and that specific one should drive the value of `ctx.sourceDir` when set (in `setSourceDirectory`). I'm not aware of any downstream affects of unsetting this here, we don't set it explicitly for web builds (that I could identify)

node-src/tasks/build.ts
jmhobbs: In this case, the skip block for this task.  I will clarify the comment.

node-src/tasks/build.ts
jmhobbs: Valid point.  When I merged the two build steps and their guard functions I also merged the logic and it came out a bit odd here.  An explicit error does make sense, I will do that.

node-src/tasks/build.ts
jmhobbs: Yes, will do.

node-src/lib/react-native/expoConfig.ts
jmhobbs: That is what I _had_ originally, but then I reversed course because I wanted to make sure we were more explicit about it; https://github.com/chromaui/chromatic-cli/pull/1297/changes/ae84f4e990bd91e98602d88954bbf11b4f54b77e

Most of it _should_ get shaken out during the build, as it is a dev dependency.  That said I'm willing to go back to that, embedding was my original plan.

justin-thurman: Ah, right, I always forget about the dependencies being shaken out. This is probably fine then. (I think? I'm no JS build expert. 😅)

node-src/ui/tasks/build.ts
jmhobbs: This should be resolved with the new split tasks and UI.

bin-src/reactNativeBuild.test.ts
jmhobbs: We unconditionally call `mkdirSync(outputDir, { recursive: true });`, instead of checking existence, so this doesn't really affect things.  I can remove it entirely actually (and will).

jmhobbs:
Updating to a separate task is good for separation, but currently both items show in the task list regardless of project type.  Looking into fixing this.

Screenshot 2026-04-30 at 3 10 47 PM

jmhobbs:
Ready for re-review!

node-src/lib/react-native/build.ts
```diff
@@ -0,0 +1,138 @@
...
+    await execa('xcodebuild', xcodebuildArguments, {
+      cwd: path.resolve('ios'),
+      stdout: logStream,
+      stderr: logStream,
+    });
+    if (!existsSync(appPath)) {
+      throw new Error(`Expected .app bundle not found at ${appPath}`);
+    }
+
+    const artifactDirectory = mkdtempSync(path.join(os.tmpdir(), 'chromatic-rn-ios-artifact-'));
```
justin-thurman: Looks like this never gets cleaned up in `finally`. I know it's a tmp directory, but worth cleaning? Not blocking.

node-src/lib/react-native/build.ts
```diff
@@ -0,0 +1,138 @@
...
+ * @returns The path to the built APK file and duration in seconds.
+ */
+export async function buildAndroid(logStream: WriteStream) {
+  const start = new Date();
+
+  logStream.write('\n[chromatic] Android build: npx expo prebuild --platform android\n');
+  await exec('npx', ['expo', 'prebuild', '--platform', 'android'], {}, logStream);
+
+  logStream.write('\n[chromatic] Android build: ./gradlew assembleRelease\n');
+  await exec('./gradlew', ['assembleRelease'], { cwd: path.resolve('android') }, logStream);
```
justin-thurman, line 74: Do we need to resolve against `ctx.sourceDir` here? I'm not sure one way or the other, just a thought.

node-src/tasks/buildReactNative.ts
```diff
@@ -0,0 +1,117 @@
...
+  ctx.log.debug('Running React Native build command:', command);
+  const label = platform === 'android' ? 'Android' : 'iOS';
+  logStream.write(`\n[chromatic] ${label} build: ${command}\n`);
+  try {
+    await runCommand(command, {
+      stdout: logStream,
+      stderr: logStream,
+    });
+  } catch (err) {
+    throw new Error(`React Native build command failed: ${command}\n${err.message}`);
```
justin-thurman, line 51: Any chance `runCommand` rejects with a non-Error, thus making `err.message` undefined?

justin-thurman (approved):
This looks good. That conditional task logic is little gnarly, but there might just not be a way around it. Though we _could_ do the swap in `node-src/tasks/index.ts`, where we build the list of tasks to be run in the first place. We have access to `ctx` there, so we could just put in `build` for web and `buildRN` or whatever for RN. Not blocking, just a thought.

A few other minor questions. I also haven't had a chance to QA yet (I've never gotten set up to run RN). Hopefully can get it today, but not certain.

jmhobbs:
> Though we could do the swap in node-src/tasks/index.ts, where we build the list of tasks to be run in the first place. We have access to ctx there, so we could just put in build for web and buildRN or whatever for RN. Not blocking, just a thought.

The issue with that is we don't have the `isReactNativeApp` and `browsers` until the `initialize` step, and then it's too late to change the task list.  I considered a multiple phases thing with Listr but it was a big cut and felt worse to me than the conditionals.

justin-thurman:
> The issue with that is we don't have the isReactNativeApp and browsers until the initialize step, and then it's too late to change the task list. I considered a multiple phases thing with Listr but it was a big cut and felt worse to me than the conditionals.

Oh, yeah, that makes sense. Yeah, this works then.

node-src/lib/react-native/build.ts
jmhobbs: `ctx.sourceDir` is the _output_ directory (the _source_ of the built storybook for the CLI to consume), not the input source, so this shouldn't be related.  We copy the output from `android/..../app-release.apk` into `ctx.sourceDir` after build (up a level).

There may be some path mangling we will need to do with unusual setups, but we have the build command escape hatch for now and we can adjust as reports come in.

node-src/lib/react-native/build.ts
jmhobbs: Tough call. We have this temp dir so that we can clean up the much larger `derivedDataPath` without losing our built artifact.  That way `artifactPath` is still present after the return, which a delete in `finally` would not allow.

I think we have a couple options, we can add comments to make it the responsibility of the caller to cleanup `artifactDirectory` OR we can require the caller to send in the `artifactDirectory` and make that ownership clear.  The latter is probably the better option.

node-src/tasks/buildReactNative.ts
jmhobbs: There are no explicit non-error rejections in our code around `runCommand`. It would have to come from `execa`, which would be a bug on their side.

PR #1302, opened 2026-04-29
Scaffolding for removing `Context` from CLI task bodies

justin-thurman:
# Description

## The Problem

This PR establishes a pattern for an incremental migration away from the `Context` god object that's currently threaded throughout the CLI code. Current, ctx accumulates user input configuration, shared dependencies (log, client, http, env, etc.), and progressively-filled-in output of every pipeline phase (i.e., the `gitInfo` task populates `ctx.git`, the `storybookInfo` task populates `ctx.storybook`, etc.). Because this is a single mutable bag of properties, callers can't tell via the type system what values are set at any given point (example pain point: you're in the `storybookInfo` task. Is `ctx.git` set? Who knows? Go look at pipeline ordering, verify that `gitInfo` runs before `storybookInfo`, and then read `gitInfo` code to be sure it sets the properties you need). Another pain point is that it's impossible to tell what a task's dependencies actually are. Example: the `auth` task accepts `ctx`. Of the literally hundreds of fields on `ctx`, what does `auth` actually depend on? (Answer: `options`, `client`, and `env`.)

It would be nice to just rip out `Context` altogether, but that has some problems. First, it would be a massive undertaking. It's threaded throughout the entire codebase, so we'd have to rethink everything, from our entrypoints, down through every function, including our UI rendering logic. This is likely not feasible on any reasonable timescale.

Additionally, `Context` is publicly exported, as are programmatic entrypoints to run a Chromatic build that take `Context` as their argument. So any backwards incompatible change would be a major version bump. We should probably explore a deprecation timeline before doing anything so drastic.

## A Potential Solution

This PR attempts to address the pain points I raised above without ripping out `Context` altogether. There are two broad goals. First, phase-specific type safety. Each task in the pipeline produces a set type. Example: the `storybookInfo` task produces a `Storybook` type, so you know exactly what data is coming out of a given task. Tasks also have narrowly defined parameters. Instead of taking the entire `Context`, I propose to define two types per task: input and dependencies. Input represents the data (either from the initial run config or previous tasks) that the task needs, and dependencies likewise represent the minimal dependencies required by the task.

The second goal is that task bodies and helpers are decoupled from context. So instead of `ctx` fields being mutated multiple callers deep into some call stack that originates in a top-level task, tasks now just return the data that they generate. The top level orchestration layer is responsible for 1) parsing `ctx` into the dependencies/input a task requires, and 2) mutating `ctx` with the result from the task.

The end result is that only the top-level task orchestration layer is allowed to touch `ctx`. `ctx` still gets mutated along the way, and there are still places where we have to lie to the type system ("yes, `ctx.git` is defined, I promise"). But these interactions happen in narrowly defined places, rather than threaded throughout the entire codebase.

Note that this assumes we can always cleanly accumulate data throughout the task and return it to the top layer. In some cases, this may get tricky, but from a cursory overview of the tasks, I believe it can be done. And if we run into somewhere where it can't, we can think about some sort of escape hatch back to context. But I hope to avoid that if at all possible.

## Implementation in this PR

This PR sets up the initial scaffolding of this pattern, as well as migrating the `auth` task, just as a proof of concept. `auth` is the first task in the pipeline and is extremely simple. Notably, it has no return type, so that part of the process isn't visible in this PR, but the overall task orchestration is present here.

I also have a working refactor of the `storybookInfo` task in PR #1306. This is a more complex task and includes a return type, if you want to see what that looks like. This PR is subject to change and is not review ready, but the core of the refactoring work is there and tests are all passing.

I recommend reviewing commit by commit.
1. First commit is a small bit of groundwork: adding a `Runtime` bag of settings. This is a concession to one line of code that mutates `ctx.options`. Mutating this value feels like an antipattern to me, as it represents (or should, anyway) the options that a user passed to the CLI invocation. Mutating it, therefore, feels odd, and I added a `runtime` field to track runtime changes to these values.
2. Second commit sets up the task orchestration. The gist is that we define `TaskInput` and `TaskResult` types, which themselves are generic over input and output. This is done so that we can type overload `createTask` to handle legacy and refactored task inputs. The refactored version of `createTask` is the orchestration layer responsible for handling `ctx`. The type generics here might be a lot, so it might help to just get a cursory overview of this commit, then look at the next commit, where we plug it all together in the `auth` task. That should make the overall structure of how this works much more clear.
3. Third commit migrates the `auth` task. Here you can see the setup of auth input and dependencies, and get a feel for how this refactoring effectively lets us limit what parts of `ctx` a task has access to.

> [!NOTE]
> The diffs on this PR are rather confusing. I recommend reviewing the old and new versions of files side-by-side, without diff.

I considered an alternative where we simple add some `Pick<Context, ...>` type annotations throughout the tasks and helper functions, and while this helps with narrowing down what a task actually depends on, it doesn't help with context mutation, and so seemed not good enough to be worth the bother, in my opinion. Happy to discuss, however.

## Future work

If we like this pattern, the plan is to migrate each task in the pipeline, one-by-one, until done. Once they're all done, we can remove the legacy `createTask` type overload. Removing `Context` altogether can be explored at a later time.

# Manual QA

Successful build, auth works, UI for auth is correct:
CleanShot 2026-04-30 at 12 42 31@2x

Deliberately wrong project token, auth errors, UI for auth is correct:
image

<details>
  <summary>📦 Published PR as canary version: <code>16.7.1--canary.1302.25332160078.0</code></summary>
  <br />

  ```bash
  npm install chromatic@16.7.1--canary.1302.25332160078.0
  # or 
  yarn add chromatic@16.7.1--canary.1302.25332160078.0
  ```
</details>

### DIFF
--- a/node-src/lib/tasks.ts
+++ b/node-src/lib/tasks.ts
@@ -2,38 +2,127 @@ import chalk from 'chalk';
 import Listr from 'listr';
 import pluralize from 'pluralize';
 
-import { Context, Task, TaskName } from '../types';
+import { Context, Deps, Task, TaskFunction, TaskName, TaskResult } from '../types';
 
 type ValueFunction = string | ((ctx: Context, task: Task) => string);
 
-type TaskInput = Omit<Listr.ListrTask<Context>, 'task'> & {
+type ListrTaskExtras = Pick<Listr.ListrTask<Context>, 'skip' | 'enabled'>;
+
+type LegacyTaskConfig = ListrTaskExtras & {
   name: TaskName;
+  title: string;
   steps: ((ctx: Context, task: Listr.ListrTaskWrapper<Context> | Task) => void | Promise<void>)[];
 };
 
-export const createTask = ({
-  name,
-  title,
-  steps,
-  ...config
-}: TaskInput): Listr.ListrTask<Context> => ({
-  title,
-  task: async (ctx: Context, task: Listr.ListrTaskWrapper<Context>) => {
-    ctx.task = name;
-    ctx.title = title;
-    ctx.startedAt = Number.isInteger(ctx.now) ? ctx.now : Date.now();
+type AdaptedTaskConfig<TInput, TOutput, TPartial = never> = ListrTaskExtras & {
+  name: TaskName;
+  title: string;
+  transitions?: {
+    pending?: (ctx: Context) => Task;
+    success?: (ctx: Context) => Task;
+  };
+  extractInput: (ctx: Context) => TInput;
+  applyOutput?: (ctx: Context, output: TOutput) => void;
+  applyPartial?: (ctx: Context, output: TPartial) => void;
+  run: TaskFunction<TInput, TOutput, Deps, TPartial>;
+};
 
-    ctx.options.experimental_onTaskStart?.({ ...ctx });
-
-    for (const step of steps) {
-      ctx.options.experimental_abortSignal?.throwIfAborted();
-      await step(ctx, task);
-    }
-
-    ctx.options.experimental_onTaskComplete?.({ ...ctx });
-  },
-  ...config,
+const buildDeps = (ctx: Context): Deps => ({
+  log: ctx.log,
+  client: ctx.client,
+  http: ctx.http,
+  env: ctx.env,
+  options: ctx.options,
+  runtime: ctx.runtime,
+  analytics: ctx.analytics,
+  pkg: ctx.pkg,
+  sessionId: ctx.sessionId,
+  packageJson: ctx.packageJson,
 });
+
+function applyAdaptedResult<TInput, TOutput, TPartial>(
+  config: AdaptedTaskConfig<TInput, TOutput, TPartial>,
+  ctx: Context,
+  listrTask: Listr.ListrTaskWrapper<Context>,
+  result: TaskResult<TOutput, TPartial>
+) {
+  switch (result.kind) {
+    case 'continue':
+      config.applyOutput?.(ctx, result.output);
+      if (config.transitions?.success)
+        transitionTo(config.transitions.success, true)(ctx, listrTask);
+      return;
+    case 'partial':
+      config.applyPartial?.(ctx, result.output);
+      ctx.skip = true;
+      return;
+    case 'skip':
+      ctx.skip = true;
+      return;
+    default:
+      result satisfies never;
+  }
+}
+
+async function runAdaptedTask<TInput, TOutput, TPartial>(
+  config: AdaptedTaskConfig<TInput, TOutput, TPartial>,
+  ctx: Context,
+  listrTask: Listr.ListrTaskWrapper<Context>
+) {
+  ctx.options.experimental_abortSignal?.throwIfAborted();
+  if (config.transitions?.pending) transitionTo(config.transitions.pending)(ctx, listrTask);
+  const input = config.extractInput(ctx);
+  const result = await config.run(buildDeps(ctx), input);
+  applyAdaptedResult(config, ctx, listrTask, result);
+}
+
+async function runLegacyTask(
+  config: LegacyTaskConfig,
+  ctx: Context,
+  listrTask: Listr.ListrTaskWrapper<Context>
+) {
+  for (const step of config.steps) {
+    ctx.options.experimental_abortSignal?.throwIfAborted();
+    await step(ctx, listrTask);
+  }
+}
+
+/**
+ * Creates a Listr task wrapper that supports two config shapes: a legacy `steps[]` shape
+ * for unmigrated tasks, and a typed `(deps, input) => Promise<TaskResult<output>>` shape
+ * (with `transitions` for UI title/output state changes) that decouples task bodies from
+ * the Context god object.
+ *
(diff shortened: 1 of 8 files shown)


node-src/lib/tasks.ts
```diff
@@ -2,39 +2,128 @@ import chalk from 'chalk';
 import Listr from 'listr';
 import pluralize from 'pluralize';
 
-import { Context, Task, TaskName } from '../types';
+import { Context, Deps, Task, TaskFunction, TaskName, TaskResult } from '../types';
 
 type ValueFunction = string | ((ctx: Context, task: Task) => string);
 
-type TaskInput = Omit<Listr.ListrTask<Context>, 'task'> & {
+type ListrTaskExtras = Pick<Listr.ListrTask<Context>, 'skip' | 'enabled'>;
```
justin-thurman, line 9: Tightening up the type here, which is used by legacy tasks as well. From the `ListrTask` type, we only ever used `skip`, `enabled`, and `title`, so rather than omitting `task`, I swapped to `Pick` to make it more explicit. `title` was moved to our types explicitly because we use a tighter type `string`, compared to the `string | undefined` used by `ListrTask`.

node-src/tasks/gitInfo.ts
```diff
@@ -197,7 +197,7 @@ export const setGitInfo = async (ctx: Context, task: Task) => {
       ctx.rebuildForBuildId = mostRecentAncestor.id;
       if (
         ['PASSED', 'ACCEPTED'].includes(mostRecentAncestor.status) &&
-        !ctx.git.matchesBranch(ctx.options.forceRebuild)
+        !ctx.git.matchesBranch(ctx.runtime.forceRebuild)
```
justin-thurman, line 200: This is the only consumer of this value, and so the only place where we need to update to `runtime`.

justin-thurman:
Skipping release, since this has no user-facing purpose.

node-src/tasks/auth.ts
```diff
@@ -17,46 +17,57 @@ const CreateAppTokenMutation = `
...
+export type AuthInput =
+  | { mode: 'cli'; projectId: string; userToken: string; projectToken: string }
+  | { mode: 'app'; projectToken: string };
 
-  if (projectToken) {
-    const { appToken } = await ctx.client.runQuery<{ appToken: string }>(CreateAppTokenMutation, {
-      projectToken,
-    });
-    return appToken;
+const getToken = async (deps: AuthDeps, input: AuthInput): Promise<string> => {
```
jwir3, line 39:
```suggestion
/**
 * Retrieves an authentication token based on the specified mode.
 *
 * In 'cli' mode, creates a CLI token by authenticating with a user token against
 * the Chromatic API, scoped to a specific project.
 *
 * In 'app' mode, creates an app token by authenticating directly with a project token.
 *
 * @param deps - The authentication dependencies, including the GraphQL client and environment config.
 * @param input - The authentication input, which determines the mode and supplies the required credentials.
 *   - `input.mode === 'cli'`: requires `projectId`, `userToken`, and `projectToken`
 *   - `input.mode === 'app'`: requires `projectToken`
 * @returns A promise resolving to the authentication token string.
 * @throws Will throw if the GraphQL query fails or the mode is not handled (exhaustive check).
 */
const getToken = async (deps: AuthDeps, input: AuthInput): Promise<string> => {
```

node-src/lib/tasks.ts
```diff
@@ -2,39 +2,128 @@ import chalk from 'chalk';
...
+  listrTask: Listr.ListrTaskWrapper<Context>
+) {
+  for (const step of config.steps) {
+    ctx.options.experimental_abortSignal?.throwIfAborted();
+    await step(ctx, listrTask);
+  }
+}
+
+/**
+ * Creates a Listr task wrapper that supports two config shapes: a legacy `steps[]` shape
```
jwir3, line 91: Good job. I like how you were able to adapt this so we can convert them at-will.

node-src/lib/tasks.ts
jwir3: 👍

jwir3 (approved):
> Additionally, Context is publicly exported, as are programmatic entrypoints to run a Chromatic build that take Context as their argument. So any backwards incompatible change would be a major version bump. We should probably explore a deprecation timeline before doing anything so drastic.

We are in the process of not providing support for everything in CLI 6.5 and prior. Perhaps we should deprecate `Context` in the next version so that in the next major version bump we can remove it?

node-src/tasks/auth.ts
justin-thurman: I added a modified version of this.
```
/**
 * Retrieves an authentication token based on the specified mode.
 *
 * In 'cli' mode, creates a CLI token by authenticating with a user token against
 * the Chromatic API, scoped to a specific project.
 *
 * In 'app' mode, creates an app token by authenticating directly with a project token.
 *
 * @param deps - The authentication dependencies, including the GraphQL client and environment config.
 * @param input - The authentication input, which determines the mode and supplies the required credentials.
 *
 * @returns A promise resolving to the authentication token string.
 */
```
Removed the explicit note about the input modes, as that's already covered by the types themselves and likely to drift out of date if anything changes. Removed the `throws` line. Neither `runQuery` or the exhaustive check can throw.

PR #1304, opened 2026-04-29
fix(checkForUpdates): prevent crash on registry timeout

### LINKED ISSUE #1303: Cannot read properties of undefined (reading 'message')
**Chromatic Issues**
We run multiple projects in a monorepo setup, and use the `chromatic --skip` command to skip unaffected projects.

When skipping multiple projects, we intermittently see the following error message:
```
             {
               "timestamp": "2026-04-28T15:24:33.671Z",
               "sessionId": "eea25289-544f-4951-be6b-0053e9b3b06c",
               "gitVersion": "2.30.2",
               "gitBranch": "<BRANCH_NAME>",
               "gitSlug": "<REPO_NAME>",
               "fromCI": true,
               "nodePlatform": "linux",
               "nodeVersion": "22.17.0",
               "packageName": "chromatic",
               "packageVersion": "16.6.0",
               "flags": {
                 "skip": "",
                 "debug": true,
                 "uploadMetadata": true,
                 "branchName": "<BRANCH_NAME>",
                 "outputDir": [],
                 "storybookBuildDir": [],
                 "externals": [],
                 "onlyStoryFiles": [],
                 "onlyStoryNames": [],
                 "untraced": [],
                 "fileHashing": true,
                 "interactive": true
               },
               "configuration": {},
               "buildScript": "storybook build --stats-json",
               "exitCode": 0,
               "exitCodeKey": "OK",
               "errorType": "TypeError",
               "errorMessage": "Cannot read properties of undefined (reading 'message')"
             }
```

**Bug report**
I believe the line at https://github.com/chromaui/chromatic-cli/blob/a172a7938c45097df5bda1f701b0e7716d5aec0e/node-src/lib/checkForUpdates.ts#L76 is causing the error when `err` is undefined.

`err` can be undefined when `rejectIn` rejects the promise using `setTimeout` at https://github.com/chromaui/chromatic-cli/blob/a172a7938c45097df5bda1f701b0e7716d5aec0e/node-src/lib/checkForUpdates.ts#L9

annaet:
Fixes https://github.com/chromaui/chromatic-cli/issues/1303

`rejectIn` was calling `setTimeout(reject, ms)`, which invoked `reject()` with no argument — rejecting the promise with `undefined`. The catch block then passed `undefined` into `shouldReportVersionCheckFailure`, which read `.message` without a guard and crashed the entire CLI with `Cannot read properties of undefined (reading 'message')` after a successful build/skip.

Fixes:
- `rejectIn` now rejects with a real `Error` describing the timeout.
- `shouldReportVersionCheckFailure` accepts `unknown` and bails early when the value isn't an `Error`, so non-Error rejections from any future code path can no longer crash the helper.

### DIFF
--- a/node-src/lib/checkForUpdates.ts
+++ b/node-src/lib/checkForUpdates.ts
@@ -6,7 +6,10 @@ import { Context } from '..';
 import outdatedPackage from '../ui/messages/warnings/outdatedPackage';
 import spawn from './spawn';
 
-const rejectIn = (ms: number) => new Promise<any>((_, reject) => setTimeout(reject, ms));
+const rejectIn = (ms: number) =>
+  new Promise<never>((_, reject) =>
+    setTimeout(() => reject(new Error(`Timed out after ${ms}ms`)), ms)
+  );
 const withTimeout = <T>(promise: Promise<T>, ms: number): Promise<T> =>
   Promise.race([promise, rejectIn(ms)]);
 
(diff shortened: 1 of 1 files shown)


codykaup:
Thanks for submitting this @annaet! We typically create a new PR for all of the CI steps and such so this PR is moving to #1305!

PR #1310, opened 2026-04-30
Allow skip without a project token

codykaup:
Sometimes you want to `--skip=<branch>` on some PRs because they won't have the project token set. Those cases, will currently fail because it can't talk to the backend API. This change allows for this scenario and logs a warning if we're skipping the build _and_ we lack authentication to the Chromatic API. If we have authentication, then it'll continue as normal.

### Before:
https://github.com/codykaup/test-app-a11y/actions/runs/25187067578/job/73847045819
image

### After:
https://github.com/codykaup/test-app-a11y/actions/runs/25187137976/job/73847288790
image

Also tested the GitHub Action (also works):
https://github.com/codykaup/test-app-a11y/actions/runs/25187384332/job/73848140108

<details>
  <summary>📦 Published PR as canary version: <code>16.10.0--canary.1310.25450492740.0</code></summary>
  <br />

  ```bash
  npm install chromatic@16.10.0--canary.1310.25450492740.0
  # or 
  yarn add chromatic@16.10.0--canary.1310.25450492740.0
  ```
</details>

### DIFF
--- a/node-src/index.ts
+++ b/node-src/index.ts
@@ -196,11 +196,13 @@ export async function runAll(ctx: InitialContext) {
       ctx.extraOptions?.configFile || ctx.flags.configFile
     );
 
-    if (await shouldSkipWithoutProjectToken(ctx)) {
+    const partialOptions = getPartialOptions(ctx);
+
+    if (await shouldSkipWithoutProjectToken(ctx, partialOptions)) {
       return;
     }
 
-    const options = getOptions(ctx);
+    const options = getOptions(ctx, partialOptions);
     (ctx as Context).options = options;
     ctx.log.setLogFile(options.logFile);
 
--- a/node-src/index.ts
+++ b/node-src/index.ts
@@ -22,7 +22,7 @@ import { isE2EBuild } from './lib/e2eUtils';
 import { emailHash } from './lib/emailHash';
 import { getConfiguration } from './lib/getConfiguration';
 import getEnvironment from './lib/getEnvironment';
-import getOptions, { getPotentialOptions } from './lib/getOptions';
+import getOptions, { getPartialOptions } from './lib/getOptions';
 import { createLogger } from './lib/log';
 import LoggingRenderer from './lib/loggingRenderer';
 import matchesBranch from './lib/matchesBranch';
--- a/node-src/index.ts
+++ b/node-src/index.ts
@@ -199,6 +199,8 @@ export async function runAll(ctx: InitialContext) {
     const partialOptions = getPartialOptions(ctx);
 
     if (await shouldSkipWithoutProjectToken(ctx, partialOptions)) {
+      ctx.log.warn(skipNoProjectToken());
+      setExitCode(ctx, exitCodes.OK);
       return;
     }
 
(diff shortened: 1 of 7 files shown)


node-src/index.ts
```diff
@@ -192,6 +195,11 @@ export async function runAll(ctx: InitialContext) {
     ctx.configuration = await getConfiguration(
       ctx.extraOptions?.configFile || ctx.flags.configFile
     );
+
+    if (await shouldSkipWithoutProjectToken(ctx)) {
+      return;
+    }
```
justin-thurman: We're now calling `getPotentialOptions` twice every run -- once in `shouldSkipWithoutProjectToken` and then again in `getOptions`. Worth parameterizing the partial options to `shouldSkipWithoutProjectToken` and also `getOptions`? Then we only call `getPotentialOptions` once, pass it to both functions.

node-src/index.ts
```diff
@@ -224,6 +232,33 @@ export async function runAll(ctx: InitialContext) {
...
+    return false;
+  }
+
+  const hasProjectCredentials =
+    !!preflightOptions.projectToken || !!(preflightOptions.projectId && preflightOptions.userToken);
+  if (hasProjectCredentials) {
+    return false;
+  }
+
+  const { branch } = await getCommitAndBranch(ctx as Context, {
```
justin-thurman: What happens if this throws? I'm guessing a very confusing error message. Maybe try/catch and return false as a default on the catch path?

justin-thurman: Potential point of regression here in the future. We're lying to the compiler `Partial<Context> as Context`. It works currently, but if `getCommitAndBranch` ever gets changed to use some field that's not available on the partial context from `getPotentialOptions`, the compiler won't know about it, and it could end up with a confusing error (or silently fall back to false if you implement the try/catch block I suggested above). Any clean way around that? Maybe update `getCommitAndBranch` to take `InitialContext`? I think this is valid, haven't audited. Could also make a new function to just get the branch, since that's all we care about here. Not sure how big a lift that is.

node-src/index.ts
```diff
@@ -224,6 +232,33 @@ export async function runAll(ctx: InitialContext) {
   }
 }
 
+async function shouldSkipWithoutProjectToken(ctx: InitialContext) {
```
justin-thurman: I feel like this should be renamed. `should*`, to me, implies a pure function that just checks some condition. But this function (like every function in this codebase 🙃) mutates context, has (minor) side effects like warning logging from `getCommitAndBranch`, can throw, and sets exit codes. Maybe `handleSkipWithoutProjectToken`?

justin-thurman (commented):
Looks good overall, a few questions though.

node-src/lib/getOptions.ts
```diff
@@ -47,17 +47,9 @@ const defaultUnlessSetOrFalse = (input: string | boolean | undefined, fallback:
...
- * @param ctx The context set when executing the CLI.
- *
- * @returns An object containing parsed options
- */
-// TODO: refactor this function
-// eslint-disable-next-line complexity, max-statements
-export default function getOptions(ctx: InitialContext): Options {
-  const { argv, env, flags, extraOptions, configuration, log, packageJson, packagePath } = ctx;
+// eslint-disable-next-line complexity
+export const getPotentialOptions = (ctx: InitialContext): Partial<Options> => {
```
justin-thurman: Oh, another note I forgot to mention: I think this name is misleading as well. `getPartialOptions` or something like that seems better. The options aren't really potential. They're actual options, just incomplete.

node-src/index.ts
codykaup: Great callout. I reworked it just as you mentioned. I don't _love_ passing those around but there's not a ton we can do since `getOptions` also does a bunch of validation at the moment.

node-src/index.ts
codykaup: Agreed - made that change!

codykaup: Since `getCommitAndBranch` only uses `ctx.log`, I trimmed the param type all the way down to that (we're doing that in other function calls in there). That should help!

jmhobbs (approved):

justin-thurman (approved):
Looks good! Appreciate the tweaks. 🙏

PR #1312, opened 2026-05-01
fix: e2e build command ignores arguments when used with npm

sentience:
This fixes two issues in the E2E (Playwright) build path when using npm (e.g. when using `chromaui/action@latest` in GitHub Actions with `playwright: true`):

- forward `--output-dir` correctly to `build-archive-storybook`
- recover correctly when Storybook writes the output directory in the newer (Storybook 10) multiline format

### Problem

For E2E (Playwright) builds, Chromatic currently constructs the npm command to build a Storybook archive from Playwright test results like this:

```sh
npm exec build-archive-storybook --output-dir=/tmp/chromatic-...
```

With npm, that does not reliably pass `--output-dir` to the binary, so the build output ends up in `./storybook-static` instead of the expected temp dir.

chromatic-cli already has a fallback for that case, where it detects the correct output directory by finding it in the Storybook CLI build log, but the parser only recognizes an older one-line log format (which seems to have been the format produced by Storybook 8). @chromatic-com/playwright@0.13.* has recently upgraded to Storybook 10, which outputs the build directory in a different, multi-line format, which chromatic-cli's fallback fails to parse. As a result, chromatic-cli cannot find the Storybook build output, and fails to find/validate the built test suite output with a message like this:

> Error: Invalid test suite build at /tmp/chromatic--2600-OhzfouEzYsTN (check the build log)

[Here's a sample build failure on one of my own projects](https://github.com/sentience/kevinyank.com/actions/runs/25145860787/job/73705766610?pr=200) triggered by the upgrade to @chromatic-com/playwright@0.13.x.

### Fix

This PR fixes both issues:

- change npm E2E command construction to use `npm exec -- ...`
- make output directory parsing handle both:
  - `Output directory: /path`
  - `Output directory:` followed by the path on the next line

### Why

This showed up for me when upgrading `@chromatic-com/playwright`, which moved the embedded Storybook version forward. The underlying npm forwarding issue was already there, but the newer Storybook log format stopped the existing recovery path from working. I originally reported it as chromaui/chromatic-e2e#312, mistaking it for a `@chromatic-com/playwright` issue, but on further investigation I traced the root cause to these issues in `chromatic-cli` – they were only triggered by the Storybook 10 upgrade in `@chromatic-com/playwright`.


node-src/tasks/build.test.ts
```diff
@@ -254,6 +254,24 @@ describe('setBuildCommand', () => {
...
+      sourceDir: './source-dir/',
+      options: { playwright: true, inAction: true },
+      git: {},
+      log: new TestLogger(),
+    } as any;
+
+    await setBuildCommand(ctx);
+
+    expect(ctx.buildCommand).toEqual(
+      'npm exec -- build-archive-storybook --output-dir=./source-dir/'
```
AriPerkkio: The `npm exec --` part looks wrong. It should be `npm exec build-archive-storybook -- --output-dir=./source-dir/`.

AriPerkkio: Oh well, it does seem to work like this too. 🤔 

Though if we want this to be `npm exec build-archive-storybook -- --output-dir`, following change would do it:

```diff
const parseNexec = ((agent, args) => {
  const map: Record<keyof typeof AGENTS, string> = {
-    npm: 'npm exec -- {0}',
+    npm: 'npm exec {0} -- {1}',
    yarn: 'yarn {0}',
    'yarn@berry': 'yarn {0}',
    pnpm: 'pnpm exec {0}',
    'pnpm@6': 'pnpm exec {0}',
    bun: 'bun run {0}',
  };

  const command = map[agent];

+  if (command.includes('{1}')) {
+    if (args.length === 1) {
+      return command.replace('{0} -- {1}', args[0]);
+    }
+
+    return command
+      .replace('{0}', args[0])
+      .replace('{1}', args.slice(1).map((c) => quote(c)).join(' '))
+      .trim();
+  }

  return command.replace('{0}', args.map((c) => quote(c)).join(' ')).trim();
}) as Runner;
```

sentience: [The docs](https://docs.npmjs.com/cli/v11/commands/npm-exec) for `npm exec` specifically show the `--` after `exec` and before the package name (`build-archive-storybook` in this case).

AriPerkkio: Good point, let's use `npm exec --` then. Weird why they chose that pattern as it's not equivalent to the `npm run <script> -- <args for script bin>`: https://docs.npmjs.com/cli/v11/commands/npm-run#description 🤔

node-src/tasks/prepare.ts
```diff
@@ -77,13 +77,24 @@ function getPathSpecsInDirectory(ctx: Context, rootDirectory: string, dirname =
...
  */
 function getOutputDirectory(buildLog: string) {
-  const outputString = 'Output directory: ';
-  const outputIndex = buildLog.lastIndexOf(outputString);
-  if (outputIndex === -1) return undefined;
-  const remainingLog = buildLog.slice(outputIndex + outputString.length);
-  const newlineIndex = remainingLog.indexOf('\n');
-  const outputDirectory = newlineIndex === -1 ? remainingLog : remainingLog.slice(0, newlineIndex);
-  return outputDirectory.trim();
+  const cleanLog = buildLog.replaceAll(/\u001B\[[0-?]*[ -/]*[@-~]/g, '');
```
AriPerkkio: We should instead use `import { stripVTControlCharacters } from 'node:util'`. It's present in Node ^16. This project (latest `main`) requires Node 22.

node-src/tasks/prepare.ts
```diff
@@ -77,13 +77,24 @@ function getPathSpecsInDirectory(ctx: Context, rootDirectory: string, dirname =
...
+
+  for (let i = lines.length - 1; i >= 0; i -= 1) {
+    const outputIndex = lines[i].lastIndexOf('Output directory:');
+    if (outputIndex === -1) continue;
+
+    const sameLineOutput = lines[i].slice(outputIndex + 'Output directory:'.length).trim();
+    if (sameLineOutput) return sameLineOutput;
+
+    for (let j = i + 1; j < lines.length; j += 1) {
+      const candidate = lines[j].replace(/^[\s│└┌├┬┴─]+/, '').trim();
```
AriPerkkio: I wonder if this should be just `.replace(/^[\s\u2500-\u257F]+/u, '')`. (Claude: `any Box Drawing glyph (U+2500–U+257F)`).

AriPerkkio:
@jmhobbs & @tevanoff, this same issue came up during Vitest QA, SB-1662 (cc. @jonniebigodes).

I've verified that only `npm exec` is affected. `yarn`, `pnpm` and `bun` work as expected.

sentience:
Thanks for the review, @AriPerkkio. Let me know if you want to push those changes you suggested. Otherwise, I'm happy to leave it with you!

AriPerkkio (approved):

jmhobbs:
Canary available at `17.3.1--canary.1385.27040599853.0`, but I imagine this will get merged (from the other branch) shortly.  Ticket filed to get our CI straightened out for external PR's as well.

jmhobbs:
This is out in v17.4.1, thank you!

PR #1328, opened 2026-05-12
Fix getUncommittedHash on Windows

valentinpalkovic:
## Summary

- `getUncommittedHash` joined three `git` commands with `;` inside `(...)` and piped the result to `git hash-object` twice. That pipeline only works under POSIX shells. On Windows it runs through `cmd.exe`, which treats `;` as a literal and folds the rest of the line into the first `git diff` invocation, so the hash always came back as the empty-blob SHA and was reduced to `''`.
- Replace the single shell-piped command with three independent `git` invocations, concatenate the file lists in JavaScript, then feed them to `git hash-object --stdin-paths` and `git hash-object --stdin` via execa's `input` option. Same output on POSIX shells, now works identically on Windows.
- Downstream impact: consumers like the Storybook visual tests addon use `uncommittedHash` to detect local changes (see `ShareSection`'s `checkOutdated`). With the broken pipeline on Windows the hash never changed, so the "Republish" prompt never appeared after a code edit.

## Test plan

- [x] `yarn vitest run node-src/git/git.test.ts` — 26 tests pass, including new cases for empty list, well-known empty-blob hash, and CRLF line endings.
- [x] `yarn lint node-src/git/git.ts node-src/git/git.test.ts` — clean.
- [x] `npx tsc --noEmit` — clean.
- [ ] Manual verification on Windows: edit a tracked file, confirm `getUncommittedHash` now returns a non-empty hash that changes with further edits.

<details>
  <summary>📦 Published PR as canary version: <code>16.10.1--canary.1328.25739265808.0</code></summary>
  <br />

  ```bash
  npm install chromatic@16.10.1--canary.1328.25739265808.0
  # or 
  yarn add chromatic@16.10.1--canary.1328.25739265808.0
  ```
</details>

### DIFF
--- a/node-src/git/git.test.ts
+++ b/node-src/git/git.test.ts
@@ -101,8 +101,69 @@ describe('getBranch', () => {
 
 describe('getUncommittedHash', () => {
   it('returns the uncommitted hash', async () => {
-    execGitCommand.mockResolvedValue('1234567890');
+    execGitCommand
+      .mockResolvedValueOnce('src/a.ts\n') // staged
+      .mockResolvedValueOnce('src/b.ts\n') // unstaged
+      .mockResolvedValueOnce('src/c.ts\n') // untracked
+      .mockResolvedValueOnce('hash-a\nhash-b\nhash-c\n') // hash-object --stdin-paths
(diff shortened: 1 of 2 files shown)


codykaup:
Closing in favor of #1332

PR #1334, opened 2026-05-18
Drop support for Node 20 and define supported Node engines

### LINKED ISSUE #1333: Define Node support using `engines` field
**Feature request**

Chromatic CLI should define supported Node versions using [`package.json`'s `"engines"`](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#engines) field. CI should then run against these specified versions.

It's better to fail early rather than run into unexpected crashes.

**Proposed solution**

Add `"engines"` to `package.json`.

**Alternative solutions**

`process.versions` checks are also fine, but using built-in functionality of runtimes is better.

**Additional context**

Coming from https://github.com/chromaui/chromatic-e2e/pull/351#issuecomment-4477147896.

codykaup:
Closes https://github.com/chromaui/chromatic-cli/issues/1333

It can be confusing if you're running Chromatic with an invalid Node version because random things can start to break and it's not clear why that would be. To be more clear about supported Node versions, this applies 2 changes:

- Adds `engines.node` to our `package.json` to signal during install
- Adds a runtime check to log a warning (not blocking because it _may_ still work without a guarantee)

When I was making this change, I realized Node 20 is end-of-life, and so I dropped Node 20 support alongside this change. 

<details>
  <summary>📦 Published PR as canary version: <code>18.0.0--canary.1334.26111344973.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.0.0--canary.1334.26111344973.0
  # or 
  yarn add chromatic@18.0.0--canary.1334.26111344973.0
  ```
</details>


justin-thurman (approved):
Thanks for doing this!

PR #1358, opened 2026-06-01
Implement Clack-based task rendering and tracer bullet the `auth` task

justin-thurman:
# Description

This PR is the first step to migrating away from Listr and to Clack as our UI library. At a high level, the PR does three things: 1) implements a task orchestration layer that decouples the UI from driving the task, 2) implements a Clack-based renderer, and 3) renders the intro banner and auth task using Clack (this is the tracer bullet to show that pieces 1 and 2 work). I advise thinking of each of the three pieces as a separate PR, but they're bundled together because I felt like we had to have the tracer bullet to prove that the system as a whole works. I'll discuss each piece separately.

### Task orchestration

The task orchestration layer is in the commit `Implement TaskRenderer interface and related plumbing`. The entrypoint is a new `runTask` function that is essentially a copy of the existing `runAdaptedTask` function from `node-src/lib/tasks.ts`. The key difference is that `runAdaptedTask` is coupled to Listr, whereas the new `runTask` function takes a `TaskRenderer` instance that is responsible for updating the UI. The task orchestration is entirely separate from the UI (the benefits of this decoupling are, in my opinion, demonstrated in the `runTask` tests, which define a simple renderer that just records the UI state transitions, and it all works as expected).

Once all tasks are ported over to this new orchestration layer, the entire `lib/tasks.ts` goes away.

### The Clack-based renderer

This is a Clack-based implementation of the `TaskRenderer` interface used by `runTask`. This piece is two commits. The first, `Port Storybook's Clack text wrapping`, is a port of some logic from Storybook. By default, Clack does not handle multiline input cleanly. There's a gutter icon used by various Clack primitives, and it gets clobbered by multiline input. Storybook solved this problem when they migrated to Clack, and we're using their solution. You can review their code [here](https://github.com/storybookjs/storybook/blob/next/code/core/src/node-logger/wrap-utils.ts) if you like. But I lifted it almost verbatim, including tests, and it's pretty complicated. The only functional change is that I'm using `{ hard: true }` as an option call to `wrapAnsi`, which will split a single long token to enforce the character limit (by default, `wrapAnsi` only splits on whitespace).

The second commit, `Implement Clack renderer`, implements the actual Clack-based renderer and a couple of helper one-off functions for, e.g., rendering the `intro`. The render is based by a Clack task-log primitive, which is one of two primitives Storybook uses in their CLI. (The other, backed by a spinner, is for longer-running tasks and will come later.) I also copied their color palette, which uses the `picocolors` library, since part of the goal of this project is to make Chromatic CLI look more like Storybook. Once everything is ported over, we can drop `chalk`.

### The tracer bullet

This is in the `Render intro and auth task with Clack` commit. It adds a small helper, `renderAuth`, that wires up the auth task and its config, alongside a `clackTaskLogRenderer`, and calls `runTask`. Then to render it, I simply removed `auth` from the list of Listr tasks that we run and add this call to `renderAuth`. Future task migrations will follow the same pattern: add a `renderX` helper -> remove the task from the Listr task list -> call the `renderX` helper.

## Limitations

>[!NOTE]
> Reminder: this is merging into a feature branch, so none of these limitations will impact customers. This is just a list of things not solved in this PR that will be solved later.

**File logging**: Listr-based tasks currently log automatically to a file when using the `--log-file` flag (or equivalent config option). I've not implemented that yet, so the auth task and the `intro` frame are no longer included in the log file. I have a solution in progress for that, to come in another PR.

**Non-interactive mode**: This is basically the same problem as file logging, as under non-interactive mode, stdout logging mirrors file logging. It will be solved in the same PR as file logging.

**Aesthetics**: The UI looks silly right now. It opens a Clack frame with intro, runs one task through it, and then there's just the regular Listr stuff from there forward. But it gives you an idea of what it will ultimately look like.

**Legacy tasks**: I've not solved the problem of how to handle the legacy tasks (the ones I didn't get to in my context god object refactor project). Broadly speaking, there are two paths forward: finish the refactoring project, or wrangle those legacy tasks to run through the Clack-based renderer. I have some idea of how to do the latter in a minimally gross way, but I'm going to cross that bridge when I come to it. Which route we go largely depends on how much time I have left in the supercycle when I get there.

**Progress bars**: I haven't yet decided how to handle progress updates from tasks that use a progress bar. Clack has some primitives for this, but I don't exactly know what the implementation will look like. I'll solve that problem when I get there. I did add the `update` method on the `TaskRenderer` interface in anticipation of this, even though the auth task doesn't use it. My hope is that I can just adjust the Clack renderer implementation to do the progress bar stuff and not have to touch the `TaskRenderer` interface going forward -- but no promises.

**Storybook**: You'll note that none of our stories changed with this PR. This is because our stories don't actually run through our renderer. They have a custom rendering implementation that takes a task state object (the stuff in `node-src/ui/tasks`) and renders them in a way that looks like Listr. Since I didn't touch any of that or change the task state objects, the stories didn't change. In a future PR, I'll work on a solution that mimics our Clack renderer and get the stories updated. But this PR is big enough as is.

# Manual QA

Here's what a successful build looks like. Clack rendering for the auth task, and then Listr for the rest.
CleanShot 2026-06-03 at 13 58 20@2x

I also added a 3 second sleep after the renderer's `start` call to demonstrate what the transition looks like.
CleanShot 2026-06-03 at 14 07 08

Here's the `head` of a log file. Note the missing auth task output, but the rest of it is still working.
```txt
14:07:11.243 ⚠ Unsupported Node.js version
             You are running Node.js v20.14.0, but Chromatic supports >=22.0.0.
             Chromatic may not work as expected. Please upgrade Node.js to a supported version based on Node's release schedule:
             https://github.com/nodejs/release#release-schedule
14:07:14.571 Retrieving git information
14:07:15.567 Retrieved git information
                 → Commit '763e9d9' on branch 'HEAD'; found 1 parent build
14:07:15.568 Collecting Storybook metadata
14:07:15.587 Collected Storybook metadata
                 → Build metadata gathered
```

And here's the error state (obtained by running a build against local dev without having the dev server running). You can see the Clack failure UI, and then our regular verbose error logging happens afterwards, unchanged.
CleanShot 2026-06-04 at 09 49 05@2x

<details>
  <summary>📦 Published PR as canary version: <code>18.0.1--canary.1358.28528503813.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.0.1--canary.1358.28528503813.0
  # or 
  yarn add chromatic@18.0.1--canary.1358.28528503813.0
  ```
</details>


node-src/renderer/auth.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Render the auth task.
+ *
+ * @param ctx The CLI context.
+ */
+export async function renderAuth(ctx: Context): Promise<void> {
+  await runTask(
+    ctx,
+    {
+      name: 'auth',
+      title: 'Authenticate',
```
justin-thurman, line 17: This comes from the `initial` object in `node-src/ui/tasks/auth.ts`. Unlike Listr, Clack has no concept of an initial state, so I figure that object can go away (that will come later, when I update our story rendering process) and we can just hardcode the value here. It's only used in `runTask` to set `ctx.title`, which is only used (AFAIK) to render the error header for our error logging that happens outside of Clack.

The `Failed to authenticate` here:
Image

node-src/renderer/auth.test.ts
```diff
@@ -0,0 +1,35 @@
...
+      {
+        name: 'auth',
+        title: 'Authenticate',
+        transitions: { pending: authenticating, success: authenticated },
+        extractInput,
+        run: runAuth,
+        applyOutput: applyAuthOutput,
+      },
+      renderer
+    );
```
codykaup: I don't love having tests that are basically a direct clone of the implementation. Are we gaining anything with this that I'm missing?

node-src/renderer/engine/index.ts
```diff
@@ -0,0 +1,157 @@
+// NOTE: buildDeps currently lives in lib/tasks.ts (the Listr engine). It isn't Listr-specific —
+// it just projects Context into Deps — so when Listr is deleted it should move here. Leaving it
+// alone for now to avoid a noisy diff.
```
codykaup, line 3: Thanks for this!

codykaup (approved):
LGTM! A couple comments for you but nothing blocking.

It's unfortunate the lengths you had to go to juggle the old and new UIs. There's probably going to be quite a bit of refactoring to simplify the setup once Listr is gone. This is a good start though!

justin-thurman:
> It's unfortunate the lengths you had to go to juggle the old and new UIs. There's probably going to be quite a bit of refactoring to simplify the setup once Listr is gone. This is a good start though!

Yeah, there'll be a bit. I'm hopeful it'll just be a bunch of 🔪, but I haven't scoped it all out yet.

node-src/renderer/auth.test.ts
justin-thurman: Yeah, I debated it along the same lines. I added it just to avoid the (unlikely) regression of someone accidentally clobbering the `render/auth.ts` file somehow, but I overlooked that our smoke tests would catch that kind of catastrophic failure. So you're right, this should go.

PR #1371, opened 2026-06-02
:wrench: Add a configuration parameter that controls git timeout.

jwir3:
This adds a configuration parameter that controls the time for a git timeout, in seconds. This is translated through to the git shell command as milliseconds (if set), and defaults to the original 20_000 if unset.

Fixes CAP-4554.

<details>
  <summary>📦 Published PR as canary version: <code>17.4.0--canary.1371.27027417409.0</code></summary>
  <br />

  ```bash
  npm install chromatic@17.4.0--canary.1371.27027417409.0
  # or 
  yarn add chromatic@17.4.0--canary.1371.27027417409.0
  ```
</details>

### DIFF
--- a/node-src/context.ts
+++ b/node-src/context.ts
@@ -1,4 +1,5 @@
 import { InitialContext } from '.';
+import { DEFAULT_GIT_TIMEOUT_SECONDS } from './git/constants';
 import GraphQLClient from './io/graphqlClient';
 import HTTPClient from './io/httpClient';
 import { getConfiguration } from './lib/getConfiguration';
@@ -29,5 +30,14 @@ export async function setupContext(
     retries: 3,
   });
   ctx.configuration = await getConfiguration(configFile);
+  if (
+    ctx.configuration.gitTimeout &&
+    ctx.configuration.gitTimeout !== DEFAULT_GIT_TIMEOUT_SECONDS
+  ) {
+    ctx.log.debug(
+      `git timeout was set to ${ctx.configuration.gitTimeout}, which is different than the default ${DEFAULT_GIT_TIMEOUT_SECONDS} seconds`
+    );
+  }
+
   return ctx as Context;
 }
--- a/node-src/git/execGit.ts
+++ b/node-src/git/execGit.ts
@@ -7,32 +7,47 @@ import { Deps } from '../types';
 import gitNoCommits from '../ui/messages/errors/gitNoCommits';
 import gitNotInitialized from '../ui/messages/errors/gitNotInitialized';
 import gitNotInstalled from '../ui/messages/errors/gitNotInstalled';
+import { DEFAULT_GIT_TIMEOUT_SECONDS } from './constants';
+
+export type GitDeps = Pick<Deps, 'log'> & { options?: { gitTimeout?: number } };
 
 const defaultOptions: Options = {
   env: { LANG: 'C', LC_ALL: 'C' }, // make sure we're speaking English
-  timeout: 20_000, // 20 seconds
   all: true, // interleave stdout and stderr
   shell: true, // we'll deal with escaping ourselves (for now)
 };
 
 /**
+ * Retrieve the git timeout in milliseconds.
+ *
+ * @param depOptions The options object for the Git command (this contains the `gitTimeout` property if it's set).
+ *
+ @returns the git timeout in milliseconds, defaulting to `DEFAULT_GIT_TIMEOUT_SECONDS` if not set.
+ */
+function getGitTimeout(depOptions: GitDeps['options']): number {
+  return depOptions?.gitTimeout ?? DEFAULT_GIT_TIMEOUT_SECONDS * 1000;
+}
+
+/**
  * Execute a Git command in the local terminal.
  *
- * @param context Standard context object.
- * @param context.log Standard context logger.
+ * @param deps Standard context object.
+ * @param deps.log Standard context logger.
+ * @param deps.options Options object for the Git command.
  * @param command The command to execute.
  * @param options Execa options
  *
  * @returns The result of the command from the terminal.
  */
 export async function execGitCommand(
-  { log }: Pick<Deps, 'log'>,
+  { log, options: depOptions }: GitDeps,
   command: string,
   options?: Options
 ) {
   try {
     log.debug(`execGitCommand: ${command}`);
-    const { all, stdout } = await runCommand(command, { ...defaultOptions, ...options });
+    const timeout = getGitTimeout(depOptions);
+    const { all, stdout } = await runCommand(command, { timeout, ...defaultOptions, ...options });
     // If the caller sets `all: false`, then `stdout` will be the output. Otherwise, `all` will
     // contain interleaved stdout and stderr.
     const output = all ?? stdout;
@@ -68,20 +83,23 @@ export async function execGitCommand(
 /**
  * Execute a Git command in the local terminal and just get the first line.
  *
- * @param context Standard context object.
- * @param context.log Standard context logger.
+ * @param deps Standard context object.
+ * @param deps.log Standard context logger.
+ * @param deps.options Options object for the Git command.
  * @param command The command to execute.
  * @param options Execa options
  *
  * @returns The first line of the command from the terminal.
  */
 export async function execGitCommandOneLine(
-  { log }: Pick<Deps, 'log'>,
+  { log, options: depOptions }: GitDeps,
   command: string,
   options?: Options
 ) {
   log.debug(`execGitCommandOneLine: ${command}`);
+  const timeout = getGitTimeout(depOptions);
   const process = runCommand(command, {
+    timeout,
     ...defaultOptions,
     buffer: false,
     ...options,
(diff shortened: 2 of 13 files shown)


jmhobbs:
no-flags

We have so many flags.  Can we make this an env var and/or a config file option only?

jwir3:
> We have so many flags. Can we make this an env var and/or a config file option only?

Sure. I can do this. It just seems as though it would be a little more hidden in that case. I liked the idea of having the documentation be in the command line itself (self-documenting, as it were). The CI config file option or ENV var would make it necessary to document on the web, which, in my opinion, makes it a little more difficult to know that it's there.

codykaup:
> NOTE: I am in the process of testing this, so I haven't verified that the CLI actually builds and runs yet.

Is this ready to be reviewed?

jmhobbs:
> It just seems as though it would be a little more hidden in that case.

I think that is a perk for this, rather than a drawback. It's not a knob we want _most_ users to turn.  That said I'm happy to defer to product on that decision.

codykaup:
> > We have so many flags. Can we make this an env var and/or a config file option only?
> 
> Sure. I can do this. It just seems as though it would be a little more hidden in that case. I liked the idea of having the documentation be in the command line itself (self-documenting, as it were). The CI config file option or ENV var would make it necessary to document on the web, which, in my opinion, makes it a little more difficult to know that it's there.

To me, this feels a lot like the `CHROMATIC_TURBOSNAP_MANIFEST_CONCURRENCY` or `STORYBOOK_BUILD_TIMEOUT` flags. It's not something you'd need to configure in most scenarios but you might need an escape hatch if something goes wrong. Sure, we could add it to the `--help` options but at some point the `--help` output stops being helpful and becomes more noise. 😞

jwir3:
> Is this ready to be reviewed?

Sorry, should've put this in draft state. It is now. I wanted to see if the CI pipeline worked as expected and generate a canary build to use to test the CI. Also, John's feedback on *NO FLAGS* was useful. :D

jwir3:
> I think that is a perk for this, rather than a drawback. It's not a knob we want most users to turn. That said I'm happy to defer to product on that decision.

No, that's ok. I've added it to the configuration instead. I'm coming around to your point of view. :D

> To me, this feels a lot like the CHROMATIC_TURBOSNAP_MANIFEST_CONCURRENCY or STORYBOOK_BUILD_TIMEOUT flags. It's not something you'd need to configure in most scenarios but you might need an escape hatch if something goes wrong. Sure, we could add it to the --help options but at some point the --help output stops being helpful and becomes more noise. 😞

Yeah, the discussion on this happened live in Slack. I'm ok either way, I just wanted to see what should constitute a new flag and what shouldn't.

jwir3:
> Is this ready to be reviewed?

Also, that note at the top should've said: 

> NOTE: I am in the process of testing this, so I haven't verified that the **Action** actually builds and runs yet.

My bad, obviously I tested the actual CLI command. I was just in the process of testing the canary build against an action.

jwir3:
@jmhobbs This should now be ready for review. I have tested both the canary action and the local CLI binary.

jwir3:
FYI, the test failures appear to be unrelated to my patch, but I can look into them and see what we need to do to fix them.

jwir3:
> FYI, the test failures appear to be unrelated to my patch, but I can look into them and see what we need to do to fix them.

Added [CAP-4576](https://linear.app/chromaui/issue/CAP-4576/inconsistent-test-failure-in-ci-on-cli-project) to track this, since my initial fix didn't resolve the weird issue.

node-src/git/execGit.ts
```diff
@@ -10,7 +10,7 @@ import gitNotInstalled from '../ui/messages/errors/gitNotInstalled';
 
 const defaultOptions: Options = {
   env: { LANG: 'C', LC_ALL: 'C' }, // make sure we're speaking English
-  timeout: 20_000, // 20 seconds
+  timeout: 20_000, // default to 20s. This is overridden by the caller if they provide a 'gitTimeout' option.
```
jmhobbs: Should this reference the new constant?

node-src/types.ts
```diff
@@ -32,6 +32,7 @@ export interface Flags {
   exitOnceUploaded?: string;
   exitZeroOnChanges?: string;
   externals?: string[];
+  gitTimeout?: number;
```
jmhobbs: Does this belong on `Flags`, since it's not a flag?

jmhobbs (commented):
This looks like it only changes the timeout for `nextCommits`.  There's _lots_ more instances of `execGitCommand`, shouldn't this apply to all (or at least most) of them?  e.g. `getCommit`, `findMergeBase`, etc.

node-src/context.ts
```diff
@@ -29,5 +31,8 @@ export async function setupContext(
     retries: 3,
   });
   ctx.configuration = await getConfiguration(configFile);
+  if (ctx.configuration.gitTimeout && ctx.configuration.gitTimeout > DEFAULT_GIT_TIMEOUT_SECONDS) {
+    ctx.log.warn(gitTimeoutBuildTimes(ctx.configuration.gitTimeout));
```
justin-thurman: Is this really needed? Seems obvious that if someone bumps a timeout, then things might take longer. This feels spammy to me.

jwir3: I wanted to make sure it got into the logs, in the event we're trying to debug longer build times.

node-src/types.ts
jwir3: Nope, I'll fix.

node-src/git/execGit.ts
jwir3: Yes, I'll fix it.

jwir3:
> This looks like it only changes the timeout for nextCommits. There's lots more instances of execGitCommand, shouldn't this apply to all (or at least most) of them? e.g. getCommit, findMergeBase, etc.

Hm, yes, I thought I changed all of them.

node-src/context.ts
justin-thurman: Don't we have access to their config in those cases? Or not? I can't recall.

jwir3: I could make this more silent, though, if you'd prefer. It seems as though we wanted to call this out, based on Cody's feedback initially during standup.

justin-thurman: Eh, I'm probably making a bigger deal out of this than necessary, given that I'm guessing very few customers will use the feature. It's shippable as-is, imo.

codykaup: It affects certain parts of the build time internal metrics but it's all client side so I'm not too worried about calling it out. A log is probably a good idea though in case someone wants to figure out why a `git` command waited so long before timing out. ¯\\\_\(ツ\)\_\/¯

codykaup: A debug log might be a better fit. Chances are, if they set a custom timeout, they probably don't want to be warned on every single build about it. 😂

jwir3: Ok, I'll make it debug instead.

node-src/git/git.ts
```diff
@@ -511,7 +503,7 @@ export async function getStorybookCreationDate(
  *
  * @returns number The number of committers
  */
-export async function getNumberOfComitters(deps: Pick<Deps, 'log'>) {
+export async function getNumberOfComitters(deps: GitDeps) {
   try {
     return await execGitCommandCountLines(deps, `git shortlog -sn --all --since="6 months ago"`, {
       timeout: 5000,
```
jmhobbs: Should we call out that this is explicitly _not_ using the new config value?

node-src/git/git.ts
jmhobbs: Here too.

node-src/context.ts
```diff
@@ -29,5 +30,11 @@ export async function setupContext(
     retries: 3,
   });
   ctx.configuration = await getConfiguration(configFile);
+  if (ctx.configuration.gitTimeout && ctx.configuration.gitTimeout > DEFAULT_GIT_TIMEOUT_SECONDS) {
+    ctx.log.debug(
+      `git timeout was set to ${ctx.configuration.gitTimeout}, which is longer than the default ${DEFAULT_GIT_TIMEOUT_SECONDS} seconds`
+    );
+  }
```
jmhobbs, line 40: Is it worth logging when the timeout is _less_ as well?  Like, if you accidentally set it to `0` or something and all the git commands fail?

node-src/lib/getConfiguration.ts
```diff
@@ -28,6 +28,7 @@ const configurationSchema = z
     exitZeroOnChanges: z.union([z.string(), z.boolean()]),
     exitOnceUploaded: z.union([z.string(), z.boolean()]),
     ignoreLastBuildOnBranch: z.string(),
+    gitTimeout: z.number().optional(),
```
jmhobbs: Should we set a `.min()` on this?  (see-also the comment about logging when less than than default)

node-src/git/execGit.ts
```diff
@@ -114,20 +121,23 @@ export async function execGitCommandOneLine(
...
  * @returns The number of lines the command returned
  */
 export async function execGitCommandCountLines(
-  { log }: Pick<Deps, 'log'>,
+  { log, options: depOptions }: GitDeps,
   command: string,
   options?: Options
 ) {
   log.debug(`execGitCommandCountLines: ${command}`);
+  const timeout = (depOptions?.gitTimeout ?? DEFAULT_GIT_TIMEOUT_SECONDS) * 1000;
```
jmhobbs: Worth DRY-ing this out? Same thing three times.

jmhobbs (approved):
Looks good, just a couple little, optional things.

node-src/git/git.ts
jwir3: Ah, I missed that. Yeah, I'll add a comment. I wonder if this _should_ be using the new timeout value, though?

node-src/git/git.ts
jwir3: Yup, will do.

node-src/context.ts
jwir3: Sure, I'll add that. I think you're right - it should probably be != rather than >.

node-src/lib/getConfiguration.ts
jwir3: Yes, I'll set a min (what do you think - 5 seconds?)

node-src/git/execGit.ts
jwir3: Yup, will do.

PR #1372, opened 2026-06-03
:robot: Propagate node configuration in order to fix CI smoketests.

jwir3:
We use `unicorn-magic` as a transitive dependency. However, it has two different build configurations: `node` and `default`. Even though we specify `platform: node`, this isn't getting transferred down to our dependency configurations, and thus it causes an error with unknown imports when run in CI.

<details>
  <summary>📦 Published PR as canary version: <code>17.2.1--canary.1372.26902877224.0</code></summary>
  <br />

  ```bash
  npm install chromatic@17.2.1--canary.1372.26902877224.0
  # or 
  yarn add chromatic@17.2.1--canary.1372.26902877224.0
  ```
</details>

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -14,6 +14,7 @@ const config: StorybookConfig = {
   viteFinal: async (config) => {
     config.resolve = {
       ...config.resolve,
+      conditions: ['node'],
       alias: {
         ...config.resolve?.alias,
         os: fileURLToPath(import.meta.resolve('os-browserify/browser')),
--- a/tsdown.config.ts
+++ b/tsdown.config.ts
@@ -14,6 +14,7 @@ export default defineConfig((options) => [
     clean: true,
     platform: 'node',
     target: 'node20', // Lowest supported Node version
+    conditions: ['node'],
     env: {
       SENTRY_ENVIRONMENT: process.env.CI ? 'production' : 'development',
       SENTRY_RELEASE: process.env.SENTRY_RELEASE || 'development',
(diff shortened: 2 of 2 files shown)


codykaup:
Did our dependencies change somehow? Why did this start failing all of a sudden?

codykaup:
> Did our dependencies change somehow? Why did this start failing all of a sudden?

I was curious if maybe something updated from under us so I rebased a draft PR I have and all CI steps passed: https://github.com/chromaui/chromatic-cli/actions/runs/26908461067?pr=1355

Where is this failure happening? I don't think it's happening on `main`. 🤔

jwir3:
> Where is this failure happening? I don't think it's happening on main. 🤔

It's happening on the PR for Coda: #1371. I don't think that PR is related at all to the failures, but I can look again. I also got both the failures (and then clean passes) in previous pushes of that PR, so I _think_ this is intermittent, which is... strange.

After doing a google search, I found this issue: https://github.com/sindresorhus/globby/issues/260 (which is where the idea came from of passing down `node` as a conditional dependency). (I obviously didn't want to take the solution of re-defining that import.

I'm pretty confident that this is the "right" thing to do here, since it seems as though we're not passing the `node` platform all the way down to our transitive dependencies, even though we do specify it as `platform: node`. Though, that said, it seems like this is a weird thing for `unicorm-magic` to be doing. I don't have a great answer for why this started happening now, nor do I have one for why it's intermittent. I can certainly investigate those questions further, though.

jwir3:
This is not the right change. The reason this is happening now is a constant that's importing `getParentCommits.ts`, which also imports `execa` into the browser story. `execa` will never work in the browser, so there's no sense forcing it to compile.

However, we're going to have to find a place to document this so that we can remember this foot gun.

PR #1379, opened 2026-06-04
Implement `broadcastRenderer` to fan out multiple `TaskRenderer`s

justin-thurman:
# Description

This PR implements a `broadcastRenderer` function to compose multiple `TaskRenderer` instances into a single `TaskRenderer`. The composed renderer fans out each hook call to the composed renderers. Errors thrown by the renderer hooks themselves get reported and do not poison siblings or the task itself.

# Manual QA

None, it's not wired up yet. Tested in #1384.

<details>
  <summary>📦 Published PR as canary version: <code>17.3.1--canary.1379.27153568836.1</code></summary>
  <br />

  ```bash
  npm install chromatic@17.3.1--canary.1379.27153568836.1
  # or 
  yarn add chromatic@17.3.1--canary.1379.27153568836.1
  ```
</details>

### DIFF
--- a/node-src/renderer/engine/broadcast.ts
+++ b/node-src/renderer/engine/broadcast.ts
@@ -1,0 +1,41 @@
+import { Task } from '../../types';
+import { TaskRenderer } from './index';
+
+/**
+ * Compose multiple `TaskRenderer`s into one that fans every lifecycle hook out to all of them.
+ *
+ * Each renderer's hook is called in its own try/catch, so a renderer throwing can neither starve
+ * its siblings nor propagate into `runTask`. A caught throw is surfaced via `onError`, which is
+ * itself wrapped in a try/catch so that a throwing reporter is silently swallowed rather than
+ * propagated.
+ *
+ * @param renderers The renderers to broadcast to.
+ * @param onError Surfaces a renderer-internal throw. If it throws, the throw is swallowed.
+ *
+ * @returns A `TaskRenderer` that forwards each hook to every renderer.
+ */
+export function broadcastRenderer(
+  renderers: TaskRenderer[],
+  onError: (error: unknown, hook: keyof TaskRenderer) => void
+): TaskRenderer {
+  const each = (hook: keyof TaskRenderer) => (state: Task) => {
+    for (const renderer of renderers) {
+      try {
+        renderer[hook](state);
+      } catch (error) {
+        try {
+          onError(error, hook);
+        } catch {
+          // Swallow: a throwing error reporter must not break broadcasting.
+        }
+      }
+    }
+  };
+
+  return {
+    start: each('start'),
+    update: each('update'),
+    succeed: each('succeed'),
+    fail: each('fail'),
+  };
+}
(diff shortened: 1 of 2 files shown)


justin-thurman:
Lint failures are due to some bad rebasing on my part. Working on fixing.

node-src/renderer/engine/broadcast.ts
```diff
@@ -0,0 +1,35 @@
...
+import { TaskRenderer } from './index';
+
+/**
+ * Compose multiple `TaskRenderer`s into one that fans every lifecycle hook out to all of them.
+ *
+ * Each renderer's hook is called in its own try/catch, so a renderer throwing can neither starve
+ * its siblings nor propagate into `runTask`. A caught throw is surfaced via `reportError`.
+ *
+ * @param renderers The renderers to broadcast to.
+ * @param reportError Surfaces a renderer-internal throw. Must not itself throw.
```
codykaup: Are we saying `reportError` can't throw because it would then throw the function and cause other
renderers from getting called? Should we wrap it to prevent the footgun?

node-src/renderer/engine/broadcast.ts
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @returns A `TaskRenderer` that forwards each hook to every renderer.
+ */
+export function broadcastRenderer(
+  renderers: TaskRenderer[],
+  reportError: (error: unknown, hook: keyof TaskRenderer) => void
```
codykaup: NIT: Should we use `onError` since we use that elsewhere?

node-src/renderer/engine/broadcast.ts
```diff
@@ -0,0 +1,35 @@
...
+): TaskRenderer {
+  const each = (hook: keyof TaskRenderer) => (state: Task) => {
+    for (const renderer of renderers) {
+      try {
+        renderer[hook](state);
```
codykaup, line 24: I assume we're okay with throwing if `renderer[hook]` is undefined?

codykaup (approved):

node-src/renderer/engine/broadcast.ts
justin-thurman: Yeah, I went back and forth on this. Highly unlikely, but a guard is relatively cheap. I just hate try/catch in a try/catch. But I'll add it and then just not look at it anymore. 🙈

node-src/renderer/engine/broadcast.ts
justin-thurman: Fixed.

node-src/renderer/engine/broadcast.ts
justin-thurman: IMO, yes. Typescript will guard against forgetting to write the hook method at all, so someone would have to intentionally do some weirdness to get undefined here.

node-src/renderer/engine/broadcast.ts
codykaup: > I just hate try/catch in a try/catch.

Trust me, you're not alone! The nested try/catches are not great but I'd rather have that than requirements in a comment. ¯\\\_\(ツ\)\_\/¯

PR #1386, opened 2026-06-08
Implement Clack-based storybook rendering

justin-thurman:
# Description

This PR adds new story rendering functionality based on what our Clack-based renderer actually emits, as well as updating the auth stories to use this new functionality. In the future, stories will be migrated as the tasks themselves are migrated. At a high level, this PR captures the ANSI output that Clack emits and pipes it through our existing `ansiHTML` path.

This is done through a Vite plugin to get around the fact that Clack can't run in the browser and so can't run in Storybook. I've added an `auth.frames.ts` that lives alongside `auth.stories.ts`. The `stories` file imports the `frames` file with a `?clack` import query. The Vite plugin intercepts this query, runs the Clack-based renderer, and then exports the raw ANSI string. This is what actually gets picked up and rendered in the Storybook.

The trickiest bit of code is the stuff in `settle.ts`. Because we're just writing to an in-memory sink here, there's not really a concept of erasing the current terminal and replacing it with something else. In a TTY terminal, Clack basically does this `emit the start event -> [business logic happens] -> emit the success event -> internally, Clack emits an ANSI control sequence to erase the current line and write a new one`. We're forcing TTY here, so if we just captured the Clack output exactly, it would be a mess of non-terminal events and control sequences. `settle.ts` is responsible for emulating the terminal control logic. So for example, in our `success` stories, we first emit the `start` event (because the renderer requires it) and then the `success` event. `settle.ts` is responsible for capturing the line-clearing control sequence and emulating it in a line buffer so that the end result is only the output of the `success` hook.

TLDR: `settle.ts` implements terminal control sequence semantics on a line buffer.

Finally, I also fixed a redundancy in the task log output where we were calling `CLI_COLORS.dim` on `success` state output. But turns out Clack's task log `message` method already dims the output in exactly the same way. This doesn't cause any problems in the terminal output, but it was causing a double ANSI color sequence in the byte stream, which was uncovered while implementing this.

As usual, I recommend reviewing commit-by-commit.

# Manual QA

The main QA is, of course, the Chromatic run showing the new stories. But I also checked a few additional things:

1. I verified local dev storybook hot reloading works when updating anything in `node-src/renderer` and anything in `ui/tasks`
2. I tested the CLI locally, everything still working as expected.
3. I ran four builds against the CLI repo in staging with the new stories, just to test for flakiness. No flakes: https://www.staging-chromatic.com/builds?appId=6a27001b1e15e303fd4eab2f

<details>
  <summary>📦 Published PR as canary version: <code>18.0.1--canary.1386.28528509298.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.0.1--canary.1386.28528509298.0
  # or 
  yarn add chromatic@18.0.1--canary.1386.28528509298.0
  ```
</details>


package.json
```diff
@@ -70,7 +70,7 @@
   ],
   "scripts": {
     "build": "clean-package && tsdown && clean-package restore",
-    "build-storybook": "storybook build --stats-json",
+    "build-storybook": "cross-env FORCE_COLOR=3 TERM=xterm-256color storybook build --stats-json",
```
justin-thurman, line 73: Need the `FORCE_COLOR` and `TERM` env vars to enable color through `picocolors` and hyperlink support.

node-src/renderer/storybook/captureTask.test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+describe('captureTask', () => {
+  // These assert only what wouldn't be captured in the regular Chromatic diff (the big one being hyperlinks)
+  let previousTermProgram: string | undefined;
+  let previousCI: string | undefined;
+  let previousColumns: number | undefined;
+
+  beforeEach(() => {
+    previousTermProgram = process.env.TERM_PROGRAM;
+    previousCI = process.env.CI;
```
codykaup: Instead of this `process.env` dance, should we use [stubEnv](https://vitest.dev/api/vi.html#vi-stubenv)? That
way we could stub `process.env` fields then do a `vi.unstubAllEnvs()` when the tests are done?

codykaup (approved):
LGTM! Some of that is DENSE but at least it's strictly for our internal Storybook rendering so it's not a huge deal.

node-src/renderer/storybook/captureTask.test.ts
justin-thurman: Oh, thanks. I always forget that little guy. Note that I still have to wrangle `process.stdout.columns` manually, since it's not an env var, but I cleaned up all the env var stuff.

justin-thurman:
> LGTM! Some of that is DENSE but at least it's strictly for our internal Storybook rendering so it's not a huge deal.

Yeah, that was my thinking as well. And to some degree, the density/complexity is unavoidable. Clack isn't really meant to be used this way. But for internal tooling, it felt acceptable to me. And we shouldn't have to touch it much. Though note I expect at least one edit within here. Once we add progress bars for upload/snapshot, we'll have some more control sequences to handle, but that should be it.

PR #1388, opened 2026-06-09
Implement mid-task UI updates

justin-thurman:
# Description

This PR adds support for tasks to make mid-task UI updates through the task's `TaskRenderer.update` call, by handing the task a `report` callback in its dependencies. This will be used to support textual-only updates (like `gitInfo` noting in the UI that it's skipping a build before calling the skip build GQL mutation) and numeric updates for things like uploading files or taking snapshots (these numeric updates will only work with certain `TaskRenderer` implementations; currently planning a progress bar renderer, but it's not yet implemented).

# Manual QA

This isn't wired up to any migrated tasks yet, so there's not much to test, but I did verify builds still work, and went through the existing auth pending, success, and failure modes. They all still work as expected. The real QA for this will come in the PR that migrates the `gitInfo` task, which uses this update hook.

<details>
  <summary>📦 Published PR as canary version: <code>17.4.1--canary.1388.27279488773.0</code></summary>
  <br />

  ```bash
  npm install chromatic@17.4.1--canary.1388.27279488773.0
  # or 
  yarn add chromatic@17.4.1--canary.1388.27279488773.0
  ```
</details>

### DIFF
--- a/node-src/renderer/engine/index.ts
+++ b/node-src/renderer/engine/index.ts
@@ -3,7 +3,7 @@
 // alone for now to avoid a noisy diff.
 import { buildDeps } from '@cli/tasks';
 
-import { Context, Deps, Task, TaskFunction, TaskName, TaskResult } from '../../types';
+import { Context, Deps, Task, TaskFunction, TaskName, TaskReporter, TaskResult } from '../../types';
 
 /**
  * A `TaskRenderer` consumes the `Task` state objects emitted by a task's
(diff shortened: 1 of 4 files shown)


node-src/renderer/engine/index.ts
```diff
@@ -80,6 +84,32 @@ export async function runTask<TInput, TOutput, TPartial = never>(
...
+ * (this callback only accepts numeric progress, not textual updates).
+ *
+ * @param ctx The CLI context (its `title` is updated when a textual update arrives).
+ * @param renderer The active renderer for the running task.
+ *
+ * @returns A `TaskReporter` closed over the current task's renderer.
+ */
+function makeReporter(ctx: Context, renderer: TaskRenderer): TaskReporter {
+  return (update) => {
+    if (update.title !== undefined) ctx.title = update.title;
```
jwir3, line 99: Personally, I'm not a big fan of single-line `if` statements, but I won't hold up review for this. Also, this could be simplified to `if (update.title)`, unless there's some explicit reason you want to test for not `undefined`.

jwir3 (approved):
One minor comment, otherwise LGTM!

node-src/renderer/engine/index.ts
justin-thurman: It's just to match Listr semantics. Our previous tasks would update the ctx.title in this way too (i.e., allowing the empty string). I'm not sure it will ever be used anywhere, but the individual renderers can decide what to do with the information.

PR #1389, opened 2026-06-09
Migrate `gitInfo` to Clack

justin-thurman:
# Description

This PR migrates the `gitInfo` task from Listr to Clack. Along the way, I found a couple oversights with how renderer `update` calls are handled. These were surfaced because this is the first task to actually use the `update` hook. I bundled the fixes to these in this PR so that it was easier to see them in action. Hopefully that doesn't make things too hard to read. I've split these functionality changes into their own commits. Here's the breakdown:

## Rendering `gitInfo` with Clack

The first commit migrates the `gitInfo` task to the `runTask` helper that renders through Clack. This includes the boilerplate stuff of defining the `renderGitInfo` helper (basically the same as the `renderAuth` helper that came before) and plugging in all the pieces. The bigger change here is around the mid-task UI updates that `gitInfo` needs when skipping a build. Previously, we had a gross `onSkippingBuild` callback that we wired up in `extractGitInfoInput`. The new pattern relies on the `deps.report` callback added in #1388, which is much cleaner. The building of that reporter happens in `renderer/engine/index.ts`. The `gitInfo` task itself knows nothing about it and just calls the reporter with the correct UI state object, which gets passed through to the renderer's `update` call. So all the plumbing and testing around `onSkippingBuild` is removed in this commit.

## Fixing Storybook rendering for `update` the renderer hook

The second commit fixes an omission in the pipeline for rendering Clack in Storybook. This logic branches on the UI state object's `status` field, but I originally overlooked the fact that mid-task UI updates (like `gitInfo`'s `skippingBuild` use `status === 'pending'`. This is also used by the starting UI state objects. So I needed to add support for an `updating` status, which calls the renderer's `update` hook. This added an additional wrinkle, because Clack doesn't allow changing a task log's title except for terminal hooks (succeed/fail), so the UI state after calling `update` is dependent on the state object used to `start` the task log. To support this flow, I added an optional `startingState` parameter to `captureTask`. When provided, we use this state object for the `renderer.start` call, before we call `renderer.update`.

## Fixing redundancy in how `clackTaskLogRenderer` renders the `update` hook

The third commit slightly modifies the text we display when we call `clackTaskLogRenderer.update`. Previously, we called it with `state.title + '\n' + state.output`. This resulted in a task log that looked something like:
```
<bold>Retrieving git info</bold>      <- inherited state.title from start call
Skipping build                        <- state.title
Skipping build for commit 'abc'       <- state.output
```
Obviously, this is a bit redundant. I updated it to only render the `state.output`, falling back to `state.title` if `state.output` is falsy, so we get:
```
<bold>Retrieving git info</bold>
Skipping build for commit 'abc'
```
Once this project is done, we should look into refactoring these `Task` UI state objects, as `TaskRenderer` consumes them in a different way than Listr does. But that's out of scope here.

## Migrating `gitInfo` stories to Clack-based renderer

The fourth commit updates the `gitInfo` stories to use the new Clack rendering pipeline. This includes 1) a boilerplate port in the pattern established in #1386, 2) adding a `startingState` argument to the `Skipping` story, and 3) updating `skippingBuild.status` to the new `'updating'` value that drives the correct rendering path in the Storybook rendering pipeline. I audited the codebase to confirm that this field is not consumed anywhere except the Storybook rendering logic. (This is a good code smell for the previously mentioned refactor, but again, out of scope for now.)

## Updating the `SkipFailed` story

When `gatherGitInfo` fails to skip a build, it throws an error with `skipFailed().output` as its error message. This error is then fed into a function that generates an error UI state object for the renderer. The `skipFailed` object itself is not used. I had to tweak the `SkipFailed` story slightly to use this runtime-derived fallback error state so that the story matches what the user actually sees. There is the risk of drift here, if the task changes what error it throws, but I feel like it's acceptable, as that risk has always been present with these stories, and tying the storybook to the task-running code would be a very large lift that would arguably be against the spirit of Storybook anyway.

>[!NOTE]
> There are also some deliberate UI messaging changes and discussion around some other updates we might make. See the UI review on Chromatic for discussion around these.

# Manual QA

I set up a [staging project](https://www.staging-chromatic.com/builds?appId=6a29a44e846d014918beb9f2) with a fresh repo.

**Build 1:** regular happy path
CleanShot 2026-06-10 at 13 57 52@2x

Log file head:
```
13:55:24.016 Chromatic CLI v17.4.1--canary.1389.27295086586.0
             https://www.chromatic.com/docs/cli 
13:55:24.032 Authenticating with Chromatic [staging] 
13:55:24.032     → Connecting to https://index.staging-chromatic.com
13:55:24.405 Authenticated with Chromatic [staging]
13:55:24.405     → Using project token '****************8273'
13:55:24.406 Retrieving git information
13:55:25.051 Retrieved git information
13:55:25.051     → Commit '1524e3b' on branch 'master'; no ancestor found
13:55:25.053 Collecting Storybook metadata
```

**Build 2:** rebuild noop

Ran the same command again on the same commit. Git info skips due to a rebuild of a fully passed/accepted build. Note the Listr tasks all skip as well, showing the `ctx.skip` propagation works.
CleanShot 2026-06-10 at 14 00 47@2x

Log file head:
```
13:59:39.641 Chromatic CLI v17.4.1--canary.1389.27295086586.0
             https://www.chromatic.com/docs/cli 
13:59:39.661 Authenticating with Chromatic [staging] 
13:59:39.662     → Connecting to https://index.staging-chromatic.com
13:59:40.181 Authenticated with Chromatic [staging]
13:59:40.182     → Using project token '****************8273'
13:59:40.183 Retrieving git information
13:59:41.052 ℹ Skipping rebuild of an already fully passed/accepted build
             A build for the same commit as the last build on the branch is considered a rebuild.
             If the last build is passed or accepted, the rebuild is skipped because it shouldn't change anything.
```

**Build 3:** explicit skip with `--skip 'master'`
CleanShot 2026-06-10 at 14 02 18@2x

Log file head:
```
14:01:51.169 Chromatic CLI v17.4.1--canary.1389.27295086586.0
             https://www.chromatic.com/docs/cli 
14:01:51.191 Authenticating with Chromatic [staging] 
14:01:51.191     → Connecting to https://index.staging-chromatic.com
14:01:51.525 Authenticated with Chromatic [staging]
14:01:51.525     → Using project token '****************8273'
14:01:51.526 Retrieving git information
14:01:51.653     → Skipping build for commit 1524e3b
14:01:51.803 Skipping build
14:01:51.804     → Skipped build for commit 1524e3b due to --skip
```
Note that this is the only skip path that does the mid-task UI update with the `skippingBuild` UI state object. That's why you see the `Skipping build for commit` line in the log file here.

I ran the same command with a local CLI build that adds a sleep after the UI transition so you can see it in action.
CleanShot 2026-06-10 at 14 08 36

<details>
  <summary>📦 Published PR as canary version: <code>17.4.2--canary.1389.27434918549.0</code></summary>
  <br />

  ```bash
  npm install chromatic@17.4.2--canary.1389.27434918549.0
  # or 
  yarn add chromatic@17.4.2--canary.1389.27434918549.0
  ```
</details>

### DIFF
--- a/node-src/renderer/engine/clack/renderer.test.ts
+++ b/node-src/renderer/engine/clack/renderer.test.ts
@@ -1,5 +1,5 @@
 import { taskLog as clackTaskLog } from '@clack/prompts';
-import { describe, expect, it, vi } from 'vitest';
+import { describe, expect, it, test, vi } from 'vitest';
 
 import { Task } from '../../../types';
 import { clackTaskLogRenderer } from './renderer';
@@ -116,18 +116,39 @@ describe('clackTaskLogRenderer', () => {
       expect(formatted).toBe('Done');
     });
 
-    it('renders the title and the output', () => {
-      const formatted = formatVia('update', { title: 'Building', output: 'compiling' });
-
+    test.for(['succeed', 'fail'])('%s transition renders title and output', (transition) => {
+      const formatted = formatVia(transition as keyof typeof TRANSITION_METHOD, {
+        title: 'Building',
+        output: 'compiling',
+      });
       expect(formatted).toContain('Building');
       expect(formatted).toContain('compiling');
     });
 
-    it('wraps long output across multiple lines', () => {
+    it('update renders only output', () => {
+      const formatted = formatVia('update', { title: 'Building', output: 'compiling' });
+
+      expect(formatted).not.toContain('Building');
+      expect(formatted).toContain('compiling');
+    });
+
+    it('update renders title when there is no output', () => {
+      const formatted = formatVia('update', { title: 'Building' });
+      expect(formatted).toContain('Building');
+    });
+
+    it('wraps long terminal-state output across multiple lines', () => {
+      const longOutput = 'foo'.repeat(40);
+      const formatted = formatVia('succeed', { title: 'Building', output: longOutput });
+
+      expect(formatted.split('\n').length).toBeGreaterThan(2);
+    });
+
+    it('wraps long update output across multiple lines', () => {
       const longOutput = 'foo'.repeat(40);
       const formatted = formatVia('update', { title: 'Building', output: longOutput });
 
-      expect(formatted.split('\n').length).toBeGreaterThan(2);
+      expect(formatted.split('\n').length).toBeGreaterThan(1);
     });
   });
 });
--- a/node-src/renderer/engine/clack/renderer.ts
+++ b/node-src/renderer/engine/clack/renderer.ts
@@ -26,7 +26,7 @@ export function clackTaskLogRenderer(output?: Writable): TaskRenderer {
       if (state.output) taskLog.message(wrapTextForClack(state.output));
     },
     update: (state: Task) => {
-      taskLog.message(taskMessageFormatter(state));
+      taskLog.message(wrapTextForClack(state.output || state.title));
     },
     succeed: (state: Task) => {
       taskLog.success(taskMessageFormatter(state));
--- a/node-src/renderer/gitInfo.ts
+++ b/node-src/renderer/gitInfo.ts
@@ -1,0 +1,37 @@
+import {
+  applyGitInfoOutput,
+  applyGitInfoPartial,
+  extractGitInfoInput,
+  gatherGitInfo,
+} from '../tasks/gitInfo';
+import { Context } from '../types';
+import { pending, skippedForCommit, skippedRebuild, success } from '../ui/tasks/gitInfo';
+import { runTask } from './engine';
+import { clackTaskLogRenderer } from './engine/clack/renderer';
+import { getRenderer } from './engine/getRenderer';
+
+/**
+ * Render the gitInfo task.
+ *
+ * @param ctx The CLI context.
+ */
+export async function renderGitInfo(ctx: Context): Promise<void> {
+  await runTask(
+    ctx,
+    {
+      name: 'gitInfo',
+      title: 'Retrieve git information',
+      transitions: {
+        pending,
+        success,
+        partial: (ctx, partial) =>
+          partial.phase === 'skip-commit' ? skippedForCommit(ctx) : skippedRebuild(),
+      },
+      extractInput: extractGitInfoInput,
+      run: gatherGitInfo,
+      applyOutput: applyGitInfoOutput,
+      applyPartial: applyGitInfoPartial,
+    },
+    getRenderer(ctx, clackTaskLogRenderer)
+  );
+}
--- a/node-src/tasks/gitInfo.test.ts
+++ b/node-src/tasks/gitInfo.test.ts
@@ -78,6 +78,7 @@ const buildDeps = (overrides: Partial<GitInfoDeps> = {}): GitInfoDeps => ({
   options: {} as any,
   runtime: {} as any,
   packageJson: {},
+  report: vi.fn(),
   ...overrides,
 });
 
@@ -87,7 +88,6 @@ const buildInput = (overrides: Partial<GitInfoInput> = {}): GitInfoInput => ({
   isLocalBuild: false,
   skip: false,
   onlyChanged: false,
-  onSkippingBuild: vi.fn(),
   ...overrides,
 });
 
@@ -241,14 +241,17 @@ describe('gatherGitInfo', () => {
 
   it('returns skip-commit partial when skip matches branch and mutation succeeds', async () => {
     client.runQuery.mockResolvedValue(true);
-    const onSkippingBuild = vi.fn();
-    const result = await gatherGitInfo(buildDeps(), buildInput({ skip: true, onSkippingBuild }));
+    const report = vi.fn();
+    const result = await gatherGitInfo(buildDeps({ report }), buildInput({ skip: true }));
 
     expectKind(result, 'partial');
     expectPhase(result.output, 'skip-commit');
     expect(result.output.git.commit).toBe(commitInfo.commit);
     expect(result.output.projectMetadata).toMatchObject({ hasRouter: true });
-    expect(onSkippingBuild).toHaveBeenCalledWith(result.output.git);
+    expect(report).toHaveBeenCalledWith({
+      title: 'Skipping build',
+      output: `Skipping build for commit ${result.output.git.commit.slice(0, 7)}`,
+    });
   });
 
   it('throws when skip matches branch but mutation returns falsy', async () => {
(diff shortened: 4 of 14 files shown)


node-src/tasks/gitInfo.test.ts
```diff
@@ -414,16 +416,4 @@ describe('extractGitInfoInput', () => {
       untraced: ['bar'],
     });
   });
-
-  it('onSkippingBuild writes git to ctx and transitions the listr task title', () => {
```
justin-thurman, line 418: `applyPartial` in `runTask` does this. We don't need it here anymore, since `skippingBuild` is refactored to no longer rely on `ctx.git` being set and instead accepts `git` directly.

node-src/tasks/gitInfo.ts
```diff
@@ -220,7 +210,8 @@ export async function gatherGitInfo(
   git.matchesBranch = (glob: string | boolean) => matchesBranch(branch, glob);
 
   if (git.matchesBranch(skip)) {
-    onSkippingBuild(git);
+    const { title, output } = skippingBuild(git);
+    report({ title, output });
```
justin-thurman, line 214: `report` doesn't consume the `status` field from the `skippingBuild` UI state object. It compiles to pass in the `skippingBuild` result directly, but I figured be explicit.

node-src/renderer/engine/clack/renderer.test.ts
```diff
@@ -116,18 +116,39 @@ describe('clackTaskLogRenderer', () => {
       expect(formatted).toBe('Done');
     });
 
-    it('renders the title and the output', () => {
+    test.for(['succeed', 'fail'])('%s transition renders title and output', (transition) => {
```
jwir3, line 119: I think `it` and `test` are aliases, so you can use `it.for` (I think), which would make it more consistent with the rest of the file, and not require the `test` alias to be imported.

node-src/renderer/engine/clack/renderer.ts
```diff
@@ -26,7 +26,7 @@ export function clackTaskLogRenderer(output?: Writable): TaskRenderer {
       if (state.output) taskLog.message(wrapTextForClack(state.output));
     },
     update: (state: Task) => {
-      taskLog.message(taskMessageFormatter(state));
+      taskLog.message(wrapTextForClack(state.output || state.title));
```
jwir3, line 29: Super minor nitpick: 

```suggestion
      taskLog.message(wrapTextForClack(state.output ?? state.title));
```

I think you want to check for nullish values here, right? So, if `state.output` is null (or `undefined`), do something. In the case you have right now, if `state.output` is empty, then it will fall back to `state.title`. Maybe document it in a comment if this is intentional?

node-src/ui/tasks/gitInfo.frames.ts
```diff
@@ -0,0 +1,41 @@
...
+export const Skipping = () => captureTask(skippingBuild(git as any), pending());
+
+export const Skipped = () => captureTask(skippedForCommit({ git } as any));
+
+export const SkippedRebuild = () => captureTask(skippedRebuild());
+
+// gitInfo registers no `failure` transition, so a skip failure renders the engine's fallback
+// failure state, built from the pending title and the error gatherGitInfo throws.
+export const SkipFailed = () =>
+  captureTask(fallbackFailureState(pending().title, new Error(skipFailed().output)));
```
jwir3, line 41: I don't think this is quite right. I think that `SkipFailed()` will go:

start(pending()) -> update(skippingBuild()) -> fail(error)

IOW, for a state with `status: error` (the output from `fallbackFailureState()`, I think the following happens:

```
renderer.start(startingState);
renderer.fail(state);
```

But, in `gatherGitInfo`, here: https://github.com/chromaui/chromatic-cli/pull/1389/changes#diff-6f0d95371c049d074aee49374336fe90e98067a9f33c2a6180e5a570c43cdd25R213 it's wired to call `renderer.update()` in `report`.  This (I think) emits a visible line in the output, right?

The runtime sequence start → update → fail is intentional and correct. The user should see the "Skipping build" message appear before the failure - this is correct. However, the _story_ doesn't replicate this behavior, I don't think.  The `SkipFailed` story calls captureTask with a `status: 'error'` state, and `drive` only does `start` → `fail` for error states. So, the captured frame is missing the update line that a real user would see. 

Again, this is pretty minor, but maybe worth fixing where `drive` actually calls `update` before `fail`?

node-src/renderer/gitInfo.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Render the gitInfo task.
+ *
+ * @param ctx The CLI context.
+ */
+export async function renderGitInfo(ctx: Context): Promise<void> {
+  await runTask(
+    ctx,
+    {
+      name: 'gitInfo',
+      title: 'Retrieve git information',
```
jwir3, line 23: This is inconsistent with `pending().title` here: https://github.com/chromaui/chromatic-cli/pull/1389/changes#diff-7e5babc382e1e08229d6f6b9f8d5aeade1a652e21b569d42a5d3103695091994R30

I think they should both be `Retrieving git information`.

jwir3 (approved):
👍 Overall, looks good. There's one situation where I think the story doesn't output the same as what the user sees. Whether you want to fix it or not is up to you, I won't hold up review on that.

One other, super nit-pick comment:

Commit d1a832c has the following commit message: 

> Clack task log renderer update hook only shows output

Which is different from all your other commit messages, which are in the imperative form. I'd recommend rebasing to change that to:

> Make Clack task log renderer update hook only show output

(These are all minor suggestions, BTW)

node-src/renderer/engine/clack/renderer.test.ts
justin-thurman: Oh, didn't realize `it.for` was a thing. In any case, I'm going to leave this alone, because spoiler: this test case gets split up in the next PR. 😆

node-src/renderer/engine/clack/renderer.ts
justin-thurman: This is intentional. The empty string is meant to fall back to title. The test codifies it better than a comment, imo, which can drift.

node-src/ui/tasks/gitInfo.frames.ts
justin-thurman: The story is actually correct here, as it's intended to capture only the end-state. Clack messages that are rendered by the `update` call get erased by the terminal call (`fail`, in this case). So the message rendered by `update(skippingBuild())` does not show up in the final output to the user. The story reflects this.

node-src/renderer/gitInfo.ts
justin-thurman: This comes from `initial.title`, not `pending().title`. I just don't reference `initial.title` directly because the entire `initial` object went away. The string is only used here, but the grammar is correct and needs to be this way to make sense when it's rendered as `Failed to retrieve git information`.

PR #1391, opened 2026-06-10
Add context around file adds/moves in `BaselineCheckoutFailedError` bails

codykaup:
This PR adds additional tags to our Sentry exceptions to add more context to our `BaselineCheckoutFailedError` error in TurboSnap. To do that, we try to determine if a field was added or moved since the baseline because checking out either will fail earlier in the TurboSnap diff process. Ideally, we would use this information to follow the file around instead of bailing but we'll start with data collection first.

I was using my sample monorepo project to test out each scenario. I'll attempt to describe how I QAed each below. It uses `Nx` with `Yarn`.

> [!NOTE]
> This addition does NOT get sent to the backend. It's merely adding Sentry details.

To force these exceptions to go to Sentry, I patched the CLI to enable it:

```
diff --git a/node-src/errorMonitoring.ts b/node-src/errorMonitoring.ts
index f673ddac..987b1b4f 100644
--- a/node-src/errorMonitoring.ts
+++ b/node-src/errorMonitoring.ts
@@ -51,10 +51,8 @@ Sentry.init({
   release: process.env.SENTRY_RELEASE || process.env.npm_package_version,
   dist: process.env.SENTRY_DIST,
   sampleRate: 1,
-  environment: process.env.SENTRY_ENVIRONMENT,
-  enabled:
-    process.env.DISABLE_ERROR_MONITORING !== 'true' &&
-    process.env.SENTRY_ENVIRONMENT !== 'development',
+  environment: 'development',
+  enabled: true,
   enableTracing: false,
   integrations: [],
   initialScope: {
```

## `baselineManifestAdded`

I added a new package to my monorepo and commited it. Then ran a build with TurboSnap enabled (`--only-changed`). Doing this, yielded the expected bail:

```
execGitCommand: git cat-file -e "ac63e5eb474358e6a3b1e9ee9a2a5ea5b436448a^{commit}"
execGitCommand result: ''
execGitCommand: git diff --name-status --find-renames=20% ac63e5eb474358e6a3b1e9ee9a2a5ea5b436448a HEAD -- ":(glob,top)**/package.json"
execGitCommand result: 'A  packages/lambda/package.json'
⚠ TurboSnap disabled due to file change
Found a package file change in packages/lambda/package.json or its module sibling yarn.lock
A full build is required because this file cannot be linked to any specific stories.
ℹ Read more at https://www.chromatic.com/docs/turbosnap#how-it-works
```

Then in Sentry:
Screenshot 2026-06-10 at 2 58 26 PM

## `baselineManifestMoved`

Following [baselineManifestAdded](#baselineManifestAdded), I moved the `lambda` package to `lambda2` then ran another build. Doing this, yielded the expected bail:

```
execGitCommand: git cat-file -e "8619400ab8ee0aaac20a6b1c39c5edd2b971cc50^{commit}"
execGitCommand result: ''
execGitCommand: git diff --name-status --find-renames=20% 8619400ab8ee0aaac20a6b1c39c5edd2b971cc50 HEAD -- ":(glob,top)**/package.json"
execGitCommand result: 'R100       packages/lambda/package.json    packages/lambda2/package.json'
⚠ TurboSnap disabled due to file change
Found a package file change in packages/lambda2/package.json or its module sibling yarn.lock
A full build is required because this file cannot be linked to any specific stories.
ℹ Read more at https://www.chromatic.com/docs/turbosnap#how-it-works
```

Then in Sentry:
Screenshot 2026-06-10 at 2 58 59 PM

## `unknownBaselineCheckoutFailure`

For this one, I forced the error by patching the CLI:

```
diff --git a/node-src/lib/turbosnap/classifyBailRootCause.ts b/node-src/lib/turbosnap/classifyBailRootCause.ts
index ab345544..2d9a4fc8 100644
--- a/node-src/lib/turbosnap/classifyBailRootCause.ts
+++ b/node-src/lib/turbosnap/classifyBailRootCause.ts
@@ -51,6 +51,7 @@ async function classifyBaselineCheckoutFailureTags(
   if (!(error instanceof BaselineCheckoutFailedError)) {
     return undefined;
   }
+  return { baseline_failure_kind: 'unknownBaselineCheckoutFailure' };
 
   const { reference, fileName } = parsePathspec(error.pathspec);
 
```

Then I built the CLI, added a new package in my monorepo, then ran another TurboSnap build. Then I got the following bail reason:

```
⚠ TurboSnap disabled due to file change
Found a package file change in packages/lambda/package.json or its module sibling yarn.lock
A full build is required because this file cannot be linked to any specific stories.
ℹ Read more at https://www.chromatic.com/docs/turbosnap#how-it-works
```

Then in Sentry:
Screenshot 2026-06-10 at 3 23 51 PM

<details>
  <summary>📦 Published PR as canary version: <code>17.6.0--canary.1391.27703306584.0</code></summary>
  <br />

  ```bash
  npm install chromatic@17.6.0--canary.1391.27703306584.0
  # or 
  yarn add chromatic@17.6.0--canary.1391.27703306584.0
  ```
</details>

### DIFF
--- a/node-src/git/git.ts
+++ b/node-src/git/git.ts
@@ -221,6 +221,73 @@ export async function getChangedFiles(deps: GitDeps, baseCommit: string, headCom
   }
 }
 
+/** The kind of change Git reported for a file in a `--name-status` diff. */
+export type ChangedFileStatus =
+  | 'added'
+  | 'deleted'
+  | 'modified'
+  | 'renamed'
+  | 'typeChanged'
+  | 'unknown';
+
+/** A single row of a `--name-status` diff, parsed into a usable shape. */
+export interface ChangedFileWithStatus {
+  /** The kind of change Git reported. */
+  status: ChangedFileStatus;
+  /** The file's current path (the new path for renames). */
+  path: string;
+  /** The original path for renames; otherwise undefined. */
+  fromPath?: string;
+}
+
+// Maps the leading character of a `--name-status` row to a readable status. Renames carry a
+// similarity score suffix (e.g. `R097`), so we only key off the first character.
+const NAME_STATUS_CODES: Record<string, ChangedFileStatus> = {
+  A: 'added',
+  D: 'deleted',
+  M: 'modified',
+  R: 'renamed',
+  T: 'typeChanged',
+};
+
+/**
+ * Get the name-status diff of a single commit or between two, with rename detection, parsed into
+ * one object per changed file.
+ *
+ * @param deps Function dependencies.
+ * @param baseCommit The base commit to diff from.
+ * @param headCommit The head commit to diff to (empty includes uncommitted changes).
+ * @param pathspec Optional pathspec limiting the diff (e.g. a `:(glob)` magic pathspec).
+ * See: https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-pathspec for details on
+ * magic pathspecs.
+ *
+ * @returns One {@link ChangedFileWithStatus} per changed file.
+ */
+export async function getChangedFilesWithStatus(
+  deps: GitDeps,
+  baseCommit: string,
+  headCommit = '',
+  pathspec = ''
+): Promise<ChangedFileWithStatus[]> {
+  const pathspecArgument = pathspec ? ` -- "${pathspec}"` : '';
+  const output = await execGitCommand(
+    deps,
+    `git diff --name-status --find-renames=20% ${baseCommit} ${headCommit}${pathspecArgument}`
+  );
+
+  return output
+    .split(newline)
+    .filter(Boolean)
+    .map((line) => {
+      const [code, ...paths] = line.split('\t');
+      const status = NAME_STATUS_CODES[code[0]] ?? 'unknown';
+      // Rename rows carry the old path first, then the new path.
+      return status === 'renamed'
+        ? { status, fromPath: paths[0], path: paths.at(-1) as string }
+        : { status, path: paths[0] };
+    });
+}
+
 /**
  * Returns a boolean indicating whether the workspace is up-to-date (neither ahead nor behind) with
  * the remote. Returns true on error, assuming the workspace is up-to-date.
--- a/node-src/git/git.ts
+++ b/node-src/git/git.ts
@@ -272,7 +272,7 @@ export async function getChangedFilesWithStatus(
   const pathspecArgument = pathspec ? ` -- "${pathspec}"` : '';
   const output = await execGitCommand(
     deps,
-    `git diff --name-status --find-renames=20% ${baseCommit} ${headCommit}${pathspecArgument}`
+    `git --no-pager diff --name-status --no-relative --find-renames=20% ${baseCommit} ${headCommit}${pathspecArgument}`
   );
 
   return output
--- a/node-src/lib/turbosnap/classifyBailRootCause.test.ts
+++ b/node-src/lib/turbosnap/classifyBailRootCause.test.ts
@@ -1,0 +1,88 @@
+import { describe, expect, it, vi } from 'vitest';
+
+import { getChangedFilesWithStatus as getChangedFilesWithStatusDep } from '../../git/git';
+import TestLogger from '../testLogger';
+import { classifyBaselineCheckoutFailure } from './classifyBailRootCause';
+import { BaselineCheckoutFailedError, LockFileParseFailedError } from './errors';
+
+vi.mock('../../git/git');
+const getChangedFilesWithStatus = vi.mocked(getChangedFilesWithStatusDep);
+
+const deps = { log: new TestLogger(), options: {} } as any;
+
+describe('classifyBaselineCheckoutFailure', () => {
+  it('returns undefined for a non-BaselineCheckoutFailedError', async () => {
+    const result = await classifyBaselineCheckoutFailure(
+      deps,
+      new LockFileParseFailedError('/tmp/checkout-abc/yarn.lock')
+    );
+
+    expect(result).toBeUndefined();
+    expect(getChangedFilesWithStatus).not.toHaveBeenCalled();
+  });
+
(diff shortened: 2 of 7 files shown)


node-src/git/git.ts
```diff
@@ -221,6 +221,73 @@ export async function getChangedFiles(deps: GitDeps, baseCommit: string, headCom
...
+export async function getChangedFilesWithStatus(
+  deps: GitDeps,
+  baseCommit: string,
+  headCommit = '',
+  pathspec = ''
+): Promise<ChangedFileWithStatus[]> {
+  const pathspecArgument = pathspec ? ` -- "${pathspec}"` : '';
+  const output = await execGitCommand(
+    deps,
+    `git diff --name-status --find-renames=20% ${baseCommit} ${headCommit}${pathspecArgument}`
```
justin-thurman: `getChangedFiles` uses `--no-relative` in its `git diff` command. I think we need that here or running from a subdirectory would have bizarre behavior. Probably use `--no-pager` too, for consistency.

node-src/lib/turbosnap/classifyBailRootCause.ts
```diff
@@ -0,0 +1,119 @@
...
+    );
+    return { baseline_failure_kind: classifyManifestChange(changes, fileName) };
+  } catch (err) {
+    // We capture the exception higher in the stack so we can simply attach the error context here.
+    Sentry.addBreadcrumb({
+      level: 'error',
+      category: 'classifyBaselineCheckoutFailure',
+      message: 'Error classifying baseline checkout failure',
+      data: {
+        error: err,
```
justin-thurman: Any concern about serialization of the error here? Would `String(err)` or `message + stack` be better?

node-src/lib/turbosnap/classifyBailRootCause.ts
```diff
@@ -0,0 +1,119 @@
...
+  deps: GitDeps,
+  error: BaselineCheckoutFailedError
+): Promise<Record<'baseline_failure_kind', BaselineCheckoutFailureKind>> {
+  const { reference, fileName } = parsePathspec(error.pathspec);
+
+  try {
+    // Determine how the file changed between the baseline and HEAD (such as if it was added, moved,
+    // renamed, etc).
+    const basename = path.basename(fileName);
+    const changes = await getChangedFilesWithStatus(
```
justin-thurman, line 62: I _think_ this will fail if we have a shallow clone because the `git diff` call inside `getChangedFilesWithStatus` will fail. We already have a `commitExists` function. Should we use that at the start of `classifyBaselineCheckoutFailureTags` to see if the commit doesn't exist, and if it doesn't, throw a new bail reason?

I know there's been talk around shallow clones lately, so they might be handled elsewhere, making this a moot point. But I can't recall for sure, so figured I'd raise it.

justin-thurman (changes requested):
I haven't reviewed this in exhaustive detail yet, but I wanted to surface a few thoughts. I think the `--no-relative` one is blocking though. I'll circle back and review more thoroughly after you respond and/or make changes.

node-src/git/git.ts
codykaup: Oof - great catch! My test project didn't show that part because the Storybook is at the root level and not in a subdirectory like a monorepo would. 😅

node-src/lib/turbosnap/classifyBailRootCause.ts
codykaup: I _think_ it'd be fine but I split it just in case.

node-src/lib/turbosnap/classifyBailRootCause.ts
codykaup: Better to be safe! I made this change.

node-src/lib/turbosnap/classifyBailRootCause.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+      const result = await classifyTagsFromError(
+        deps,
+        new BaselineCheckoutFailedError('abc123:libs/app/package.json')
+      );
+
+      expect(result).toEqual({ baseline_failure_kind: 'baselineManifestAdded' });
+    });
+
+    it('classifies a file rename as "baselineManifestMoved" over "baselineManifestAdded" when the diff has both rows', async () => {
```
justin-thurman, line 72: For these tests, it seems like the load-bearing code is the glob stuff, but since we're mocking `getChangedFilesWithStatus`, that doesn't really get tested. Doing this without mocks is probably too heavy, but is it worth adding a line asserting the call args? That way if someone breaks/edits the call, it will at least get raised as something to consider and account for the by these tests?

Something like `expect(getChangedFilesWithStatus).toHaveBeenCalledWith(deps, 'abc123', 'HEAD', ':(glob,top)**/package.json')` I think would do it, though I haven't confirmed.

justin-thurman (approved):
Looks good! One non-blocking thought around the tests.

node-src/lib/turbosnap/classifyBailRootCause.test.ts
codykaup: That's a good suggestion. I normally don't like doing that for tests but I think it's a better option than using the real thing on a moving repo like this. I'll get that updated.

codykaup: Added in https://github.com/chromaui/chromatic-cli/pull/1391/commits/8a9fd9140df92ce352f09ac792ace4acd1b7ce55.

justin-thurman: Yeah, I'm not normally a fan of this kind of assert either, but given how important that call is, it seems worth it. Especially with the comment you added around it, so we know it's an intentionally brittle test that might need changing in the future.

codykaup: I couldn't agree more. Thanks again for bringing this up!

PR #1394, opened 2026-06-12
Skip upload, verify, and snapshot tasks when detecting no changes from TurboSnap

codykaup:
We're updating the behavior of a TurboSnap build that is identical to its ancestor (such as an unrelated change in a monorepo). Instead of going through the entire build process and uploading an identical Storybook, we're going to improve that flow. To do that, we:

1. Tell the backend service if TurboSnap didn't detect a change
2. The backend service determines if the build should be `SKIPPED` or not (not every fully TurboSnapped build can be `SKIPPED` because it's currently featured flagged and there are limitations with ancestors)
3. If the build is `SKIPPED`, we simply log and exit (no Storybook upload or snapshots 🎉)

> [!NOTE]
> This is using a temporary output setup. The designed UI will be added later.

### QA

1. In the Chromatic app, add the `account.features.useTurboSkip` feature flag on your personal account.
2. Run a build with TurboSnap enabled to establish a baseline
3. Commit a change that doesn't affect stories (such as `README.md`)
4. Run another TurboSnap build (it should skip the upload, verify, and snapshot steps)

<details>
  <summary>📦 Published PR as canary version: <code>17.8.0--canary.1394.28176252408.0</code></summary>
  <br />

  ```bash
  npm install chromatic@17.8.0--canary.1394.28176252408.0
  # or 
  yarn add chromatic@17.8.0--canary.1394.28176252408.0
  ```
</details>


node-src/tasks/upload.ts
```diff
@@ -100,6 +102,23 @@ export const waitForSentinels = async (ctx: Context, task: Task) => {
   }
 };
 
+export const finishUpload = (ctx: Context, task: Task) => {
+  // The build may have been marked SKIPPED from the backend detecting no changes from TurboSnap.
+  if (ctx.skip) {
+    ctx.log.info(buildTurboSkipped());
+
+    // Render as ↓ [skipped] rather than a ✔ checkmark to make it look like we never entered this step.
```
skitterm, line 110: Nice touch!

node-src/tasks/upload.ts
```diff
@@ -100,6 +102,23 @@ export const waitForSentinels = async (ctx: Context, task: Task) => {
...
+    setOutput('')(ctx, task); // Clear the "Starting publish" output so no subtitle shows
+
+    // Cast as a real Listr `Task` object which all have a `.skip(message)` method. This allows you
+    // to skip after a task has started.
+    (task as Listr.ListrTaskWrapper<Context>).skip('');
```
skitterm, line 115: Thanks for the description, i wouldn't have caught the nuance otherwise.

node-src/lib/upload.ts
```diff
@@ -137,6 +155,11 @@ export async function uploadBuild(
   ctx.uploadedBytes = 0;
   ctx.uploadedFiles = 0;
 
+  let turboSnapStatus = 'UNUSED';
+  if (ctx.turboSnap) {
+    turboSnapStatus = ctx.turboSnap.bailReason ? 'BAILED' : 'APPLIED';
+  }
```
skitterm: [nitpick] Now that this logic is here, should this logic be removed from the other place we're determining `turboSnapStatus` (I think it was later on in the Verify step)?

skitterm (approved):
Nice @codykaup! Not sure if this was ready for a formal review, but it looks good to me. One nitpick below, but not a blocker.

node-src/lib/upload.ts
codykaup: Great question - We'll need it in both places!

This one is used to determine if the build can be `SKIPPED` or not but the [other one](https://github.com/chromaui/chromatic-cli/blob/723883d7e11174ee02f13381527bc9187aa7e44a/node-src/tasks/verify/publishBuild.ts#L46) is for setting the exact TurboSnap status on the build (in case the build wasn't `SKIPPED` for whatever reason).

skitterm:
@codykaup just to coordinate, is this CLi version (the canary) ready for use with your changes on the Index (to use the upload step as well)? Just wanted to see if we can start using this version instead of my POC version for QA-ing and such until a stable version is published.

codykaup:
> @codykaup just to coordinate, is this CLi version (the canary) ready for use with your changes on the Index (to use the upload step as well)? Just wanted to see if we can start using this version instead of my POC version for QA-ing and such until a stable version is published.

@skitterm yep! I've been using it to test the dev changes!

codykaup:
@justin-thurman this is the PR that is going to impact what you're working on.

node-src/lib/upload.ts
```diff
@@ -185,6 +222,24 @@ export async function uploadBuild(
       return options.onError?.(new Error('Upload rejected due to user error'));
     }
 
+    // The index service may decide this build can be skipped (e.g. TurboSnap determined no story
+    // files were affected). In that case, we don't need to upload anything and can skip the
+    // remainder of the build tasks.
+    if (uploadBuild.build?.status === 'SKIPPED') {
+      ctx.log.debug('Build skipped TurboSnap, not uploading build files');
```
skitterm: [Nitpick] Might be nice to have the `buildId` in the log to differentiate it.

node-src/tasks/upload.test.ts
```diff
@@ -267,6 +271,192 @@ describe('uploadStorybook', () => {
...
+      onlyStoryFiles: [],
+    } as any;
+    await uploadStorybook(ctx, {} as any);
+
+    expect(client.runQuery).toHaveBeenCalledWith(expect.stringMatching(/PrepareBuild/), {
+      buildId: '2',
+      runtimeSpecs: [],
+      skipped: true,
+    });
+  });
```
skitterm, line 528: It'd be good to have a test case for the opposite (`prepareBuild` mutation never gets called if the build is not skipped), to ensure we aren't introducing a regression.

skitterm (approved):
@codykaup I see one test case that would be worth adding (see comment below), though that could come on a follow-up PR. Nice work!

justin-thurman:
Yes, it has already clobbered me. 🫠 I forgot to pause on this before working on `upload`. Reviewing now!

node-src/lib/upload.ts
```diff
@@ -185,6 +222,24 @@ export async function uploadBuild(
...
+
+      // Ensure we're preparing the skipped build so all build stats carry over from its
+      // ancestor
+      await ctx.client.runQuery(PrepareSkippedBuildMutation, {
+        buildId: uploadBuild.build.id,
+        runtimeSpecs: [], // Required by the API contract but we won't have any runtime specs
+        skipped: true,
+      });
```
justin-thurman: We never inspect `userErrors` here like we do for the upload build mutation. Is that something we're worried about? Seems like theoretically, we could have user errors, but we'd just set `ctx.skip` and continue.

node-src/lib/upload.ts
```diff
@@ -137,6 +167,11 @@ export async function uploadBuild(
   ctx.uploadedBytes = 0;
   ctx.uploadedFiles = 0;
 
+  let turboSnapStatus = 'UNUSED';
+  if (ctx.turboSnap) {
+    turboSnapStatus = ctx.turboSnap.bailReason ? 'BAILED' : 'APPLIED';
+  }
```
justin-thurman: Quite a few magic strings floating around this file. Worth a `type TurboSnapStatus = 'UNUSED' |  'BAILED' | ...`?

justin-thurman (approved):
Looks good. A couple non-blocking thoughts.

node-src/lib/upload.ts
codykaup: `userErrors` is specific to the upload build mutation. However, this would retry a couple times then throw which would loudly fail the run. It's not the end of the world but I think it'd be better to log (and report) the error the Sentry then continue. This mutation is for us so it failing shouldn't impact the user since there's nothing for them to fix.

node-src/lib/upload.ts
codykaup: Good call. Duplicated the type from the backend.

node-src/tasks/upload.test.ts
codykaup: Great idea - done!

node-src/lib/upload.ts
codykaup: The build ID is elsewhere in the log stream but I added it here too so it's more obvious without you needing to look around.

node-src/lib/upload.ts
codykaup: @justin-thurman WDYT? ☝️

justin-thurman: This looks good to me -- so long as there's no downside to continuing on a failing mutation. This surprises me, as I would have expected the index to need that mutation to go through for baseline purposes. But this is pure intuition on my part, I'm not familiar with the index side.

justin-thurman (approved):

node-src/lib/upload.ts
codykaup: The `prepareBuild` mutation is just for billing metrics and the like. If the CLI fails to call it successfully, there's not much we can do since the build is already marked `SKIPPED` on the backend.

codykaup: All the baseline logic for TurboSnap happens before this because it's required for TurboSnap to run. `uploadBuild` (step before this) is the hand-off spot where the CLI say "there are no changes" and the Index either confirms ("yes, let's skip everything") or denies ("run a build like normal anyway").

If we reach the `publishBuild` mutation here, the Index has decided to "skip everything" and the CLI is basically done. We're just poking the Index again to have the Index migrate a bunch of data around (nothing critical to skipping the build AFAIK).

PR #1400, opened 2026-06-18
Automatically upload metadata files on TurboSnap bails

codykaup:
We want to help with debugging TurboSnap bails without going back and forth with the customer too much because it's slow and a little annoying as a customer. This change automatically uploads the metadata files (mostly log files) to Chromatic for quicker debugging and a better support experience.

Note: You can opt out of this behavior by setting `--upload-metadata=false` or `--no-upload-metadata`.

List of changes in this PR:

1. Writes debug logs to `LOG_FILE` despite what the UI is showing
2. Writes `LOG_FILE` in case we need to upload for debugging (unless opted out)
3. Moves the question of "do you want me to add a `chromatic` script to your `package.json` since it makes more sense to happen at the end rather than before file uploads

<details>
  <summary>📦 Published PR as canary version: <code>17.7.0--canary.1400.28114016138.0</code></summary>
  <br />

  ```bash
  npm install chromatic@17.7.0--canary.1400.28114016138.0
  # or 
  yarn add chromatic@17.7.0--canary.1400.28114016138.0
  ```
</details>

### DIFF
--- a/action-src/main.ts
+++ b/action-src/main.ts
@@ -202,7 +202,7 @@ async function run() {
         storybookLogFile: maybe(storybookLogFile),
         traceChanged: maybe(traceChanged),
         untraced: maybe(untraced),
-        uploadMetadata: maybe(uploadMetadata, false),
+        uploadMetadata: maybe(uploadMetadata),
         vitest: maybe(vitest),
         zip: maybe(zip, false),
         junitReport: maybe(junitReport),
--- a/node-src/index.ts
+++ b/node-src/index.ts
@@ -30,7 +30,10 @@ import parseArguments from './lib/parseArguments';
 import { exitCodes, setExitCode } from './lib/setExitCode';
 import { uploadMetadataFiles } from './lib/uploadMetadataFiles';
 import { rewriteErrorMessage } from './lib/utilities';
-import { writeChromaticDiagnostics } from './lib/writeChromaticDiagnostics';
+import {
+  removeChromaticDiagnostics,
+  writeChromaticDiagnostics,
+} from './lib/writeChromaticDiagnostics';
 import getTasks from './tasks';
 import { Context, Flags, Options } from './types';
 import { endActivity } from './ui/components/activity';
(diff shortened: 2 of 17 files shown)


codykaup:
We decided to expand this one to all of TurboSnap so I'll make that adjustment and reopen.

action-src/main.ts
```diff
@@ -202,7 +202,7 @@ async function run() {
         storybookLogFile: maybe(storybookLogFile),
         traceChanged: maybe(traceChanged),
         untraced: maybe(untraced),
-        uploadMetadata: maybe(uploadMetadata, false),
+        uploadMetadata: maybe(uploadMetadata),
```
codykaup, line 205: Default to `undefined` instead of `false` here.

node-src/index.ts
```diff
@@ -211,17 +213,49 @@ export async function runAll(initialContext: InitialContext) {
...
+    if (ctx.options.uploadMetadata === undefined && isTurboSnapEnabled(ctx)) {
+      ctx.log.info('Uploading metadata files automatically because TurboSnap was enabled');
+    }
     await uploadMetadataFiles(ctx);
   }
+
+  // Clean up the temporary log file now that it has been uploaded.
+  if (ctx.options.logFile && !ctx.options.persistLogFile) {
+    ctx.log.removeLogFile();
+  }
```
justin-thurman: Unless I'm misreading, we might also be automatically writing the diagnostics file too, but we're not cleaning it up here. Should it be cleaned up too?

node-src/lib/log.ts
```diff
@@ -88,18 +88,31 @@ export interface Logger {
...
   disable() {
     this.append = () => {};
     this.queue = [];
+    this.stream?.end();
+    this.stream = undefined;
+  },
+  remove(onError: LogFunction) {
+    this.disable();
+    if (this.filepath) {
+      rm(this.filepath, { force: true }, (err) => {
```
justin-thurman: Isn't this async? Should we either await or use `rmSync`? I think this is only working right now because we `await Sentry.flush` in `main.ts`.

justin-thurman (changes requested):
Overall, looks good. A couple questions though that are potentially blocking. The first is cleaning up the diagnostics file (see inline comment). Second, it looks like we're now creating these log files in `share.ts` too because we call `getOptions`, but we never call `removeLogFile`. That's only called in `runAll`.

I also reviewed the Clack work, and I'm pretty sure this will just work without changes there. 🎉

node-src/lib/log.ts
codykaup: Ahh good call! I was following `initialize` but, yes, we probably want the sync version just in case!

node-src/index.ts
codykaup: Yeah we should clean that up too. I also realized we want these to persist with `--debug` too so I updated that while I was in there.

codykaup:
> Second, it looks like we're now creating these log files in `share.ts` too because we call `getOptions`, but we never call `removeLogFile`. That's only called in `runAll`.

I always forget about that code path. 🤦 

I fixed that up by adding a `logFile` argument to the API but defaulting to just not writing a log file in the first place.

codykaup:
@justin-thurman thanks for the review. I updated a bunch of stuff and it's ready for another look! Everything is in a separate commit (including updating the branch) so it's easier to see what updated.

justin-thurman (approved):
Looks good!

PR #1407, opened 2026-06-25
Don't bail on irrelevant manifest parsing failures

codykaup:
One unfortunate discovery is that we bail on any promise rejection when trying to parse _any_ manifest pair (`package.json` and lock file) in the repository. This tends to cause issues in monorepos where things update frequently.

The main scenario I tested was simply adding a new package into my test monorepo. I didn't add it to the Storybook or anything. TurboSnap bailed immediately because it attempt to checkout the baseline version of the new package's `package.json` and it failed because it didn't exist in the baseline.

To fix this, we can simply ignore any manifest pairs that are irrelevant to the Storybook. I considered a couple ways to accomplish this:

1. Flipping the `preview-stats.json` read and the manifest parsing but the changes were more substantial than I cared for (especially if we're looking at replacing this entire algorithmanyway)
2. (This PR) Ensuring the manifests exist in the baseline then bailing on any that don't have a baseline but appear in the Storybook build output

<details>
  <summary>📦 Published PR as canary version: <code>17.7.2--canary.1407.28198929320.0</code></summary>
  <br />

  ```bash
  npm install chromatic@17.7.2--canary.1407.28198929320.0
  # or 
  yarn add chromatic@17.7.2--canary.1407.28198929320.0
  ```
</details>

### DIFF
--- a/node-src/lib/turbosnap/getPackageFilesInBuild.test.ts
+++ b/node-src/lib/turbosnap/getPackageFilesInBuild.test.ts
@@ -1,0 +1,91 @@
+import { describe, expect, it } from 'vitest';
+
+import { Context, Stats } from '../../types';
+import { getPackageFilesInBuild } from './getPackageFilesInBuild';
+
+const makeContext = (rootPath: string, baseDirectory = ''): Context =>
+  ({
+    git: { rootPath },
+    storybook: { baseDir: baseDirectory },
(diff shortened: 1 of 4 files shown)


codykaup:
Discovered some edge cases which is going to expand the size of this to an unreasonable amount. 😞

PR #1412, opened 2026-07-01
Setup smoke-test directory for tests

codykaup:
Closes CAP-4690

We have some smoke tests that ensure we can run a Chromatic build against various environments. The steps build the CLI then run a build against a Storybook in a subdirectory of the repo. However, this leans on the dependencies of the repo itself and _not_ a self-contained Storybook project. This builds out the smoke test subdirectory so it has its own dependencies!

I also
1. Updated the CI smoke test workflow to run this Chromatic build in a separate directory
    a. This isolates the Storybook project's `node_modules` from the `chromatic-cli`
2. Removed a bunch of duplication in the smoke tests by using a [matrix](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations)
3. Cleaned up a bunch of legacy code in the CI pipeline that is no longer needed (renaming the dependency, for example)

> [!NOTE]
> This changes the names of our CI steps so the `required` steps will be pending. I'll update them once this is approved.

<details>
  <summary>📦 Published PR as canary version: <code>18.0.2--canary.1412.28804888459.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.0.2--canary.1412.28804888459.0
  # or 
  yarn add chromatic@18.0.2--canary.1412.28804888459.0
  ```
</details>


.github/workflows/smoke-test-action-next.yml
justin-thurman, line 1: This one is missing `YARN_ENABLE_IMMUTABLE_INSTALLS: false`, compared to `smoke-test-action.yml` and `smoke-test.yml`. That'll fail, right? But this branch only runs when we hit the merge queue, so this action hasn't run, so we haven't seen the bug?

Possibly worth extracting the `Build and pack the CLI` block into a reusable piece to avoid this kind of drift?

smoke-tests/.gitignore
justin-thurman, line 1: Should we be ignoring lockfiles here too? If I'm reading all this correctly, this is the file that travels with `cp -r $RUNNER_TEMP`. Since we then wouldn't be ignoring lockfiles, `yarn install` etc. would have uncommited changes. Seems bad with, e.g., turbosnap.

smoke-tests/pnpm-workspace.yaml
justin-thurman, line 1: Is this speculative? No `pnpm` use in the matrix.

justin-thurman (changes requested):
Looks great, thanks for doing this! 🎉 One potentially blocking fix on `YARN_ENABLE_IMMUTABLE_INSTALLS`, unless I'm misunderstanding something. But overall, 💯.

smoke-tests/pnpm-workspace.yaml
codykaup: I went back and forth on what to do with `pnpm` and apparently left it in a half state. 🫠 

Since I ended up adding the matrix, I'll just toss in a `pnpm` test there.

.github/workflows/smoke-test-action-next.yml
codykaup: Good eye! Yes, we want to disable it there too.



> Possibly worth extracting the `Build and pack the CLI` block into a reusable piece to avoid this kind of drift?

I'm hoping to remove the `action-next` flow entirely at some point so I'm okay with it for now. WDYT?

smoke-tests/.gitignore
codykaup: It's not a huge deal but, yes, we should ignore the lock files here instead of the root `.gitignore` like I'm doing now.

.github/workflows/smoke-test-action-next.yml
justin-thurman: Sounds good to me.

codykaup:
@justin-thurman made all those adjustments in separate commits for you.

Watching the CI run was a little painful so I also added a `node_modules` cache and a single "build the CLI tarball" step that gets reused throughout the matrix. 🎉

justin-thurman (approved):
Looks good, though also looks like CI might be stuck? 🤔

codykaup:
> Looks good, though also looks like CI might be stuck? 🤔

Included in the description note!

> This changes the names of our CI steps so the required steps will be pending. I'll update them once this is approved.

justin-thurman:
> Included in the description note!

🤦

PR #1417, opened 2026-07-17
Reorganize TurboSnap to return results instead of mutating `ctx`

codykaup:
TurboSnap can be hard to reason about because we read and mutate a massive context object throughout the trace. This PR focuses on pulling out the mutations so we return TurboSnap results instead. Doing this will make it easier to reason about and extend later.

I also did some clean up on the missing stats file bail because it actually just fails the build.

<details>
  <summary>📦 Published PR as canary version: <code>18.1.1--canary.1417.29849782362.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.1.1--canary.1417.29849782362.0
  # or 
  yarn add chromatic@18.1.1--canary.1417.29849782362.0
  ```
</details>

### DIFF
--- a/node-src/lib/turbosnap/index.test.ts
+++ b/node-src/lib/turbosnap/index.test.ts
@@ -175,11 +175,12 @@ describe('traceChangedFiles', () => {
       } as any;
       const result = await traceChangedFiles(ctx);
 
-      // Putting this in an if statement to avoid TS errors about type shape
-      if (result.status !== 'bailed') {
-        throw new Error(`Expected result.status to be 'bailed', but got '${result.status}'`);
-      }
-      expect(result.turboSnap.bailReason).toEqual(expectedBailReason);
+      expect(result).toMatchObject({
+        status: 'bailed',
+        turboSnap: {
+          bailReason: expectedBailReason,
+        },
+      });
       expect(captureException).toHaveBeenCalledTimes(1);
       expect(captureException).toHaveBeenCalledWith(error, {
         tags: {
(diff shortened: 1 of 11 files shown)


node-src/types.ts
```diff
@@ -549,7 +549,6 @@ interface TurboSnapBailReasonBase {
   changedStorybookFiles?: string[];
   changedStaticFiles?: string[];
   changedExternalFiles?: string[];
-  missingStatsFile?: true;
```
codykaup, line 552: This bail reason actually isn't a bail but it fails the build entirely. Therefore, we can clean it up from the bail reasons.

node-src/lib/turbosnap/index.test.ts
```diff
@@ -237,18 +279,26 @@ describe('traceChangedFiles', () => {
...
     } as any;
-    const onlyStoryFiles = await traceChangedFiles(ctx);
+    const result = await traceChangedFiles(ctx);
 
-    expect(ctx.turboSnap.bailReason).toBeUndefined();
-    expect(onlyStoryFiles).toStrictEqual(deps);
+    expect(result).toMatchObject({
+      status: 'traced',
+      onlyStoryFiles: deps,
+      turboSnap: {}, // checking there is no bailReason set
```
justin-thurman, line 287: Wouldn't this match any object? Maybe this:
```suggestion
      turboSnap: { bailReason: undefined }, // checking there is no bailReason set
```

node-src/lib/turbosnap/index.test.ts
```diff
@@ -261,9 +311,12 @@ describe('traceChangedFiles', () => {
       git: { changedFiles: ['./example.js', './package.json'], packageMetadataChanges },
       turboSnap: {},
     } as any;
-    await traceChangedFiles(ctx);
+    const result = await traceChangedFiles(ctx);
 
-    expect(ctx.turboSnap.bailReason).toBeUndefined();
+    expect(result).toMatchObject({
+      status: 'traced',
+      turboSnap: {}, // checking there is no bailReason set
```
justin-thurman, line 318: Likewise.
```suggestion
      turboSnap: { bailReason: undefined }, // checking there is no bailReason set
```

justin-thurman (approved):
Looks good. One test nitpick (assuming I'm not mistaken!)

node-src/lib/turbosnap/index.test.ts
codykaup: I actually started there but it failed because we're sending back a literal `{}`. If I set _any_ field, it'll error!

image

PR #1418, opened 2026-07-20
Isolate TurboSnap version 1.0

codykaup:
This PR just moves all existing TurboSnap files into a `lib/turbosnap/1.0` folder and sets up the entrypoint `lib/turbosnap/index.ts` file and associated tests. This is to help reduce the noise as I start building out TurboSnap 2.0 which should make it easier to review.

<details>
  <summary>📦 Published PR as canary version: <code>18.0.2--canary.1418.29846242655.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.0.2--canary.1418.29846242655.0
  # or 
  yarn add chromatic@18.0.2--canary.1418.29846242655.0
  ```
</details>


node-src/lib/turbosnap/1.0/index.ts
```diff
@@ -0,0 +1,97 @@
...
+ * @param ctx The context set when executing the CLI.
+ *
+ * @returns The trace result: skipped, bailed, or traced with the affected story files.
+ */
+// eslint-disable-next-line complexity
+export async function traceChangedFiles(ctx: Context): Promise<TraceChangedFilesResult> {
+  // The caller checks for ctx.fileInfo.statsPath before calling this function, so we can safely assert that it exists
+  // here.
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const { statsPath } = ctx.fileInfo!;
```
justin-thurman: This is pretty nitpicky, but since we lost the type narrowing due to this moving to a new function from where we called
```ts
if (!ctx.fileInfo?.statsPath) {
  ctx.turboSnap.bailReason = { missingStatsFile: true };
  throw new Error(missingStatsFile({ legacy: !nonLegacyStatsSupported }));
}
```
is there any risk some future caller could import this function and use it somewhere outside the current code path, where this invariant doesn't hold? Seems unlikely, since this is presumably a deprecated 1.0 library now, which is why I'm calling this a nitpicky comment. But we could add the same check as above to this function, or add `statsPath` as a separate parameter alongside `ctx` and then pass it directly, and skip the destructuring altogether.

justin-thurman (approved):
Looks good! One nitpicky inline comment, and a nitpicky comment in general: `1.0` strikes me as an odd directory name. `v1` feels more familiar to me personally. Definitely not blocking.

node-src/lib/turbosnap/1.0/index.ts
codykaup: My hope was to remove the `ctx` field entirely and force the caller to pass exact details. I can add this check back in just to be safe for now though.

codykaup:
> `1.0` strikes me as an odd directory name. `v1` feels more familiar to me personally.

Fair point! I was leaning on the project name but `vN` sounds better! 😂

PR #1424, opened 2026-08-03
Support more `.storybook/main` config structures

codykaup:
When we started reading the `.storybook/main` config file, we looked at some specific formats. In TurboSnap 2.0, we need to determine static directories and such so we need to expand the structures we support to read more flavors of the config file to give us a better chance at resolving the Storybook config details.

<details>
  <summary>📦 Published PR as canary version: <code>18.2.1--canary.1424.31514307173.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.2.1--canary.1424.31514307173.0
  # or 
  yarn add chromatic@18.2.1--canary.1424.31514307173.0
  ```
</details>

### DIFF
--- a/node-src/lib/getOptions.ts
+++ b/node-src/lib/getOptions.ts
@@ -315,7 +315,6 @@ export const getPartialOptions = (ctx: InitialContext): Partial<Options> => {
  * @returns An object containing parsed options
  */
 // TODO: refactor this function
-// eslint-disable-next-line complexity
 export default function getOptions(
   ctx: InitialContext,
   partialOptions = getPartialOptions(ctx)
@@ -354,17 +353,7 @@ export default function getOptions(
     process.exit(252);
   }
 
-  const { scripts } = packageJson;
-  if (typeof buildScriptName !== 'string') {
-    buildScriptName = 'build-storybook';
-    if (!scripts[buildScriptName]) {
-      const [key] =
-        Object.entries(scripts as Record<string, string>).find(([, script]) =>
-          script.startsWith('build-storybook')
-        ) || [];
-      if (key) buildScriptName = key;
-    }
-  }
+  buildScriptName = findBuildScriptName(packageJson.scripts, buildScriptName);
 
   // The missingBuildScriptName throw that previously lived here has moved to
   // storybookInfo.ts, where it is gated on !isReactNativeApp. It could not remain
@@ -373,4 +362,29 @@ export default function getOptions(
   // getStorybookMetadata reads options.buildScriptName during the storybookInfo task
   // and needs the value resolved before that task runs.
   return { ...options, buildScriptName };
+}
+
+/**
+ * Resolves the package.json script that builds Storybook.
+ *
+ * @param scripts The project's package.json scripts.
+ * @param buildScriptName The script name the user asked for, if any.
+ *
+ * @returns The user's script name, or the one that looks like a Storybook build.
+ */
+export function findBuildScriptName(
+  scripts: Record<string, string> = {},
+  buildScriptName?: string
+) {
+  if (buildScriptName) {
+    return buildScriptName;
+  }
+
+  if (scripts['build-storybook']) {
+    return 'build-storybook';
+  }
+
+  const [key] =
+    Object.entries(scripts).find(([, command]) => command.startsWith('build-storybook')) || [];
+  return key ?? 'build-storybook';
 }
--- a/node-src/lib/getStorybookInfo.test.ts
+++ b/node-src/lib/getStorybookInfo.test.ts
@@ -2,6 +2,12 @@ import TestLogger from '@cli/testLogger';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 
 import getStorybookInfo from './getStorybookInfo';
+import { getStorybookMetadata } from './getStorybookMetadata';
+
+vi.mock('./getStorybookMetadata', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('./getStorybookMetadata')>();
+  return { ...actual, getStorybookMetadata: vi.fn(actual.getStorybookMetadata) };
+});
 
 vi.useFakeTimers();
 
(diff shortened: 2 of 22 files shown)


node-src/lib/getStorybookMetadata.ts
```diff
@@ -221,47 +249,62 @@ export const findStorybookConfigFile = async (
...
   // @ts-expect-error __non_webpack_require__ is only defined when bundled with webpack, and allows us to bypass webpack's module system to require files at runtime
   // eslint-disable-next-line unicorn/prefer-module
   const r = typeof __non_webpack_require__ === 'undefined' ? require : __non_webpack_require__;
 
-  let mainConfig;
-  let v7 = false;
   try {
-    mainConfig = await r(path.resolve(configDirectory, 'main'));
-    deps.log.debug({ configDirectory, mainConfig });
+    const mainConfig = await r(path.resolve(configDirectory, 'main'));
```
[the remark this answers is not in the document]
codykaup: This is running on the user's/CI machine so they already have permissions to read those files.

node-src/lib/getOptions.ts
```diff
@@ -374,3 +363,28 @@ export default function getOptions(
...
+
+/**
+ * Resolves the package.json script that builds Storybook.
+ *
+ * @param scripts The project's package.json scripts.
+ * @param buildScriptName The script name the user asked for, if any.
+ *
+ * @returns The user's script name, or the one that looks like a Storybook build.
+ */
+export function findBuildScriptName(
```
codykaup, line 375: The change here is exporting the function to be used within tests and TurboSnap 2.0 later.

node-src/lib/getStorybookMetadata.ts
```diff
@@ -177,24 +198,31 @@ export const findBuilder = async (mainConfig, v7) => {
 // TODO: Update this when we start tracking refs within the project.json file; if refs are tracked there, we can skip this logic
 // Only used by Chromatic - surfaces Storybook refs and is used when announcing a build.
 // The refs are consumed by the MCP Addon for hosted Storybooks with composition on Chromatic.
-const findReferences = async (mainConfig, v7) => {
+const findReferences = async (mainConfig, isAstConfig) => {
   // The MCP Addon was first added within version 9; there is no need to check for older versions
-  if (!mainConfig || !v7) {
+  if (!mainConfig || !isAstConfig) {
```
justin-thurman, line 203: There's a lot going on here, so apologies if this is off-base, but it seems like `isAstConfig` is no longer the version signal that `v7` was. It just means the config was evaluated rather than read directly. Wouldn't that mean we're not safe to bail out here on `!isAstConfig` because we'd lose refs that actually do exist?

justin-thurman (commented):
I'm still in the middle of reviewing this, but I have to wrap up for the evening. Wanted to raise a question first though. I'll continue my review tomorrow.

node-src/lib/getStorybookMetadata.ts
justin-thurman: While I was digging around, I found that it looks like `refs` can be a function? That would be truthy, gets passed along and serialized somewhere in GQL? I'm not sure if that's handled or not. Just throwing it out there since you're poking around right there anyway.

codykaup: `v7` was when this parser started failing and we needed to fix it (https://github.com/chromaui/chromatic-cli/pull/747) but the underlying meaning was to flag how the file was parsed. This was just a rename as an attempt to be a little more clear about it.

> Wouldn't that mean we're not safe to bail out here on `!isAstConfig` because we'd lose refs that actually do exist?

Yep but based on the comment above, I think that's fine (we don't use refs in TurboSnap 2.0)!

```
// The MCP Addon was first added within version 9; there is no need to check for older versions
```

codykaup: > I found that it looks like refs can be a function?

I saw that too but was too afraid to fix the bug since it's been around for a while and unrelated to what I was doing. 😅

node-src/lib/getStorybookInfo.test.ts
```diff
@@ -120,7 +126,20 @@ describe('getStorybookInfo', () => {
...
   describe('with --storybook-build-dir', () => {
-    it('returns version from packageJson', async () => {
+    it('combines prebuilt metadata with static directories derived from source', async () => {
+      const ctx = getContext({
+        options: { storybookBuildDir: 'bin-src/__mocks__/normalProjectJson' },
+        packageJson: { dependencies: REACT },
+      });
+      expect(await getStorybookInfo(ctx)).toEqual({
+        builder: { name: '@storybook/builder-webpack5', packageVersion: expect.any(String) },
+        staticDir: ['static'],
```
justin-thurman: It looks like this comes from the CLI repo's static dir itself, which could be confusing if anything changes. Worth pointing at one of our fixtures so we don't have that silent coupling? I tested this locally, and it seems to do the trick:
```patch
diff --git a/node-src/lib/getStorybookInfo.test.ts b/node-src/lib/getStorybookInfo.test.ts
index 6fa68b31..cd7c466b 100644
--- a/node-src/lib/getStorybookInfo.test.ts
+++ b/node-src/lib/getStorybookInfo.test.ts
@@ -128,12 +128,18 @@ describe('getStorybookInfo', () => {
   describe('with --storybook-build-dir', () => {
     it('combines prebuilt metadata with static directories derived from source', async () => {
       const ctx = getContext({
-        options: { storybookBuildDir: 'bin-src/__mocks__/normalProjectJson' },
+        options: {
+          storybookBuildDir: 'bin-src/__mocks__/normalProjectJson',
+          storybookConfigDir: 'node-src/__mocks__/storybookMainConfig/js-cjs/.storybook',
+        },
         packageJson: { dependencies: REACT },
       });
       expect(await getStorybookInfo(ctx)).toEqual({
         builder: { name: '@storybook/builder-webpack5', packageVersion: expect.any(String) },
-        staticDir: ['static'],
+        staticDir: [
+          'node-src/__mocks__/storybookMainConfig/js-cjs/.storybook/static',
+          'node-src/__mocks__/storybookMainConfig/js-cjs/public',
+        ],
         version: expect.any(String),
       });
     });
@@ -160,12 +166,15 @@ describe('getStorybookInfo', () => {
 
     it('returns the correct metadata for Storybook 6', async () => {
       const ctx = getContext({
-        options: { storybookBuildDir: 'bin-src/__mocks__/sb6ProjectJson' },
+        options: {
+          storybookBuildDir: 'bin-src/__mocks__/sb6ProjectJson',
+          storybookConfigDir: 'node-src/__mocks__/storybookMainConfig/builder-js-esm/.storybook',
+        },
         packageJson: { dependencies: REACT },
       });
       expect(await getStorybookInfo(ctx)).toEqual({
         builder: { name: 'webpack4', packageVersion: '6.5.16' },
-        staticDir: ['static'],
         version: '6.5.16',
       });
     });
```
I didn't put much thought into which fixture to point at. There might be better ones.

justin-thurman (approved):
Looks good. One minor suggestion on tests.

node-src/lib/getStorybookInfo.test.ts
codykaup: Great eye! I reworked it in https://github.com/chromaui/chromatic-cli/pull/1424/commits/5c93ca1d7a2e1f916f3d8ebf978c5e301ea9b1a7 so all those files live in the same directory.

Great callout on this one! 👏

justin-thurman (approved):

PR #1428, opened 2026-08-06
Restore patch build functionality

### LINKED ISSUE #1423: --patch-build never checks out the merge base — publishes the current checkout as a base-branch build
**Bug:** In [`node-src/tasks/index.ts`](https://github.com/chromaui/chromatic-cli/blob/main/node-src/tasks/index.ts), the task-selection ternary has identical arms:

```ts
export const runPatchBuild = [prepareWorkspace, ...runUploadBuild, restoreWorkspace];

export default function index(ctx: Context): Listr.ListrTask<Context>[] {
  const tasks =
    ctx.options.patchHeadRef && ctx.options.patchBaseRef ? runUploadBuild : runUploadBuild;
```

The true arm should presumably be `runPatchBuild`, which is defined directly above but never used anywhere. As a result, `prepareWorkspace` / `restoreWorkspace` never run: `--patch-build=<head>...<base>` skips the merge-base lookup and checkout entirely, builds whatever is currently checked out, and publishes it labeled with the base branch's name.

**Impact:** Users following the app's "Patch build required for target branch" instructions silently publish their *feature branch's* Storybook as a *base-branch* build — creating an incorrect baseline that makes the PR then diff against itself. Reproduced with both v17.6.0 and v18.1.0 (in the compiled bundles the condition is reduced to a comma-expression no-op: `let t=(e.options.patchHeadRef&&e.options.patchBaseRef,OM)`).

**Repro:** In any repo, on a feature branch ahead of main, run:

```
npx chromatic --patch-build=<branch>...main
```

Observe the log reports `Commit '<feature-branch HEAD sha>' on branch 'main'`, there is no "Preparing your workspace" step, and no merge-base checkout happens (`git status` during the build confirms the feature branch is still checked out).

The comment above the task list ("snapshot migrated to the Clack renderer…") suggests this regressed during that migration.

jmhobbs:
In a previous bad merge we lost the patch build workflow.  This restores it adds a regression test.

The simple fix (go back to using `runPatchBuild`) no longer worked after the Clack migration so the process was moved into `runBuild` directly.

Closes chromaui/chromatic-cli#1423

<details><summary>📦 Published PR as canary version: 18.3.0--canary.1428.31514305056.0</summary>

✨ Test out this PR locally via:

```bash
npm install chromatic@18.3.0--canary.1428.31514305056.0
# or 
yarn add chromatic@18.3.0--canary.1428.31514305056.0
```

</details>

### DIFF
--- a/node-src/tasks/prepareWorkspace.ts
+++ b/node-src/tasks/prepareWorkspace.ts
@@ -1,69 +1,82 @@
-import Listr from 'listr';
-
 import { checkout, findMergeBase, getUpdateMessage, isClean, isUpToDate } from '../git/git';
 import installDependencies from '../lib/installDependencies';
-import { exitCodes, setExitCode } from '../lib/setExitCode';
-import { createTask, transitionTo } from '../lib/tasks';
-import { Context, Task } from '../types';
+import { exitCodes, TaskFailure } from '../lib/setExitCode';
+import { Context, Deps, TaskFunction } from '../types';
 import mergeBaseNotFound from '../ui/messages/errors/mergeBaseNotFound';
 import workspaceNotClean from '../ui/messages/errors/workspaceNotClean';
 import workspaceNotUpToDate from '../ui/messages/errors/workspaceNotUpToDate';
-import {
-  checkoutMergeBase,
-  initial,
-  installingDependencies,
-  lookupMergeBase,
-  pending,
-  success,
-} from '../ui/tasks/prepareWorkspace';
-import { runRestoreWorkspace } from './restoreWorkspace';
 
-export const runPrepareWorkspace = async (ctx: Context, task: Task) => {
-  const { patchHeadRef, patchBaseRef } = ctx.options;
+export type PrepareWorkspaceDeps = Pick<Deps, 'log' | 'options' | 'report'>;
+
+export interface PrepareWorkspaceInput {
+  patchHeadRef: string;
+  patchBaseRef: string;
+}
+
+export interface PrepareWorkspaceOutput {
+  mergeBase: string;
+}
+
+export const extractPrepareWorkspaceInput = (ctx: Context): PrepareWorkspaceInput => ({
+  patchHeadRef: ctx.options.patchHeadRef,
+  patchBaseRef: ctx.options.patchBaseRef,
+});
+
+export const runPrepareWorkspace: TaskFunction<
+  PrepareWorkspaceInput,
+  PrepareWorkspaceOutput,
+  PrepareWorkspaceDeps
+> = async (deps, input) => {
+  const { patchHeadRef, patchBaseRef } = input;
 
   // Make sure the git repo is in a clean state (no changes / untracked files).
-  if (!(await isClean(ctx))) {
-    setExitCode(ctx, exitCodes.GIT_NOT_CLEAN, true);
-    ctx.log.error(workspaceNotClean());
-    throw new Error('Working directory is not clean');
+  if (!(await isClean(deps))) {
+    deps.log.error(workspaceNotClean());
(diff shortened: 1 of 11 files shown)


jmhobbs:
A test run;

```bash
$ npx chromatic@18.1.1--canary.1428.31130633925.0 --patch-build=a-feature-branch...main --project-token=********09da
Need to install the following packages:
chromatic@18.1.1--canary.1428.31130633925.0
Ok to proceed? (y) y

┌  Chromatic CLI v18.1.1--canary.1428.31130633925.0
│
│  https://www.chromatic.com/docs/cli
│
◆  Prepared your workspace
│  Checked out commit 'ed391c1' on 'main'
│
◆  Authenticated with Chromatic [staging]
│  Using project token '****************09da'
│
◆  Retrieved git information
│  Commit 'ed391c1' on branch 'main'; found 1 parent build
│
◆  Collected Storybook metadata
│  Build metadata gathered
│
◆  Initialized build
│  Build 279 initialized
│
◆  Storybook built in 5 seconds
│  View build log at /*******/test-cases/build-storybook.log
│
◆  Preparation complete
│  Storybook files validated and prepared for upload
│
◆  Publish complete in 4 seconds
│  Uploaded 61 files (12.41 MB)
│
◆  Started build 279
│  View build details at https://www.staging-chromatic.com/build?appId=65c27b1ba...
│
◆  Build 279 passed!
│  Tested 12 stories across 6 components; captured 12 snapshots in 1 minute 40
│  seconds


Command failed with exit code 128: git rev-parse '"@{upstream}"'

fatal: no upstream configured for branch 'a-feature-branch'

✔ Storybook published
We found 6 components with 12 stories.
ℹ View your Storybook at https://************.staging-chromatic.com/

✔ Build 279 passed!
1 visual change was found in this build.
ℹ View build details at https://www.staging-chromatic.com/build?************

Discarding changes and restoring head location

Restored your workspace

⚠ No 'chromatic' script found in your package.json
Would you like me to add it for you? [y/N]n

ℹ No problem. You can add it to your package.json yourself like so:
"scripts": {
  "chromatic": "npx chromatic --force-rebuild --patch-build=a-feature-branch...main --project-token=****************09da
}
```

node-src/tasks/prepareWorkspace.ts
```diff
@@ -1,65 +1,82 @@
...
   } catch (err) {
-    ctx.mergeBase = undefined;
-    setExitCode(ctx, exitCodes.NPM_INSTALL_FAILED);
-    ctx.log.error(err);
-    await runRestoreWorkspace(ctx); // make sure we clean up even when something breaks
-    throw new Error('Failed to install dependencies');
+    deps.log.error(err);
+    throw new TaskFailure('Failed to install dependencies', {
+      exitCode: exitCodes.NPM_INSTALL_FAILED,
+    });
```
justin-thurman, line 76: The `finally` block in `index.ts` doesn't cover this case. It's gated on `if (ctx.mergeBase)`, but `ctx.mergeBase` is only set by `applyPrepareWorkspaceOutput`, which requires a successful task result. But if we're in this block, we've already checked out the merge base, but we won't get a `continue` result. So we need to restore separately here.
```suggestion
    deps.log.error(err);
    await runRestoreWorkspace(deps); // checkout already happened; don't strand the user
    throw new TaskFailure('Failed to install dependencies', {
      exitCode: exitCodes.NPM_INSTALL_FAILED,
    });
```

justin-thurman (changes requested):
Looks good overall, and fits the Clack paradigm. Just one missed cleanup spot that needs to be addressed.

justin-thurman (approved):

PR #1432, opened 2026-08-10
Improve Storybook main config parsing

codykaup:
Relates to CAP-4864

This PR is to isolate the Storybook main config parsing so callers don't need to know _how_ the config was parsed. The caller simply gets a return function to read the config values.

I also reorganized some helper functions to remove unnecessary exports and move them to the bottom of the file to keep the important bits at the top.

> [!NOTE]
> This now exposes `refs` for non-AST parsed config files!

---

<details>
  <summary>📦 Published PR as canary version: <code>18.4.1--canary.1432.32179171462.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.4.1--canary.1432.32179171462.0
  # or 
  yarn add chromatic@18.4.1--canary.1432.32179171462.0
  ```
</details>

### DIFF
--- a/node-src/lib/getStorybookMetadata.ts
+++ b/node-src/lib/getStorybookMetadata.ts
@@ -109,6 +109,10 @@ async function findBuilder(mainConfig?: MainConfigReader) {
 // Only used by Chromatic - surfaces Storybook refs and is used when announcing a build.
 // The refs are consumed by the MCP Addon for hosted Storybooks with composition on Chromatic.
 async function findReferences(mainConfig?: MainConfigReader) {
+  if (!mainConfig) {
+    return {};
+  }
+
   // An evaluated config can hand back a `refs` function, which we can't make sense of.
   const references = objectField(mainConfig, 'refs');
   return references ? { refs: references } : {};
@@ -358,7 +362,7 @@ async function findStorybookVersion({ env, log, options, packageJson }: Storyboo
  * @returns The field's value when it is a plain object, otherwise `undefined`.
  */
 function objectField(
-  mainConfig: MainConfigReader | undefined,
+  mainConfig: MainConfigReader,
   field: string
 ): Record<string, unknown> | undefined {
   const value = mainConfig?.readField(field);
--- a/node-src/lib/getStorybookMetadata.ts
+++ b/node-src/lib/getStorybookMetadata.ts
@@ -17,13 +17,20 @@ import { viewLayers } from './viewLayers';
 /**
  * Reads the `-c` and `-s` flags out of the project's Storybook build script.
  *
- * @param deps The resolved options and the project's package.json.
+ * @param input The project's package.json and the resolved build script name.
+ * @param input.buildScriptName The package.json script that builds Storybook.
+ * @param input.packageJson The project's package.json.
  *
  * @returns The config directory and static directories the build script names, if any.
  */
-export const findConfigFlags = async (deps: Pick<StorybookInfoDeps, 'options' | 'packageJson'>) => {
-  const { buildScriptName } = deps.options;
-  const { scripts = {} } = deps.packageJson;
+export async function findConfigFlags({
+  buildScriptName,
+  packageJson,
+}: {
+  buildScriptName?: string;
+  packageJson: StorybookInfoDeps['packageJson'];
+}) {
+  const { scripts = {} } = packageJson;
   if (!buildScriptName || !scripts[buildScriptName]) return {};
 
   const { flags } = meow({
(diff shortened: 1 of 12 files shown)


node-src/lib/getStorybookMetadata.ts
```diff
@@ -193,118 +103,131 @@ export const findBuilder = async (mainConfig, isAstConfig) => {
...
 
 // TODO: Update this when we start tracking refs within the project.json file; if refs are tracked there, we can skip this logic
 // Only used by Chromatic - surfaces Storybook refs and is used when announcing a build.
 // The refs are consumed by the MCP Addon for hosted Storybooks with composition on Chromatic.
-const findReferences = async (mainConfig, isAstConfig) => {
-  // The MCP Addon was first added within version 9; there is no need to check for older versions
-  if (!mainConfig || !isAstConfig) {
-    return {};
-  }
-
```
justin-thurman, line 206: Removing this is a behavior change, like what we discussed in a previous PR around this code. It's a _good_ change, but is it worth calling out in the PR description, so that it's clear refs are now surfaced in non-AST-parsed configs?

node-src/lib/getStorybookMetadata.ts
```diff
@@ -138,35 +45,36 @@ export const findConfigFlags = async (deps: Pick<StorybookInfoDeps, 'options' |
...
+ * @returns The builder name and package version, or an unknown builder without a config.
  */
-export const readMainConfigField = (mainConfig: any, isAstConfig: boolean, field: string) => {
-  if (!mainConfig) return undefined;
-  if (isAstConfig) return mainConfig.getSafeFieldValue([field]);
-  return mainConfig.default?.[field] ?? mainConfig[field];
-};
-
-export const findBuilder = async (mainConfig, isAstConfig) => {
+export async function findBuilder(mainConfig?: MainConfigReader) {
```
justin-thurman: Pretty sure this is no longer imported anywhere.

node-src/lib/getStorybookMetadata.ts
```diff
@@ -321,4 +244,127 @@ export const getStorybookMetadata = async (
...
+ * Reads a field only when it holds a plain object, which is all our callers can use. An evaluated
+ * config can hand back a function or a class instance where a parsed AST hands back nothing.
+ *
+ * @param mainConfig The loaded main config, or undefined when it could not be read.
+ * @param field The top-level field to read.
+ *
+ * @returns The field's value when it is a plain object, otherwise `undefined`.
+ */
+function objectField(
+  mainConfig: MainConfigReader | undefined,
```
justin-thurman: This is only used in `findBuilder` and `findReferences`. In `findBuilder`, you already bail early when mainConfig is undefined. If you added a similar bail in `findReferences`, you could narrow this param type.

node-src/lib/getStorybookMetadata.ts
```diff
@@ -193,118 +103,131 @@ export const findBuilder = async (mainConfig, isAstConfig) => {
...
     path.posix.isAbsolute(directory) ? directory : path.posix.join(safeConfigDirectory, directory)
   );
 
   return resolvedDirectories.length > 0 ? { staticDir: resolvedDirectories } : {};
-};
+}
+
+// The main config files we parse into an AST when `require()` of the config fails. Widening this
+// widens what lands on `ctx.storybook`.
+const MAIN_CONFIG_PATTERN = /^main\.[cm]?[jt]sx?$/;
```
justin-thurman: I think there's some knowledge duplication here. `uploadMetadataFiles.ts:35` has an inline regex that I think is attempting to capture the same information as this pattern, though it skips .cjs/.mjs/.mts. Should we use this const there too? Widens the scope of that function, but maybe that's correct?

justin-thurman (approved):
This is great! Really like this abstraction. A few non-blocking thoughts.

node-src/lib/getStorybookMetadata.ts
codykaup: Nice find - thank you!

node-src/lib/getStorybookMetadata.ts
codykaup: That's the delicate line I'm trying to walk. It's technically more correct but will it start bailing when it didn't previously? I'm starting to lean toward correct instead of duplicating/extending logic all over the place. WDYT?

node-src/lib/getStorybookMetadata.ts
codykaup: Great eye! Made that change.

node-src/lib/getStorybookMetadata.ts
justin-thurman: Which would be the "correctness" move here? Reusing `MAIN_CONFIG_PATTERN`?

codykaup: Yeah. Expanding it to support all the extra extensions.

justin-thurman: Then yeah, I agree. I think the drift risk here is real, and could result in some confusing debugging in our future if not addressed.

codykaup:
@justin-thurman requested another review since some things shifted around based on your feedback. Should be tiny!

justin-thurman (approved):
🎉

PR #1433, opened 2026-08-10
Resolve `ctx.storybook` paths to absolute paths

codykaup:
Relates to CAP-4864

Since TurboSnap v2 needs to eventually hash the contents of both directories, we have to go the extra mile to parse out where those directories live. Instead of isolating that behavior to v2 and duplicating logic all over the place, we're going to expand `ctx.storybook` to have more reliable output so v1 and v2 can both benefit.

This PR is a fairly substantial because it changes the contract a bit. We had relative and absolute paths all over the place and it was tough to figure out which one you had at the time. This changes that behavior so we store absolute paths and the features that need relative paths can derive them on whatever resolution root they need (for example, TurboSnap needs to be anchored to the Storybook package directory since the builder output is based on that).

---

<details>
  <summary>📦 Published PR as canary version: <code>18.5.0--canary.1433.32179165729.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.5.0--canary.1433.32179165729.0
  # or 
  yarn add chromatic@18.5.0--canary.1433.32179165729.0
  ```
</details>


node-src/lib/turbosnap/storybookDirectories.ts
```diff
@@ -0,0 +1,146 @@
...
+  buildScriptName?: string;
+}
+
+/**
+ * Reads the Storybook config and static directories out of the project's own source config.
+ *
+ * TurboSnap v2 derives these here rather than reading them off `ctx.storybook`, because that
+ * metadata resolves the directories against the cwd, and v2 hashes files relative to the Storybook
+ * project root to match the builder output.
+ *
```
justin-thurman: This is tingling my spidey-senses. So `getStorybookMetadata` has its own knowledge about how to find config/static directories. Can we reuse this helper in `getStorybookMetadata` so that there's not this knowledge duplication? I think if we change the `input.projectRoot` to be a more generic `input.root`, and then we pass project root for turbosnap v2 usage and pass cwd for storybook metadata, then we could remove a lot of stuff from `getStorybookMetadata`, including the somewhat confusing merge loop in lines 235-243, and avoid any sort of drift here.

It's also just generally confusing to me that the storybook config dir in `getStorybookMetadata` is NOT project-relative, but this config dir is. How can a user use the config dir flag effectively? There's one flag, and in `getStorybookMetadata`, it's absolute, whereas here, it's project-relative. Unless I'm just totally missing something, which is possible!

node-src/lib/turbosnap/storybookDirectories.ts
```diff
@@ -0,0 +1,146 @@
...
+  buildScriptName,
+}: StorybookDirectoriesInput): Promise<StorybookDirectories> {
+  const buildScriptFlags = await readBuildScriptFlags(projectRoot, buildScriptName, log);
+  const configDirectory = toProjectRelative(
+    projectRoot,
+    explicitConfigDirectory ?? buildScriptFlags.configDir ?? '.storybook'
+  );
+
+  const staticDirectories =
+    explicitStaticDirectories ??
```
justin-thurman: Explicit static dirs bypass `toProjectRelative` that's used in `deriveStaticDirectories`. Intentional? Worth a note if so?

node-src/lib/turbosnap/storybookDirectories.test.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    expect(directories).toEqual({ configDir: '.storybook', staticDirs: [] });
+    expect(readMainConfig).not.toHaveBeenCalled();
+  });
+
+  it("merges the build script's -s with main.*'s staticDirs, resolved against the config dir", async () => {
+    readJsonMock.mockResolvedValue({
+      scripts: { 'build-storybook': 'storybook build -s public' },
+    });
+    mockMainConfig(['./assets', { from: '../shared', to: '/shared' }]);
```
justin-thurman: I like the `../shared` case here. But it raises an interesting point. As far as I can tell, the parser will happily accept something like `../../shared`, which might end up outside of the project root. Could that cause problems? If so, should we error/log a warning/something else? In any case, a test pinning the behavior might be worthwhile.

justin-thurman (changes requested):
Overall, looks good! I do have some questions about storybook config dirs and relative vs. absolute paths. Not necessarily any code changes needed here, but I'd like to make sure I understand some of this before I approve.

node-src/lib/turbosnap/storybookDirectories.test.ts
codykaup: Great question! I just tested this out and you can point the static directory anywhere on the system but it shouldn't cause a problem for us. The Storybook build fails if the directory doesn't exist which would fail before we ever run TurboSnap.

Since all the file paths are relative, it should be fine no matter where the file lives (inside the project or not) because it shows up as `../../shared`.

I can write up a test for it quick though so it's clear that is intended behavior.

node-src/lib/turbosnap/storybookDirectories.ts
codykaup: I avoided doing that to not touch v1 as much as possible but it it's going to be rough either way. I'm playing around with this a little to see if we can just make `ctx.storybook` correct in the first place without breaking v1. 👀

codykaup:
Converting to draft so I can poke at improving `ctx.storybook` instead of deriving v1 and v2 differently.

codykaup:
Okay @justin-thurman I've been staring at this for way too long now but I think it's in a better spot. Just as you mentioned, the goal of this PR turned into improving `ctx.storybook` by
1. Expanding supported configuration shapes we can read
2. Normalizing its path fields to absolute paths and letting callers derive relative paths where required

I've been poking at this for so long, I bet I missed something obvious so I apologize in advance for that!

node-src/lib/getStorybookInfo.ts
```diff
@@ -2,47 +2,67 @@ import { pathExistsSync } from 'fs-extra';
...
+  discovered: Partial<Omit<Storybook, 'projectRoot'>>,
+  deps: StorybookInfoDeps,
+  projectRoot: AbsolutePath
+): Storybook {
+  return {
+    ...discovered,
+    projectRoot,
+    configDir:
+      discovered.configDir ??
+      path.resolve(projectRoot, deps.options.storybookConfigDir ?? '.storybook'),
```
justin-thurman: Minor nitpick here: this `deps.options.storybookConfigDir ?? '.storybook'` re-implements config directory precedence. Should we use `findConfigDirectory` here too, or is there a reason this diverges?

node-src/lib/getStorybookMetadata.ts
```diff
@@ -216,40 +248,88 @@ export async function readMainConfig(
...
+  if (!options.storybookConfigDir) {
+    return path.resolve(projectRoot, buildScriptConfigDirectory ?? '.storybook');
+  }
+
+  const candidates = [
+    path.resolve(projectRoot, options.storybookConfigDir),
+    path.resolve(options.storybookConfigDir),
+  ];
+  for (const candidate of candidates) {
+    if (await pathExists(candidate)) return candidate;
```
justin-thurman: Do you think it's worth a `log.debug` here, showing what the candidate directories are? Might aid in debugging any "why is this config dir being used?" support issues.

node-src/lib/getStorybookMetadata.ts
```diff
@@ -216,40 +248,88 @@ export async function readMainConfig(
...
  *
  * @returns The Storybook metadata, or an empty object when none could be found.
  */
-export async function getStorybookMetadata(deps: StorybookInfoDeps): Promise<Partial<Storybook>> {
-  const configDirectory = deps.options.storybookConfigDir ?? '.storybook';
+export async function getStorybookMetadata(
+  deps: StorybookInfoDeps,
+  projectRoot: AbsolutePath
+): Promise<Partial<Omit<Storybook, 'projectRoot'>>> {
+  const buildScript = findConfigFlags({
```
justin-thurman, line 264: This moved out of the `Promise.allSettled` below. Worth a try/catch? Is there anything meaningful to do on a catch?

justin-thurman (approved):
This looks great! The refactor got pretty big, but I think it's way better than it was, either on main or in your previous iteration. Very obvious now how storybook config directory discovery works. 🎉 The PR description is now a bit stale. Might be worth an update for future readers. Other than that, a few minor notes, nothing blocking.

codykaup:
> The PR description is now a bit stale. Might be worth an update for future readers.

@justin-thurman I wonder if you got a stale version of the description because I updated it already. Would you mind refreshing and seeing if it shows up? 🙏

node-src/lib/getStorybookMetadata.ts
codykaup: Good call. Added!

node-src/lib/getStorybookInfo.ts
codykaup: There's definitely some cleanup that needs to happen in here because there's duplication littered around. I've moved it to use `findConfigDirectory` to remove _some_ of that duplication but we should probably merge Storybook info with Storybook metadata so it handles everything under a single interface.

node-src/lib/getStorybookMetadata.ts
codykaup: Not a bad idea — added it to that `meow` call!

codykaup:
@justin-thurman did you want to take another peek at the latest commits?

justin-thurman (approved):
👍

PR #1441, opened 2026-08-14
[TurboSnap v2] Add story detection

codykaup:
Related to CAP-4864

This change grabs all the story files from the manifest in a similar way to TurboSnap v1. A lot of the logic is duplicated but that's on purpose. We want to keep them isolated so we can iterate on v2 in the future without worrying about breaking v1.

---

<details>
  <summary>📦 Published PR as canary version: <code>18.6.1--canary.1441.32890871950.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.6.1--canary.1441.32890871950.0
  # or 
  yarn add chromatic@18.6.1--canary.1441.32890871950.0
  ```
</details>

### DIFF
--- a/node-src/lib/turbosnap/v2/storyDetection.ts
+++ b/node-src/lib/turbosnap/v2/storyDetection.ts
@@ -90,7 +90,7 @@ export function detectStoryFiles(
     // those.
     const matchedStoryImporters = (module.reasons ?? [])
       .map((reason) => reason.moduleName)
-      .filter(Boolean)
+      .filter((name): name is FilePath => typeof name === 'string')
       .map((name) => normalizeStatsPath(name, projectRoot, statsRoot))
       .filter((importer) => storyImporters.has(importer));
 
@@ -189,7 +189,7 @@ function collectStoryImporters(
 
     const importers = (module.reasons ?? [])
       .map((reason) => reason.moduleName)
-      .filter(Boolean)
+      .filter((name): name is FilePath => typeof name === 'string')
       .map((name) => canonical(name));
 
     const importedByEntry = importers.some((importer) => entryFiles.has(importer));
--- a/node-src/lib/turbosnap/v2/storyDetection.ts
+++ b/node-src/lib/turbosnap/v2/storyDetection.ts
@@ -10,6 +10,19 @@ export interface OnDiskFiles {
   has(filePath: FilePath): boolean;
 }
 
+/**
+ * What story detection needs to resolve stats paths and tell real files apart from the
+ * require-context glob.
+ */
+interface StoryDetectionContext {
+  // The absolute Storybook project root that module paths anchor against.
+  projectRoot: AbsolutePath;
+  // The directory relative stats paths are named from.
+  statsRoot: AbsolutePath;
+  // Which canonical paths have a real file on disk.
+  onDiskFiles: OnDiskFiles;
+}
+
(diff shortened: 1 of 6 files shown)


node-src/lib/turbosnap/v2/storyDetection.ts
```diff
@@ -0,0 +1,230 @@
...
+    // An external is a graph leaf, so it can never import a story. It has no on-disk file either,
+    // which would otherwise let it through as a require-context.
+    if (EXTERNAL_MODULE.test(module.name)) {
+      continue;
+    }
+
+    const importers = (module.reasons ?? [])
+      .map((reason) => reason.moduleName)
+      .filter((name): name is FilePath => typeof name === 'string')
+      .map((name) => canonical(name));
```
justin-thurman: The syntax is a little different, but unless I'm misreading, this is the exact same loop that occurs in `detectStoryFiles`. This is a fairly complex loop, and I could see its requirements changing over time, and we'd need to remember to patch it in both places. Not to mention we're doing this work twice. Since `detectStoryFiles` is the only caller of this function, what do you think about a small loop over `status.modules` in `detectStoryFiles` that produces an array of processed stats modules that then gets passed into `collectStoryImporters`? Then we have a single point of truth for what a stats module should look like.

justin-thurman (approved):
This is really solid! It's big, but well structured and nicely isolates a lot of complexity. One suggestion for a structural tweak, non-blocking.

node-src/lib/turbosnap/v2/storyDetection.ts
codykaup: I originally built this so it's a little easier to understand each part individually. I can play around with merging those together and we'll see how gross it gets!

codykaup:
Going to put this back in draft to take a peek at a single loop in `detectStoryFiles`.

node-src/lib/turbosnap/v2/storyDetection.ts
codykaup: This was an awesome suggestion. This turned out way cleaner than it was before. Thanks @justin-thurman!

node-src/lib/turbosnap/v2/storyDetection.ts
```diff
@@ -0,0 +1,262 @@
...
+
+/**
+ * Detects which modules in the stats are story files.
+ *
+ * @param context The context required for story detection. See {@link StoryDetectionContext}.
+ * @param stats The stats file to parse.
+ *
+ * @returns The canonical paths of the story files.
+ */
+export function detectStoryFiles(context: StoryDetectionContext, stats: Stats): Set<FilePath> {
```
codykaup, line 72: In case the diff of the commit gets weird, I shifted the context required to detect story files to be the first argument.

node-src/lib/turbosnap/v2/storyDetection.ts
```diff
@@ -0,0 +1,262 @@
...
+  }
+
+  return storyFiles;
+}
+
+/**
+ * A raw stats module after the per-module work every step needs: the canonical file path, whether it
+ * has a file on disk, and its canonical importer names.
+ */
+interface ProcessedModule {
```
codykaup, line 99: @justin-thurman this is that internal type from the single parse step.

justin-thurman (approved):
Looks great! Thanks for all the back and forth on this. This came together as a really clean implementation for a really complicated feature. 🎉

PR #1442, opened 2026-08-14
[TurboSnap v2] Build the full TurboSnap manifest

codykaup:
Relates to CAP-4864

> [!NOTE]
> This is a sizable PR but all these changes felt related enough to include in a single PR. Our goal is to get this version of TurboSnap 2.0 into a testing phase and make adjustments later so I'm opening this PR a little early because it's built against a working version of the algorithm.

This builds the remaining functions to produce the TurboSnap manifest. This manifest is what is then uploaded to the Index service which determines what changed based on the baseline's version of the manifest.

It looks something like:

```
{
  "storybookHash": "<a single hash of the entire storybook>",
  "storybookFilesHashes": {
    "storybookGlobals": "<a single hash of all globals>",
    "storybookVersion": "<storybook version string>",
    "storybookConfigFiles": "<a single hash of all config files",
    "staticFiles": "<a single hash of all static files>"
  },
  "storybookConfigFiles": {
    <file path of config files>: <hash of that source file>
  },
  "staticFiles": {
    <file path of static files>: <hash of that source file>
  },
  "storyFiles": {
    <file path of story files>: <hash of that source file>
  },
  "unrecognizedStoryEntries": [],
  "attribution": {
    "storyReachable": [<list of files that link to a story>],
    "previewSubtree": [<list of files that link to .storybook/preview.ts> (or the configured dir if it's not default)],
    "storybookGlobals": [<list of files that don't link to a story but are floating in the preview-stats.json>],
  },
  "files": {
    <file path>: {
      "hash": "<hash of source file>",
      "dependencies": [<list of dependencies for the file>],
    }
    ...
  }
}
```

The main entrypoint is going to be `manifest.ts` which orchestrates all the `preview-stats.json` operations. The caller will eventually:
1. `buildManifest()`
2. `serializeManifest()`
3. `writeManifest()`

---

<details>
  <summary>📦 Published PR as canary version: <code>18.6.1--canary.1442.32890865792.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.6.1--canary.1442.32890865792.0
  # or 
  yarn add chromatic@18.6.1--canary.1442.32890865792.0
  ```
</details>

### DIFF
--- a/node-src/lib/turbosnap/v2/manifest.ts
+++ b/node-src/lib/turbosnap/v2/manifest.ts
@@ -10,12 +10,8 @@ import {
   rollUpFileHashes,
   TurboSnapFile,
 } from './graph';
-import {
-  hashOutOfGraphFiles,
-  OutOfGraphFiles,
-  OutOfGraphInput,
-  rollUpOutOfGraphFiles,
-} from './outOfGraphFiles';
+import { ManifestInput } from './manifestInput';
+import { hashOutOfGraphFiles, OutOfGraphFiles, rollUpOutOfGraphFiles } from './outOfGraphFiles';
 import { normalizeStatsPath } from './paths';
 import { ProjectFiles } from './projectFiles';
 import { readStatsGraph } from './statsGraph';
@@ -81,26 +77,16 @@ interface ManifestFile {
  * {@link readStatsGraph}'s job; everything here works in canonical paths.
  *
  * @param stats The stats file to parse.
- * @param projectRoot The absolute Storybook project root that module paths anchor against.
- * @param outOfGraph Where to find the Storybook inputs that are never bundler inputs; see
- * {@link OutOfGraphInput}.
- * @param statsRoot The absolute directory relative stats paths are named from. Defaults to the
- * project root.
+ * @param input Where the project is and what to read it with; see {@link ManifestInput}.
  *
  * @returns The manifest containing the file hashes, story file hashes, Storybook config file hashes,
  * and Storybook hash.
  */
 export async function buildManifest(
   stats: Stats,
-  projectRoot: string,
-  outOfGraph: OutOfGraphInput,
-  statsRoot = projectRoot
+  input: ManifestInput
 ): Promise<TurboSnapManifest> {
-  const { files, hashes, storyFiles } = await readStatsGraph(stats, {
-    projectRoot,
-    statsRoot,
-    projectFiles: outOfGraph.projectFiles,
-  });
+  const { files, hashes, storyFiles } = await readStatsGraph(stats, input);
 
   const { h64ToString } = await xxHashWasm();
   const storyFileHashes = new Map<FilePath, FileHash>();
@@ -119,7 +105,7 @@ export async function buildManifest(
     files,
     hashes,
     storyReachable,
-    normalizeStatsPath(outOfGraph.configDir, projectRoot),
+    normalizeStatsPath(input.configDir, input.projectRoot),
     h64ToString
   );
 
@@ -127,12 +113,12 @@ export async function buildManifest(
   // upgrade there. Track the version instead; it is a plain string, not a hash.
   storybookFileHashes.set(
     STORYBOOK_VERSION_KEY,
-    resolveStorybookVersion(projectRoot, outOfGraph.projectFiles)
+    resolveStorybookVersion(input.projectRoot, input.projectFiles)
   );
 
   // Storybook's config directory and static assets are never bundler inputs, so nothing above can see
   // them change. They get their own roll-ups; see rollUpOutOfGraphFiles.
-  const outOfGraphFiles = await hashOutOfGraphFiles(outOfGraph, projectRoot);
+  const outOfGraphFiles = await hashOutOfGraphFiles(input);
   for (const [key, hash] of rollUpOutOfGraphFiles(outOfGraphFiles, h64ToString)) {
     storybookFileHashes.set(key, hash);
   }
--- a/node-src/lib/turbosnap/v2/manifest.ts
+++ b/node-src/lib/turbosnap/v2/manifest.ts
@@ -18,7 +18,7 @@ import {
 } from './outOfGraphFiles';
 import { ProjectFiles } from './projectFiles';
 import { readStatsGraph } from './statsGraph';
-import { STORYBOOK_VERSION_KEY } from './storybookFileKeys';
+import { STORYBOOK_VERSION_KEY, StorybookFileKey } from './storybookFileKeys';
 import { collectStorybookFiles, FileAttribution } from './storybookFiles';
 import { resolveStorybookVersion } from './storybookVersion';
 
@@ -37,7 +37,7 @@ export interface TurboSnapManifest {
    * catch-all, the `storybookConfigFiles` and `staticFiles` out-of-graph sweeps, and the Storybook
    * version (the plain version string, not a hash of it).
    */
-  storybookFileHashes: Map<FilePath, FileHash | StorybookVersion>;
+  storybookFileHashes: Map<StorybookFileKey, FileHash | StorybookVersion>;
   /** Rolled-up hash per story file, covering only that story's own transitive subtree. */
   storyFileHashes: Map<FilePath, FileHash>;
   /**
@@ -50,7 +50,10 @@ export interface TurboSnapManifest {
    * `storybookConfigFiles` and `staticFiles` maps.
    */
   outOfGraphFiles: OutOfGraphFiles;
-  /** The list of all files in the Storybook. This is a parsed version of `preview-stats.json` */
+  /**
+   * The unpruned graph parsed from `preview-stats.json`, including synthetic transit nodes used by
+   * roll-ups. Synthetic nodes are omitted only when the manifest is serialized.
+   */
   files: Map<FilePath, TurboSnapFile>;
 }
 
@@ -134,16 +137,13 @@ export async function buildManifest(
     hashEntryIdentities(storyFileHashes) + hashEntryIdentities(storybookFileHashes)
   );
 
-  // Done after hashing so the graph used above is complete.
-  pruneSyntheticFiles(files, hashes);
-
   return {
-    files,
-    storyFileHashes,
-    storybookFileHashes,
     storybookHash,
+    storybookFileHashes,
+    storyFileHashes,
     attribution,
     outOfGraphFiles,
+    files,
   };
 }
 
@@ -164,9 +164,12 @@ export function serializeManifest(manifest: TurboSnapManifest): ManifestFile {
 
   const files: ManifestFile['files'] = {};
   for (const [filePath, file] of manifest.files) {
+    if (file.hash === '') continue;
     files[filePath] = {
       hash: file.hash,
-      dependencies: [...file.dependencies].sort(),
+      dependencies: [...file.dependencies]
+        .filter((dependency) => manifest.files.get(dependency)?.hash)
+        .sort(),
     };
   }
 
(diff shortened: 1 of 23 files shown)


node-src/lib/turbosnap/v2/outOfGraphFiles.ts
```diff
@@ -0,0 +1,150 @@
...
+): Promise<OutOfGraphFiles> {
+  const staticDirectories = input.staticDirs.map((directory) =>
+    path.resolve(projectRoot, directory)
+  );
+
+  const [configPaths, staticPaths] = await Promise.all([
+    input.projectFiles.listTree(path.resolve(projectRoot, input.configDir)),
+    // A file can only belong to one section, so collect the static directories first and let them win
+    // below. All the fixtures nest `.storybook/static/`, and v1 tests `isStaticFile` before
+    // `isStorybookFile` for exactly that reason (getDependentStoryFiles.ts:284-292).
```
codykaup: Delete file location

node-src/lib/turbosnap/v2/outOfGraphFiles.ts
```diff
@@ -0,0 +1,150 @@
...
+ */
+export async function hashOutOfGraphFiles(
+  input: OutOfGraphInput,
+  projectRoot: string
+): Promise<OutOfGraphFiles> {
+  const staticDirectories = input.staticDirs.map((directory) =>
+    path.resolve(projectRoot, directory)
+  );
+
+  const [configPaths, staticPaths] = await Promise.all([
```
codykaup: This part seems magical and needs to be more obvious

node-src/lib/turbosnap/v2/storybookFiles.ts
```diff
@@ -0,0 +1,121 @@
...
+  STORYBOOK_PREVIEW_KEY,
+  StorybookFileKey,
+} from './storybookFileKeys';
+
+// Matches the project-root `<configDir>/preview.*` on a canonical manifest path (whose in-project
+// keys start `./`), so a preview under node_modules is not treated as the project's preview config.
+// Path matching is the only consistent way to find it: the config-entry import edge is spelled three
+// incompatible ways across builders (vite has no such edge at all, leaving preview a detached root).
+// `configDir` is a project setting, not always `.storybook`, so the pattern is built per call.
+function previewConfigPattern(configDirectory: string): RegExp {
```
codykaup: Is there a better way to do this? If not, we need some more info on what this does

node-src/lib/turbosnap/v2/projectFiles.ts
```diff
@@ -116,37 +121,41 @@ function listFilesRecursively(
...
-    // A symlink is neither, so ask `stat`, which follows it. Anything else with no bytes of its own
-    // — a socket, a device, a broken link — contributes nothing.
+    let entries;
     try {
-      const stats = statSync(entryPath);
-      if (stats.isDirectory()) {
-        return listFilesRecursively(entryPath, visitedDirectories);
-      }
-      return stats.isFile() ? [entryPath] : [];
+      entries = readdirSync(directory, { withFileTypes: true });
```
justin-thurman, line 132: Are there any errors that might propagate up here that we would need to handle differently than just `return []`? Nothing concrete comes to mind (other than bizarre stuff like someone setting abnormal permissions on a directory), but figured I'd ask, since the blanket `catch` seems like it could swallow up some stuff.

node-src/lib/turbosnap/v2/outOfGraphFiles.ts
```diff
@@ -0,0 +1,148 @@
+import path from 'path';
+
+import { FileHash, FilePath, rollUpEntryHashes } from './graph';
+import { normalizeStatsPath } from './paths';
+import { ProjectFiles } from './projectFiles';
+import { STATIC_FILES_KEY, STORYBOOK_CONFIG_KEY, StorybookFileKey } from './storybookFileKeys';
+
+/**
+ * Where to look for the out-of-graph inputs, and what to read them with. The directories are
+ * project-root-relative, as they arrive on `ctx.storybook`.
```
justin-thurman: Isn't this not the case? I thought these paths were absolute (IIRC from one of our previous reviews). If so, should this rework to use absolute paths so we're using the same thing everywhere?

justin-thurman (commented):
I'm still in the middle of reviewing, but I'm out of time for today, so going to raise a couple questions now and get back to this tomorrow.

node-src/lib/turbosnap/v2/outOfGraphFiles.ts
codykaup: Ahh yes it should use the new absolutes. I'll get that part updated!

node-src/lib/turbosnap/v2/projectFiles.ts
codykaup: That's probably where we want to throw and handle a bail at some point but that'd be a future change. Do you think we should go ahead and throw now so this function is "proper" and doesn't lie in the output?

node-src/lib/turbosnap/v2/outOfGraphFiles.ts
codykaup: Fixed in https://github.com/chromaui/chromatic-cli/pull/1442/commits/9cc3334d5efd2804b9396229df4c758244dcf98f

node-src/lib/turbosnap/v2/manifest.ts
```diff
@@ -0,0 +1,222 @@
...
+ * @param projectFiles How to write the disk.
+ */
+export function writeManifest(
+  manifest: TurboSnapManifest,
+  outputDirectory: string,
+  projectFiles: ProjectFiles
+) {
+  projectFiles.writeFile(
+    path.join(outputDirectory, 'turbosnap-manifest.json'),
+    JSON.stringify(serializeManifest(manifest), undefined, 2)
```
jmhobbs: Do we want to have indent spacing on this file?  These can get rather large I assume and that's just extra bytes to push and store.  `jq . <` is pretty trivial for debugging.

jmhobbs: If we _do_ want indent, how about `"\t"` instead of `2`? Saves us one byte per at least.

node-src/lib/turbosnap/v2/paths.ts
```diff
@@ -5,42 +5,87 @@ import { relativeTo } from '../../getStorybookProjectRoot';
...
-    ? module.modules.map((inner) => inner.nameForCondition ?? inner.name)
-    : [module.nameForCondition ?? module.name];
-  return names.filter(Boolean);
+export function moduleFileNames(module: Module): FilePath[] {
+  const root = module.nameForCondition ?? module.name;
+  const names = [
+    root,
+    ...(module.modules ?? []).map((inner) => inner.nameForCondition ?? inner.name),
+  ];
+  return [...new Set(names.filter(Boolean).map((name) => stripConcatenatedModuleSuffix(name)))];
```
jmhobbs, line 29: That is....quite the expression 😂

jmhobbs (approved):
Approved!  Wait for Justin's too before you merge though please.  That was a heck of a PR, I need a nap now 😓

node-src/lib/turbosnap/v2/manifest.ts
codykaup: I went back and forth but I'm leaning your way now that you bring it up. These files are much more likely to _not_ be read than to actually be read by anybody. Let's save the bytes.

node-src/lib/turbosnap/v2/paths.ts
codykaup: At least it's a helper function and we won't have to look at it very often (hopefully). 😂

node-src/lib/turbosnap/v2/manifest.ts
codykaup: Done in https://github.com/chromaui/chromatic-cli/pull/1442/commits/8db36e78d02b7639c326baee8bf807005135bdf4

node-src/lib/turbosnap/v2/projectFiles.ts
justin-thurman: Not necessarily, but adding some logging here would be helpful, I think, in determining whether we even need to make the change. Right now, failure here is silent, which is scary.

node-src/lib/turbosnap/v2/manifest.ts
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * @returns The manifest containing the file hashes, story file hashes, Storybook config file hashes,
+ * and Storybook hash.
+ */
+export async function buildManifest(
+  stats: Stats,
+  projectRoot: string,
+  outOfGraph: OutOfGraphInput,
+  statsRoot = projectRoot
+): Promise<TurboSnapManifest> {
```
justin-thurman, line 88: The `projectRoot`, `statsRoot`, and `projectFiles` (from `outOfGraph.projectFiles`) move all throughout this pipeline, all in slightly different arrangements, but basically plumbing the same data around. Should this be a single type? Something like `ManifestInput`, that contains the above? You could promote `configDir` and `staticDirs` up to this object and then delete the `OutOfGraphInput` altogether (or just make it a `Pick` type that only `outOfGraphFiles.ts` cares about).

I don't _think_ this is going into god-object territory because this whole pipeline shares a single concern: building the manifest. I'm not 100% certain this is the right move, but something about `OutOfGraphInput` is bothering me. _Especially_ code outside `outOfGraphFiles.ts` reaching into that input to get `projectFiles`. That happens here in the call to `readStatsGraph`. That feels very odd. `manifestFixtures.ts` also seems to be written in a way to get around this lack of a single type, in that you have a hand-rolled `StatsContext` type, and you end up creating a stats context and an out of graph input that are very similar. Seems like a unified type could eliminate that too.

Non-blocking, of course, and I could be way off. But it's got just a bit of a smell for me here.

node-src/lib/turbosnap/v2/manifest.ts
```diff
@@ -0,0 +1,222 @@
...
+    statsRoot,
+    projectFiles: outOfGraph.projectFiles,
+  });
+
+  const { h64ToString } = await xxHashWasm();
+  const storyFileHashes = new Map<FilePath, FileHash>();
+  for (const storyFile of storyFiles) {
+    const subtree = collectTransitiveDependencies(files, storyFile);
+    storyFileHashes.set(storyFile, rollUpFileHashes(hashes, subtree, h64ToString));
+  }
```
justin-thurman, line 102: This same loop happens in `collectStorybookFiles` two lines down to calculate the `storyReachable` union. Worth just passing that union into `collectStorybookFiles` to avoid walking all the stories twice?

justin-thurman (approved):
Looks good! A couple non-blocking thoughts.

node-src/lib/turbosnap/v2/projectFiles.ts
codykaup: > Right now, failure here is silent, which is scary.

💯 there's actually no logs anywhere right now. I was going to add all that in a future PR because all of these changes were large enough without it. 😅

codykaup: I'll mark this one as resolved and I'll make sure we have a log for this one.

node-src/lib/turbosnap/v2/manifest.ts
codykaup: I tried to keep a lot of the things traveling together as a single type but I missed this one. I added it in https://github.com/chromaui/chromatic-cli/pull/1442/commits/1cb14de8cb32b9af4f4a2ed29d4fa756782c4f87 and it turned out much cleaner. Awesome suggestion here!

node-src/lib/turbosnap/v2/manifest.ts
codykaup: I originally wrote this to keep the context of what each function does together (so not passing in a bunch of pre-compiled info) then I was going to worry about performance later. I think this is still a worthy change though so I did it in https://github.com/chromaui/chromatic-cli/pull/1442/commits/747811f59a93f12b54716c5af7686f020d1cf85f.

I don't think it's too confusing of a change but I've also been staring at this for way longer than normal. 😂

justin-thurman:
Looks good! This is big, but it looks really well designed. Appreciate all the effort and the back and forth. 🙏

PR #1444, opened 2026-08-14
[TurboSnap v2] Upload TurboSnap manifest to Index

codykaup:
Relates to CAP-4864

This takes the built manifest file and uploads pieces to the Index for analysis. The Index then decides which stories need to be retested.

I also noticed some duplication when checking for files within `node_modules` directories so I cleaned that up while I was in there.

---

<details>
  <summary>📦 Published PR as canary version: <code>18.6.1--canary.1444.32890875728.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.6.1--canary.1444.32890875728.0
  # or 
  yarn add chromatic@18.6.1--canary.1444.32890875728.0
  ```
</details>

### DIFF
--- a/node-src/lib/turbosnap/v2/manifest.attribution.test.ts
+++ b/node-src/lib/turbosnap/v2/manifest.attribution.test.ts
@@ -242,8 +242,8 @@ describe('buildManifest attribution closure', () => {
     const after = await buildManifest(stats, input);
 
     // Editing the inner file used to leave the manifest byte-identical.
-    expect(after.storybookFileHashes.get(globalsKey)).not.toBe(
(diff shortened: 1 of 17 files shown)


jmhobbs (approved):
Looks good. My only question is around the new `NODE_MODULES_SEGMENT` value. That regex is pretty dense now, could we get some coverage on it? Would also help document it as well.

codykaup:
> Looks good. My only question is around the new `NODE_MODULES_SEGMENT` value. That regex is pretty dense now, could we get some coverage on it? Would also help document it as well.

Great call — added those!

PR #1459, opened 2026-08-21
[TurboSnap v2] Add logging

codykaup:
Closes CAP-4982

Just as the title says: This adds some logging around TurboSnap v2.

FYI, I think I need to take a closer look at `projectFiles.ts` later because it returns an empty array in case it's a misconfigured static directory or symlink. Those should not be errors but failing to read some directories/files _should_ be failures because an incomplete tree cannot be trusted.

---

<details>
  <summary>📦 Published PR as canary version: <code>18.5.1--canary.1459.32530862424.1</code></summary>
  <br />

  ```bash
  npm install chromatic@18.5.1--canary.1459.32530862424.1
  # or 
  yarn add chromatic@18.5.1--canary.1459.32530862424.1
  ```
</details>

### DIFF
--- a/bin-src/register.js
+++ b/bin-src/register.js
@@ -10,6 +10,8 @@ const commands = {
   'react-native-build': () =>
     import('./reactNativeBuild').then(({ main }) => main(process.argv.slice(3))),
   trace: () => import('./trace').then(({ main: traceMain }) => traceMain(process.argv.slice(3))),
+  'turbosnap-manifest': () =>
+    import('./turbosnapManifest').then(({ main }) => main(process.argv.slice(3))),
   'trim-stats-file': () =>
     import('./trimStatsFile').then(({ main: trimMain }) => trimMain(process.argv.slice(3))),
 };
--- a/bin-src/turbosnapManifest.test.ts
+++ b/bin-src/turbosnapManifest.test.ts
@@ -1,0 +1,149 @@
+import { buildManifest, serializeManifest } from '@cli/turbosnap/v2/manifest';
+import { realProjectFiles } from '@cli/turbosnap/v2/projectFiles';
+import { existsSync } from 'fs';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { getRepositoryRoot } from '../node-src/git/git';
+import TestLogger from '../node-src/lib/testLogger';
+import { readStatsFile } from '../node-src/tasks/readStatsFile';
+import { main } from './turbosnapManifest';
+
+const testLogger = new TestLogger();
+const manifest = { storybookHash: 'abc' };
+const serialized = { storybookHash: 'abc', storyFiles: {} };
+const projectFiles = { readFile: vi.fn() };
+
+vi.mock('fs', async (importOriginal) => ({
+  ...(await importOriginal<typeof import('fs')>()),
+  existsSync: vi.fn(() => true),
+}));
+
+vi.mock('@cli/turbosnap/v2/manifest', () => ({
+  buildManifest: vi.fn(() => manifest),
+  serializeManifest: vi.fn(() => serialized),
+}));
+
+vi.mock('@cli/turbosnap/v2/projectFiles', () => ({
+  realProjectFiles: vi.fn(() => projectFiles),
+}));
+
+vi.mock('../node-src/git/git', () => ({
+  getRepositoryRoot: vi.fn(() => '/repo'),
+}));
+
+vi.mock('../node-src/tasks/readStatsFile', () => ({
+  readStatsFile: vi.fn(() => ({ modules: [] })),
+}));
+
+vi.mock('../node-src/lib/log', () => ({
+  createLogger: vi.fn(() => testLogger),
+}));
+
+beforeEach(() => {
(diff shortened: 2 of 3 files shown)


justin-thurman:
@codykaup, the stack below this PR doesn't compile. I know if the stack merges together, that doesn't matter much, but should some of the changes from this PR be pushed down the stack to make #1457 compile?

Agent-driven notes on what would have to move down:
origin/cody/-ts-v2-run-with-v1 does not compile:

node-src/lib/turbosnap/index.ts(38,31): error TS2345: ... not assignable to TraceChangedFilesInput
node-src/lib/turbosnap/index.ts(46,21): error TS2554: Expected 1 arguments, but got 0.
node-src/lib/turbosnap/v2/projectFiles.test.ts(271,5): error TS2554: Expected 1 arguments, but got 0.
plus bin-src/turbosnapManifest.test.ts fails there (1 failed / 15 passed).

This PR is what fixes them. So four of its seven changed files — log: in the traceChangedFilesV2 input, realProjectFiles(ctx.log), and both test updates — are really 1457's work. If 1457 merges alone, main is red. Move them down a commit.

node-src/lib/turbosnap/v2/index.ts
```diff
@@ -56,62 +56,49 @@ export async function traceChangedFiles(
...
+      'Failed to build manifest for TurboSnap v2; this does not affect TurboSnap v1',
+      error
+    );
     Sentry.captureException(error, {
       fingerprint: ['TurboSnap v2', 'Failed to build manifest'],
     });
     return { status: 'fallback' };
   }
-  logManifestOutput(input.log, manifest);
+  input.log.debug('Generated manifest for TurboSnap v2');
```
justin-thurman: Worth adding the storybook hash to this line? Nothing else logs that that I can see.

node-src/lib/turbosnap/v2/manifest.ts
```diff
@@ -109,6 +109,14 @@ export async function buildManifest(
     normalizeStatsPath(input.configDir, input.projectRoot),
     h64ToString
   );
+  if (attribution.previewSubtree.size > 0) {
+    input.log.debug(`Found preview config subtree with ${attribution.previewSubtree.size} files`);
+  }
+  if (attribution.storybookGlobals.size > 0) {
+    input.log.debug(
+      `Found ${attribution.storybookGlobals.size} global files not linked to a story`
+    );
```
justin-thurman: Is the `size > 0` guard right here? Seems like this would hide a plausible failure mode where some misconfiguration resulted in, e.g., an empty preview subtree.

node-src/lib/turbosnap/v2/projectFiles.ts
```diff
@@ -126,7 +126,7 @@ function listFilesRecursively(
   try {
     realDirectory = realpathSync(directory);
   } catch (error) {
-    log.warn(`Failed to resolve ${directory} when listing files`, error);
+    log.debug(`Failed to resolve ${directory} when listing files`, error);
```
justin-thurman: Is demoting these to debug the right call? I know it doesn't matter currently, because we fall back to V1, but if we intend to make V2 the only path eventually, this seems like it could bite quietly.

justin-thurman (commented):
Overall, looks good. A few questions, the big one being about the bottom of the stack being red.

codykaup:
> @codykaup, the stack below this PR doesn't compile. I know if the stack merges together, that doesn't matter much, but should some of the changes from this PR be pushed down the stack to make #1457 compile?

What happened?! I literally tested it against the `main` version of the backend to make sure all the fields were there... I'll take a peek at it and get that version working. Thanks for the heads up!

node-src/lib/turbosnap/v2/projectFiles.ts
codykaup: I've gone back and forth on this (hence the diff) but I'm almost leaning the real log levels now. Right now, there's no indication on if v2 is ultimately used by the backend or not so we should just assume it's running to play it safe. I'll update that!

node-src/lib/turbosnap/v2/index.ts
codykaup: We could but I'm not sure how useful that is. I originally had all of those hashes log out but it's pretty useless without the full manifest file to understand what contributed to the hashes.

node-src/lib/turbosnap/v2/manifest.ts
codykaup: An empty preview subtree isn't abnormal but I believe we should almost always have _some_ globals. Are you thinking we just always log so it's clear what we found?

node-src/lib/turbosnap/v2/index.ts
justin-thurman: Fair enough!

node-src/lib/turbosnap/v2/manifest.ts
justin-thurman: In my opinion, yes. Debug logs are meant to be at least somewhat noisy. And let's be honest: Claude doesn't care, and no one else is reading them.

PR #1462, opened 2026-08-24
Support PNPM when installing dependencies

codykaup:
We used an older package (`yarn-or-npm`) to install dependencies for patch builds. As the name would suggest, this doesn't support PNPM so this PR changes that!

<details>
  <summary>📦 Published PR as canary version: <code>18.7.0--canary.1462.32890848472.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.7.0--canary.1462.32890848472.0
  # or 
  yarn add chromatic@18.7.0--canary.1462.32890848472.0
  ```
</details>

### DIFF
--- a/node-src/lib/installDependencies.ts
+++ b/node-src/lib/installDependencies.ts
@@ -5,7 +5,7 @@ import { runCommand } from './shell/shell';
 /**
  * Install dependencies using the package manager specified in the project's package.json.
  *
- * @returns A promise that resolves when the command completes.
+ * @returns The result of the install command.
  */
 export async function installDependencies() {
   const command = await getCliCommand(parseNi, [], { programmatic: true });
@@ -13,5 +13,5 @@ export async function installDependencies() {
     throw new Error('Unable to determine the package manager install command');
   }
 
-  return runCommand(command);
+  return runCommand(command, { timeout: 10 * 60 * 1000 });
 }
--- a/node-src/lib/installDependencies.ts
+++ b/node-src/lib/installDependencies.ts
@@ -1,10 +1,17 @@
-import { parseNi, run } from '@antfu/ni';
+import { getCliCommand, parseNi } from '@antfu/ni';
+
+import { runCommand } from './shell/shell';
 
 /**
  * Install dependencies using the package manager specified in the project's package.json.
  *
  * @returns A promise that resolves when the command completes.
  */
-export function installDependencies() {
-  return run(parseNi, [], { programmatic: true });
+export async function installDependencies() {
+  const command = await getCliCommand(parseNi, [], { programmatic: true });
+  if (!command) {
+    throw new Error('Unable to determine the package manager install command');
+  }
+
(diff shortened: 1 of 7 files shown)


node-src/lib/installDependencies.ts
```diff
@@ -1,22 +1,10 @@
...
-  });
-
-export default installDependencies;
+/**
+ * Install dependencies using the package manager specified in the project's package.json.
+ *
+ * @returns A promise that resolves when the command completes.
+ */
+export function installDependencies() {
+  return run(parseNi, [], { programmatic: true });
```
jmhobbs: It looks like [`run`](https://github.com/antfu-collective/ni/blob/33c16cd18fba6ccaf40751de173a67409a06d95a/src/runner.ts#L102) will stream output now, instead of collecting it for the resolve/reject like we did before.  Will that cause an issue for the consumers?

jmhobbs (approved):
Overall looks good, just one question!

node-src/lib/installDependencies.ts
codykaup: Woah nice catch! I was testing this in `--debug` but, sure enough, it logs all the `pnpm install` output directly to the interactive UI. 🫠 

Fixed in https://github.com/chromaui/chromatic-cli/pull/1462/commits/e7fb25432f2c7eb14b287b00b487dc16fc1f8c9c by using our function for running CLI commands to avoid this situation altogether. I should've done that to start with but figured `return run()` was cleaner (it wasn't).

codykaup: I also added a timeout so the install command doesn't silently hang indefinitely. Then I updated the tests a little so they're closer to the behavior we want!

jmhobbs: 🎉

node-src/lib/installDependencies.ts
```diff
@@ -1,22 +1,17 @@
...
+ *
+ * @returns The result of the install command.
+ */
+export async function installDependencies() {
+  const command = await getCliCommand(parseNi, [], { programmatic: true });
+  if (!command) {
+    throw new Error('Unable to determine the package manager install command');
+  }
+
+  return runCommand(command, { timeout: 10 * 60 * 1000 });
```
jmhobbs, line 16: Not a blocker, but most (all?) of our other operations (esp. git) have an environment variable configurable timeout.  Should we just go ahead and add one now instead of later?  10 minutes is plenty of time, but folks might actually want to turn it down rather than up.  Your call, might be a YAGNI.

jmhobbs (approved):

node-src/lib/installDependencies.ts
codykaup: I considered it but chopped it up to a YAGNI. Especially since TurboSnap v2 doesn't use this path at all.

jmhobbs: Cool. Send it!

PR #1463, opened 2026-08-24
TurboSnap v2 release (runs alongside v1)

codykaup:
Closes CAP-4979

This PR merges PRs #1457 and #1459 together since `gh stack` closed them for some reason.

This PR ultimately:
1. Runs TurboSnap v2 alongside v1
2. Adds some logging throughout the v2 pipeline at the proper levels (despite not impacting the results unless flagged into v2 on the Index service)

> [!NOTE]
> We're uploading the manifest file to S3 no matter what. We don't detect if the manifest was actually written in the run (so a previous build's manifest can show up on the latest build). I decided to ignore this for now since no other metadata file handles this case either and it's a much larger change.

<details>
  <summary>📦 Published PR as canary version: <code>18.8.0--canary.1463.33183522467.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.8.0--canary.1463.33183522467.0
  # or 
  yarn add chromatic@18.8.0--canary.1463.33183522467.0
  ```
</details>

### DIFF
--- a/node-src/lib/turbosnap/index.ts
+++ b/node-src/lib/turbosnap/index.ts
@@ -24,8 +24,7 @@ export async function traceChangedFiles(ctx: Context): Promise<TraceChangedFiles
   if (!ctx.fileInfo?.statsPath) {
     // If we don't know the SB version, we should assume we don't support `--stats-json`
     const nonLegacyStatsSupported =
-      ctx.storybook?.version &&
-      semver.gte(semver.coerce(ctx.storybook.version) || '0.0.0', '8.0.0');
+      ctx.storybook.version && semver.gte(semver.coerce(ctx.storybook.version) || '0.0.0', '8.0.0');
 
     throw new Error(missingStatsFile({ legacy: !nonLegacyStatsSupported }));
   }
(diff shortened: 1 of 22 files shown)


codykaup:
FYI added support for rebuilds. Previously, it would skip TurboSnap entirely. Instead, we still want to run v2 to keep the manifest updated but the Index won't check what files changed.

node-src/lib/turbosnap/index.ts
```diff
@@ -15,15 +20,46 @@ import { traceChangedFiles as traceChangedFilesV1 } from './v1';
  */
 export async function traceChangedFiles(ctx: Context): Promise<TraceChangedFilesResult> {
   if (!ctx.turboSnap || ctx.turboSnap.unavailable) return { status: 'skipped' };
-  if (!ctx.git.changedFiles) return { status: 'skipped' };
```
jmhobbs, line 18: Should this have stayed up here?  What if git info bails, like on a first build? Should we even bother with all the other work below?

jmhobbs (approved):
One question, but otherwise 👍🏻

node-src/lib/turbosnap/index.ts
codykaup: Great question! Nope, this needs to be moved like it is. `ctx.git.changedFiles` is only used in v1 so it shouldn't impact v2. Even if we're not going to use the results (first build, rebuild, etc.), then we need to upload the current state to the Index so that future builds have something to compare against once TurboSnap is actually used.

PR #1465, opened 2026-08-25
fix: `chromaui/action` to resolve correct E2E package when multiple are installed

AriPerkkio:
- Closes SB-1674.
- Closes TE-1.

Fix verified by @jonniebigodes in the ticket above.

If user has multiple npm packages installed that define `build-archive-storybook`, Chromatic CLI may pick the wrong one when it does `npm exec -- build-archive-storybook`. Npm has support for defining package name as `npm exec --package=@chromatic-com/playwright -- build-archive-storybook`, but even that doesn't work. The tickets above have minimal reproductions if you'd like to try running the commands yourself.

To fix this we'll check if we can resolve path to the installed package and call it directly. If it's not installed, we fallback to the old logic.

First commit adds more tests to `getE2EBuildCommand`, second one applies the fix.

<details>
  <summary>📦 Published PR as canary version: <code>18.7.2--canary.1465.33192636470.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.7.2--canary.1465.33192636470.0
  # or 
  yarn add chromatic@18.7.2--canary.1465.33192636470.0
  ```
</details>

### DIFF
--- a/node-src/lib/e2e.test.ts
+++ b/node-src/lib/e2e.test.ts
@@ -1,10 +1,44 @@
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, onTestFinished, vi } from 'vitest';
 
 import { getE2EBuildCommand } from './e2e';
 import { exitCodes } from './setExitCode';
 import TestLogger from './testLogger';
+import { patchModulePath } from './testUtilities';
+
+const PLAYWRIGHT_BUILD_ARCHIVE_BINARY = '@chromatic-com/playwright/bin/build-archive-storybook';
+
+// Mock @antfu/ni to detect package manager as npm
+vi.mock(import('@antfu/ni'), async (importOriginal) => ({
+  ...(await importOriginal()),
+  getCliCommand: async (runner, args, options) => runner('npm', args, options),
+}));
 
 describe('getE2EBuildCommand', () => {
+  const deps = { options: { inAction: false } as any, log: new TestLogger() };
+
+  it("can resolve E2E package when it's installed", async () => {
+    const expectedBinary = '/path/to/playwright/bin/build-archive-storybook';
+    const restore = patchModulePath(PLAYWRIGHT_BUILD_ARCHIVE_BINARY, expectedBinary);
+    onTestFinished(restore);
+
+    const command = await getE2EBuildCommand(deps, 'playwright', ['--output-dir=./source-dir/']);
+
+    expect(command).toBe(`node ${expectedBinary} --output-dir=./source-dir/`);
+  });
+
+  it('throws original package resolving errors', async () => {
+    const restore = patchModulePath(
+      PLAYWRIGHT_BUILD_ARCHIVE_BINARY,
+      'any',
+      new Error('ERR_PACKAGE_PATH_NOT_EXPORTED')
+    );
+    onTestFinished(restore);
+
+    await expect(
+      getE2EBuildCommand(deps, 'playwright', ['--output-dir=./source-dir/'])
+    ).rejects.toThrow('ERR_PACKAGE_PATH_NOT_EXPORTED');
+  });
+
   it('throws a TaskFailure with the MISSING_DEPENDENCY exit code when the E2E package is not installed', async () => {
     const deps = { options: { inAction: false } as any, log: new TestLogger() };
 
@@ -14,6 +48,16 @@ describe('getE2EBuildCommand', () => {
       name: 'TaskFailure',
       exitCode: exitCodes.MISSING_DEPENDENCY,
       userError: true,
+    });
+  });
+
+  describe('in action', () => {
+    const deps = { options: { inAction: true } as any, log: new TestLogger() };
+
+    it('invokes E2E package via package manager', async () => {
+      const command = await getE2EBuildCommand(deps, 'playwright', ['--output-dir=./source-dir/']);
+
+      expect(command).toBe('npm exec -- build-archive-storybook --output-dir=./source-dir/');
     });
   });
 });
--- a/node-src/lib/e2e.test.ts
+++ b/node-src/lib/e2e.test.ts
@@ -60,24 +60,15 @@ describe('getE2EBuildCommand', () => {
         patchModulePath(PLAYWRIGHT_BUILD_ARCHIVE_BINARY, '/path/to/playwright-bin'),
         patchModulePath(CYPRESS_BUILD_ARCHIVE_BINARY, '/path/to/cypress-bin'),
       ];
-      onTestFinished(() => void cleanups.reverse().map((cleanup) => cleanup()));
+      onTestFinished(() => {
+        for (const cleanup of cleanups.reverse()) {
+          cleanup();
+        }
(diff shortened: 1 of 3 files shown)


node-src/lib/e2e.ts
```diff
@@ -39,15 +42,30 @@ export async function getE2EBuildCommand(
...
+    // Try resolving the binary from working directory first, then fallback to package manager.
+    try {
+      const projectRequire = createRequire(path.join(process.cwd(), 'package.json'));
+      return [
+        'node',
+        projectRequire.resolve(`${dependencyName}/bin/${buildBinName}`),
+        ...buildCommandOptions,
+      ].join(' ');
+    } catch (err) {
+      if (err.code !== 'MODULE_NOT_FOUND') {
```
codykaup: Do we care about the specific error code coming back? Should we always fallback on any error?

Also, should we log it so debug logs say which path it took?

node-src/lib/e2e.ts
```diff
@@ -39,15 +42,30 @@ export async function getE2EBuildCommand(
...
   buildCommandOptions: string[]
 ) {
-  // The action cannot "peer depend" on or import anything. So instead, we must attempt to exec
-  // the binary directly.
+  const dependencyName = `@chromatic-com/${flag}`;
+
   if (deps.options.inAction) {
+    // Try resolving the binary from working directory first, then fallback to package manager.
+    try {
+      const projectRequire = createRequire(path.join(process.cwd(), 'package.json'));
```
codykaup, line 50: Do we need to account for monorepos that may be running at the root of the repo?

node-src/lib/e2e.test.ts
```diff
@@ -16,4 +51,39 @@ describe('getE2EBuildCommand', () => {
...
+
+  describe('in action', () => {
+    const deps = { options: { inAction: true } as any, log: new TestLogger() };
+
+    it('can resolve E2E package when multiple are installed', async () => {
+      const cleanups = [
+        patchModulePath(PLAYWRIGHT_BUILD_ARCHIVE_BINARY, '/path/to/playwright-bin'),
+        patchModulePath(CYPRESS_BUILD_ARCHIVE_BINARY, '/path/to/cypress-bin'),
+      ];
+      onTestFinished(() => void cleanups.reverse().map((cleanup) => cleanup()));
```
codykaup: Although clever, we don't do this `void` thing anywhere else. Should be be more clear instead?


```suggestion
      onTestFinished(() => {
        for (const cleanup of cleanups.reverse()) cleanup();
      });
```

node-src/lib/testUtilities.ts
```diff
@@ -9,16 +9,21 @@ import { Module } from 'node:module';
  *
  * @param from Input of `require.resolve(<from>)`
  * @param to Output of the mocked call, as in `const to = require.resolve(<from>)`
+ * @param error
```
codykaup: Need to fill out the JSDoc description here!

node-src/lib/e2e.test.ts
```diff
@@ -1,10 +1,45 @@
...
+  const deps = { options: { inAction: false } as any, log: new TestLogger() };
+
+  it("can resolve E2E package when it's installed", async () => {
+    const expectedBinary = '/path/to/playwright/bin/build-archive-storybook';
+    const restore = patchModulePath(PLAYWRIGHT_BUILD_ARCHIVE_BINARY, expectedBinary);
+    onTestFinished(restore);
+
+    const command = await getE2EBuildCommand(deps, 'playwright', ['--output-dir=./source-dir/']);
+
+    expect(command).toBe(`node ${expectedBinary} --output-dir=./source-dir/`);
```
codykaup, line 27: Should we be checking if `getCliCommand` was called to verify we're able to `require()` it?

node-src/lib/e2e.ts
```diff
@@ -39,15 +42,30 @@ export async function getE2EBuildCommand(
...
+        ...buildCommandOptions,
+      ].join(' ');
+    } catch (err) {
+      if (err.code !== 'MODULE_NOT_FOUND') {
+        throw err;
+      }
+    }
+
+    // The action cannot "peer depend" on or import anything. So instead, we must attempt to exec the binary directly.
+    // This will fail if user has both `@chromatic-com/playwright` and `@chromatic-com/cypress` installed.
```
codykaup: Are `@chromatic-com/playwright` and `@chromatic-com/cypress` examples? Could this also fail if `@chromatic-com/vitest` and `@chromatic-com/playwright` are installed?

node-src/lib/e2e.ts
```diff
@@ -39,15 +42,30 @@ export async function getE2EBuildCommand(
   flag: 'playwright' | 'cypress' | 'vitest',
   buildCommandOptions: string[]
 ) {
-  // The action cannot "peer depend" on or import anything. So instead, we must attempt to exec
-  // the binary directly.
+  const dependencyName = `@chromatic-com/${flag}`;
+
   if (deps.options.inAction) {
```
codykaup, line 47: Weird question but do we need to special-case `inAction`? Seems like the flow should be roughly the same for the CLI and the action?

node-src/lib/e2e.ts
AriPerkkio: Hmm, good point. Maybe we'll just fallback to old logic always. Let's remove this handling completely.

node-src/lib/e2e.ts
AriPerkkio: I don't think there's reliable way for that. 🤔 
We'll just need to fallback to the old logic if that's the case.

node-src/lib/e2e.test.ts
AriPerkkio: The `getCliCommand` is only used by `inAction: true` tests, and affects the output of `getE2EBuildCommand` in a way that we don't need to assert what was called internally.

node-src/lib/e2e.ts
AriPerkkio: Let's rephrase this to:

```suggestion
    // This will fail if user has multiple packages that define identically named `build-archive-storybook` bin.
```

node-src/lib/e2e.ts
AriPerkkio: `inAction` is the only code branch that can end up calling `build-archive-storybook` via `npm exec`, `pnpm exec` etc. I don't have the history why this is built in such way, but it's something to do with:

https://github.com/chromaui/chromatic-cli/blob/3b4eaf63ed5b25742fdca8c222532daec53ce1c7/node-src/lib/e2e.ts#L42-L43

node-src/lib/e2e.ts
codykaup: Alrighty! I'm fine shipping it, as-is. It may come out to be a YAGNI anyway.

node-src/lib/e2e.ts
codykaup: Weird 🫠

codykaup (approved):

PR #1468, opened 2026-08-27
fix: descriptive TurboSnap errors in Vitest runs

AriPerkkio:
- Fixes TE-4

If user runs `chromatic --vitest --only-changed` but hasn't had `chromaticPlugin({ turboSnap: true })` in the previous `vitest run` run, they currently see following message:

image

After this PR they see:

image

- [x] @jonniebigodes to verify preview build works as expected

<details>
  <summary>📦 Published PR as canary version: <code>18.7.4--canary.1468.34369807992.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.7.4--canary.1468.34369807992.0
  # or 
  yarn add chromatic@18.7.4--canary.1468.34369807992.0
  ```
</details>

### DIFF
--- a/node-src/lib/checkStorybookBaseDirectory.ts
+++ b/node-src/lib/checkStorybookBaseDirectory.ts
@@ -5,6 +5,7 @@ import path from 'path';
 import { getRepositoryRoot } from '../git/git';
 import { Context, Stats } from '../types';
 import { invalidStorybookBaseDirectory } from '../ui/messages/errors/invalidStorybookBaseDirectory';
+import { turboSnapDisabledVitest } from '../ui/messages/errors/turboSnapDisabledVitest';
 import { exitCodes, setExitCode } from './setExitCode';
 
 /**
@@ -24,10 +25,19 @@ export async function checkStorybookBaseDirectory(ctx: Context, stats: Stats) {
   const { storybookBaseDir: storybookBaseDirectory = path.relative(repositoryRoot, '') } =
     ctx.options;
 
-  // Find all js(x)/ts(x) files in stats that are not in node_modules
+  // Find all js(x)/ts(x) files in stats that are not in node_modules, and are not Storybook's virtual files
   const sourceModuleFiles = stats.modules.filter(
-    (module: any) => !module.name.includes('node_modules') && /\.(js|jsx|ts|tsx)$/.test(module.name)
+    (module: any) =>
+      !module.name.includes('node_modules') &&
+      module.name !== './storybook-stories.js' && // builder-webpack5 virtual file
+      /\.(js|jsx|ts|tsx)$/.test(module.name)
   );
+
+  // This can happen if `vitest run` is run without the chromaticPlugin({ turboSnap: true })
+  // and `chromatic --vitest --only-changed` is run afterwards.
+  if (ctx.options.vitest && sourceModuleFiles.length === 0) {
+    throw new Error(turboSnapDisabledVitest());
+  }
 
   // GitHub Actions seems to have a default ulimit of 1024, so we limit concurrency to stay under
   const limitConcurrency = pLimit(1000);
--- a/node-src/lib/checkStorybookBaseDirectory.ts
+++ b/node-src/lib/checkStorybookBaseDirectory.ts
@@ -36,6 +36,7 @@ export async function checkStorybookBaseDirectory(ctx: Context, stats: Stats) {
   // This can happen if `vitest run` is run without the chromaticPlugin({ turboSnap: true })
   // and `chromatic --vitest --only-changed` is run afterwards.
   if (ctx.options.vitest && sourceModuleFiles.length === 0) {
+    setExitCode(ctx, exitCodes.INVALID_OPTIONS, true);
     throw new Error(turboSnapDisabledVitest());
   }
 
--- a/node-src/ui/messages/errors/turboSnapDisabledVitest.ts
+++ b/node-src/ui/messages/errors/turboSnapDisabledVitest.ts
@@ -1,0 +1,15 @@
+import chalk from 'chalk';
+import { dedent } from 'ts-dedent';
+
+import { error, info } from '../../components/icons';
+import link from '../../components/link';
+
+const docsUrl = 'https://www.chromatic.com/docs/vitest/turbosnap';
+
+export const turboSnapDisabledVitest = () =>
+  dedent(chalk`
+    ${error} TurboSnap failed
+    Vitest run did not produce source file information for TurboSnap.
+    Make sure you've enabled ${chalk.bold('chromaticPlugin({ turboSnap: true })')} in your test Vitest configuration before running tests.
+    ${info} Read more at ${link(docsUrl)}
+  `);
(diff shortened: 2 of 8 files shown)


node-src/lib/checkStorybookBaseDirectory.ts
AriPerkkio, line 49: If we end up here with empty `sourceModuleFiles`, the `Promise.any` will also throw. We may end up with `preview-stats.json` that doesn't contain source files if user runs Vitest without `turboSnap: true`. It's different error case than the one `throw new Error(invalidStorybookBaseDirectory())` below indicates.

Btw Node REPL has nice easter egg for `Promise.any([])` 😅 
- https://github.com/nodejs/node/issues/65581

node-src/lib/checkStorybookBaseDirectory.ts
```diff
@@ -24,11 +25,20 @@ export async function checkStorybookBaseDirectory(ctx: Context, stats: Stats) {
...
+      !module.name.includes('node_modules') &&
+      module.name !== './storybook-stories.js' && // builder-webpack5 virtual file
+      /\.(js|jsx|ts|tsx)$/.test(module.name)
   );
 
+  // This can happen if `vitest run` is run without the chromaticPlugin({ turboSnap: true })
+  // and `chromatic --vitest --only-changed` is run afterwards.
+  if (ctx.options.vitest && sourceModuleFiles.length === 0) {
+    throw new Error(turboSnapDisabledVitest());
+  }
```
AriPerkkio, line 41: We could also move this check in its own isolated module `checkPreviewStats.ts` that would be called between these two lines:

https://github.com/chromaui/chromatic-cli/blob/3b4eaf63ed5b25742fdca8c222532daec53ce1c7/node-src/lib/turbosnap/v1/index.ts#L87-L89

As Vitest runs currently generate Storybook with `builder-webpack5`, we always end up with `preview-stats.json`. But we only add the source module info when `turboSnap: true` option is passed to the `chromaticPlugin()` plugin. So we need to check if the content has any source modules - it's kind of same as checking if stats file exists in pure Storybook runs.

node-src/ui/messages/errors/turboSnapDisabledVitest.ts
```diff
@@ -0,0 +1,15 @@
...
+
+const docsUrl = 'https://www.chromatic.com/docs/vitest/turbosnap';
+
+export const turboSnapDisabledVitest = () =>
+  dedent(chalk`
+    ${error} TurboSnap failed
+    Vitest run did not produce source file information for TurboSnap.
+    Make sure you've enabled ${chalk.bold('chromaticPlugin({ turboSnap: true })')} in your test Vitest configuration before running tests.
+    ${info} Read more at ${link(docsUrl)}
+  `);
```
AriPerkkio, line 15: This structure is inspired by

https://github.com/chromaui/chromatic-cli/blob/3b4eaf63ed5b25742fdca8c222532daec53ce1c7/node-src/ui/messages/errors/invalidPatchBuild.ts#L7-L15

node-src/ui/messages/errors/turboSnapDisabledVitest.ts
```diff
@@ -0,0 +1,15 @@
...
+import link from '../../components/link';
+
+const docsUrl = 'https://www.chromatic.com/docs/vitest/turbosnap';
+
+export const turboSnapDisabledVitest = () =>
+  dedent(chalk`
+    ${error} TurboSnap failed
+    Vitest run did not produce source file information for TurboSnap.
+    Make sure you've enabled ${chalk.bold('chromaticPlugin({ turboSnap: true })')} in your test Vitest configuration before running tests.
+    ${info} Read more at ${link(docsUrl)}
```
jmhobbs, line 14: @AriPerkkio did we run this past DX?  I know we've done a lot of vocabulary changes around TurboSnap, want to ensure this is compliant.

AriPerkkio: No, haven't asked DX about this. Lets get review from them too.

node-src/lib/checkStorybookBaseDirectory.ts
```diff
@@ -24,11 +25,20 @@ export async function checkStorybookBaseDirectory(ctx: Context, stats: Stats) {
...
+    (module: any) =>
+      !module.name.includes('node_modules') &&
+      module.name !== './storybook-stories.js' && // builder-webpack5 virtual file
+      /\.(js|jsx|ts|tsx)$/.test(module.name)
   );
 
+  // This can happen if `vitest run` is run without the chromaticPlugin({ turboSnap: true })
+  // and `chromatic --vitest --only-changed` is run afterwards.
+  if (ctx.options.vitest && sourceModuleFiles.length === 0) {
+    throw new Error(turboSnapDisabledVitest());
```
jmhobbs, line 40: Should this call `setExitCode`, similar to `invalidStorybookBaseDirectory` down below?

node-src/lib/checkStorybookBaseDirectory.ts
```diff
@@ -24,11 +25,20 @@ export async function checkStorybookBaseDirectory(ctx: Context, stats: Stats) {
   const { storybookBaseDir: storybookBaseDirectory = path.relative(repositoryRoot, '') } =
     ctx.options;
 
-  // Find all js(x)/ts(x) files in stats that are not in node_modules
+  // Find all js(x)/ts(x) files in stats that are not in node_modules, and are not Storybook's virtual files
   const sourceModuleFiles = stats.modules.filter(
-    (module: any) => !module.name.includes('node_modules') && /\.(js|jsx|ts|tsx)$/.test(module.name)
+    (module: any) =>
+      !module.name.includes('node_modules') &&
+      module.name !== './storybook-stories.js' && // builder-webpack5 virtual file
```
jmhobbs, line 32: Does this apply to any other builders?  Is that virtual file likely to ever move or change?  If so is there any _other_ mechanism to detect it?  I know TS v2 collected some together, maybe some overlap there? https://github.com/chromaui/chromatic-cli/blob/ec1c832cd653065464dcbae53ac24979f5169022/node-src/lib/turbosnap/v2/storyDetection.ts#L30

jmhobbs: Oh, I see on re-read, sorry 😅 
> As Vitest runs currently generate Storybook with builder-webpack5

AriPerkkio: It's safe to always ignore this file, even if we move away from `builder-webpack5`. At the moment it's the only one that needs to be ignored.

node-src/ui/messages/errors/turboSnapDisabledVitest.ts
AriPerkkio: Pinged DX for a review 👀

AriPerkkio: Ok, we got 👍 from @kylegach.

jmhobbs (approved):

PR #1471, opened 2026-08-28
:sparkles: Track pnpm-lock.yaml as a "package metadata file".

jwir3:
Currently, `isPackageMetadataFile()` returns `false` for `pnpm-lock.yaml`. If a commit's only meaningful change is to that file, `git.packageMetadataChanges` in `gitInfo.ts` won't include it. Because `packageMetadataChanges` ends up empty, `findChangedDependencies()`, which is gated on `isPackageMetadataFile()` having non-zero length, is never called at all, so no dependency changes are ever computed for that commit. `getDependentStoryFiles()` still runs, but since `pnpm-lock.yaml` isn't part of the module graph, it finds nothing affected. The result is that a build is created, but with zero story files flagged as changed.

This PR changes the behavior so that `pnpm-lock.yaml` is flagged as a "package metadata file". This routes to `findChangedDependencies()` (which, oddly, already _does_ parse and check `pnpm-lock.yaml`), and dependency changes are identified in the same way they would be if a change to `yarn.lock` was made _without_ a change to `package.json`. 

Fixes CAP-4975.

<details>
  <summary>📦 Published PR as canary version: <code>18.8.0--canary.1471.33213179804.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.8.0--canary.1471.33213179804.0
  # or 
  yarn add chromatic@18.8.0--canary.1471.33213179804.0
  ```
</details>

### DIFF
--- a/node-src/__mocks__/dependencyChanges/pnpm-catalog-after/package.json
+++ b/node-src/__mocks__/dependencyChanges/pnpm-catalog-after/package.json
@@ -1,0 +1,7 @@
+{
+  "name": "storybook-demo",
+  "version": "1.0.0",
+  "dependencies": {
+    "is-odd": "catalog:"
+  }
+}
--- a/node-src/__mocks__/dependencyChanges/pnpm-catalog-after/pnpm-lock.yaml
+++ b/node-src/__mocks__/dependencyChanges/pnpm-catalog-after/pnpm-lock.yaml
@@ -1,0 +1,37 @@
+lockfileVersion: '9.0'
+
+settings:
+  autoInstallPeers: true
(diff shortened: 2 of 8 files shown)


codykaup:
@jwir3 I tested this out on a PNPM monorepo that's using [catalogs](https://pnpm.io/catalogs) and it's not detecting the version change even after this change. Is this change working for you?

jwir3:
@codykaup Can you send me the monorepo you're using?

codykaup:
> @codykaup Can you send me the monorepo you're using?

Yes! https://github.com/codykaup/pnpm-monorepo

PR #1478, opened 2026-09-08
Set component owners if present

justin-thurman:
# Description

This PR sets component owners based on the `COMPONENTOWNERS` file during the `upload` task. A few functional requirements inform the implementation:
1. Setting component owners is best effort and should never fail the build.
2. We only support `COMPONENTOWNERS` files at the repo root.
3. The CLI does no parsing/validation of the file. We just send it.

Because of that first requirement, the added function swallows and logs its own errors because in this case, "what should happen when this fails?" is knowledge that is owned by the feature, not something that only the caller knows. So good information hiding practice suggests putting the try/catch inside the function itself. I just wanted to call that decision out, since it's the opposite of usual best practices.

I also decided to put this mutation inside the `initialize` task, alongside announcing the build, rather than during the upload stage that was previously discussed in the ticket. I felt that aligned better with what this work is doing (setting a type of build metadata), whereas putting it in upload aligned with the implementation detail of how that work was done (uploading a file, though technically not, since we're sending the file's contents in a mutation, not uploading it anywhere, but you get the idea). This seemed like a better fit to me. It also allows us to set component owners even on dry runs, matching announce build.

# Manual QA

I ran three builds:
1. No `COMPONENTOWNERS` file present.
2. `COMPONENTOWNERS` file is present (copied the file from the monorepo)
3. `COMPONENTOWNERS` directory present, forcing an error.

All three builds succeeded. No CLI output for builds 1 and 2. Build 3 showed the warning log `Unable to set component owners. EISDIR: illegal operation on a directory, read`.

Mongo shows no component owners for builds 1 and 3, and successful component owners for build 2:
```
database> db.componentOwners.findOne({buildId: ObjectId('6aa172c60a55c4809b5097f6')})
null
database> db.componentOwners.findOne({buildId: ObjectId('6aa1734a0a55c4809b509800')})
{
  _id: ObjectId('6aa1734a65c02b8fed816c8a'),
  buildId: ObjectId('6aa1734a0a55c4809b509800'),
  createdAt: ISODate('2026-09-09T14:55:06.731Z'),
  rules: [
    { pattern: '*', owners: [ 'estelle@chromatic.com' ] },
    {
      pattern: 'Example/Button/*',
      owners: [ 'kate@chromatic.com', 'estelle@chromatic.com' ]
    },
    { pattern: 'Example/Header/*', owners: [ 'jennifer@chromatic.com' ] }
  ],
  updatedAt: ISODate('2026-09-09T14:55:06.731Z')
}
database> db.componentOwners.findOne({buildId: ObjectId('6aa173ba0a55c4809b50980a')})
null
database>
```

<details>
  <summary>📦 Published PR as canary version: <code>18.9.0--canary.1478.34497647280.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.9.0--canary.1478.34497647280.0
  # or 
  yarn add chromatic@18.9.0--canary.1478.34497647280.0
  ```
</details>

### DIFF
--- a/node-src/lib/componentowners.ts
+++ b/node-src/lib/componentowners.ts
@@ -1,0 +1,41 @@
+import { readFile } from 'fs/promises';
+import path from 'path';
+
+import { Deps } from '../types';
+
+const SetComponentOwnersMutation = `
+  mutation SetComponentOwners($buildId: ObjID!, $content: String!) {
+    setComponentOwners(buildId: $buildId, content: $content) {
+      buildId
+    }
+  }
+`;
+
+/**
+ * Tries to set component owners based on the COMPONENTOWNERS file, if present at the repository root.
+ * Setting component owners is best effort and should never fail the build. Because of this, this function
+ * never throws.
+ *
+ * @param deps Dependencies (the GraphQL client, the logger)
+ * @param rootPath Absolute path to the repository root.
+ * @param buildId ID of the announced build.
+ */
+export async function trySetComponentOwnersIfPresent(
+  deps: Readonly<Pick<Deps, 'client' | 'log'>>,
+  rootPath: string,
+  buildId: string
+) {
+  try {
+    const content = await readFile(path.join(rootPath, 'COMPONENTOWNERS'), 'utf8');
+    await deps.client.runQuery(SetComponentOwnersMutation, {
+      buildId,
+      content,
+    });
+  } catch (err) {
+    if ((err as NodeJS.ErrnoException).code === 'ENOENT') {
+      // COMPONENTOWNERS file not present. Skip and move on.
+      return;
+    }
+    deps.log.warn('Unable to set component owners.', err);
+  }
+}
--- a/node-src/lib/componentowners.test.ts
+++ b/node-src/lib/componentowners.test.ts
@@ -1,0 +1,77 @@
+import { mkdir, mkdtemp, rm, writeFile } from 'fs/promises';
+import { tmpdir } from 'os';
+import path from 'path';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { trySetComponentOwnersIfPresent } from './componentowners';
+import TestLogger from './testLogger';
+
+describe('trySetComponentOwnersIfPresent', () => {
+  let rootPath: string;
+  let log: TestLogger;
+  const client = { runQuery: vi.fn() };
+
+  beforeEach(async () => {
+    rootPath = await mkdtemp(path.join(tmpdir(), 'componentowners-'));
+    log = new TestLogger();
+    client.runQuery.mockReset();
+  });
+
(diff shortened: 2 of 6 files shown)


node-src/lib/componentowners.ts
```diff
@@ -0,0 +1,41 @@
...
+ * @param rootPath Absolute path to the repository root.
+ * @param buildId ID of the announced build.
+ */
+export async function trySetComponentOwnersIfPresent(
+  deps: Readonly<Pick<Deps, 'client' | 'log'>>,
+  rootPath: string,
+  buildId: string
+) {
+  try {
+    const content = await readFile(path.join(rootPath, 'COMPONENTOWNERS'), 'utf8');
```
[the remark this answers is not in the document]
justin-thurman: @AikidoSec ignore: This code runs on the user's own machine. There's no read file risk.

justin-thurman:
cc @katestam, in case you want to do any QA with the canary.

node-src/lib/componentowners.ts
jmhobbs, line 29: Do we want to bail early if this is over a certain size, or just let the index control that?

node-src/lib/componentowners.ts
```diff
@@ -0,0 +1,41 @@
...
+    await deps.client.runQuery(SetComponentOwnersMutation, {
```
jmhobbs, line 30: Should we limit retries on this? Esp. for a big component owners file, could waste time churning.

node-src/lib/componentowners.ts
jmhobbs, line 29: Are we explicitly calling out in docs somewhere that this must be all caps?  I'm thinking of a user on a case-insensitive filesystem with a `ComponentOwners` file seeing it work locally then it failing in CI because the Linux FS is case-sensitive.  Kind of a stretch though, and not a blocker.

jmhobbs (approved):

node-src/lib/componentowners.ts
justin-thurman: It's already limited to 2 by default in the GQL client implementation. That seemed fine to me.

node-src/lib/componentowners.ts
justin-thurman: I was thinking we just let the index control it. I wouldn't want to duplicate information on the limit, especially since the CLI version could get stale since users don't upgrade. Nothing on the CLI would be enforcing anyway since users can just fork it.

justin-thurman: I'm not sure if it's documented, but it's definitely what the project is using so far (as seen in the monorepo). @katestam, do you know the answer to this one?

katestam: yes, we are using `COMPONENTOWNERS`!

PR #1481, opened 2026-09-08
Communicate the number of ignored tests

jmhobbs:
Update the CLI output to better match the GitHub comment UI.

Show the number of changes to be accepted as baselines, and add the count and direct link to ignored tests section of the build page.

<details>
  <summary>📦 Published PR as canary version: <code>18.8.2--canary.1481.34954158001.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.8.2--canary.1481.34954158001.0
  # or 
  yarn add chromatic@18.8.2--canary.1481.34954158001.0
  ```
</details>


node-src/ui/messages/errors/buildHasChanges.ts
```diff
@@ -7,23 +7,38 @@ import link from '../../components/link';
 
 export default ({ build, exitCode, isOnboarding }) => {
   const url = isOnboarding ? build.app.setupUrl : build.webUrl;
+  const unstableUrl = `${url}#unstable`;
 
   const changes: any[] = [];
-  if (build.changeCount > 0) {
+
+  if (build.changeCount > 0 && build.accessibilityChangeCount > 0) {
```
justin-thurman, line 14: This nested if/else is pretty confusing. What do you think about this version? Output is identical, but no if/else.
```ts
export default ({ build, exitCode, isOnboarding }) => {
  const url = isOnboarding ? build.app.setupUrl : build.webUrl;

  const kinds = [
    build.changeCount > 0 && 'visual',
    build.accessibilityChangeCount > 0 && 'accessibility',
  ].filter(Boolean);
  const changeTotal = (build.changeCount || 0) + (build.accessibilityChangeCount || 0);

  const changesLine =
    changeTotal > 0 &&
    chalk`${error} {bold ${pluralize(`${kinds.join(' and ')} changes`, changeTotal, true)} must be accepted as ${pluralize('baseline', changeTotal)}.} Review at ${link(url)}`;
  const ignoredLine =
    build.ignoredCount > 0 &&
    chalk`{bold ${pluralize('test', build.ignoredCount, true)} ${pluralize('was', build.ignoredCount)} ignored in this build.} Review at ${link(`${build.webUrl}#unstable`)}`;

  return dedent(chalk`
    ${[changesLine, ignoredLine].filter(Boolean).join('\n\n')}

    ${info} For CI/CD use cases, this command failed with exit code ${exitCode}
    ...
  `);
};
```

node-src/ui/messages/errors/buildHasChanges.stories.ts
```diff
@@ -23,14 +24,20 @@ export const BuildHasChangesNotOnboarding = () => buildHasChanges(context);
...
     ...context,
-    build: { ...context.build, accessibilityChangeCount: 0 },
+    build: { ...context.build, accessibilityChangeCount: 0, ignoredCount: 0 },
   });
 
 export const BuildHasChangesAccessibilityOnly = () =>
   buildHasChanges({
     ...context,
-    build: { ...context.build, changeCount: 0 },
+    build: { ...context.build, changeCount: 0, ignoredCount: 0 },
```
justin-thurman, line 33: I think the accessibility counting is dead. `SnapshotBuildQuery` doesn't query for `accessibilityChangeCount`. We should either add that to the query if we need it or drop it from the stories and `buildHasChanges`.

node-src/ui/messages/errors/buildHasChanges.ts
```diff
@@ -7,23 +7,38 @@ import link from '../../components/link';
 
 export default ({ build, exitCode, isOnboarding }) => {
   const url = isOnboarding ? build.app.setupUrl : build.webUrl;
+  const unstableUrl = `${url}#unstable`;
```
justin-thurman, line 10: This would add the `#unstable` anchor to the onboarding URL when we're onboarding, which wouldn't do anything. Should this use the build URL unconditionally? The refactored version of this I posted assumes this is the case.

node-src/ui/tasks/snapshot.ts
```diff
@@ -27,6 +27,7 @@ export const stats = ({
     testCount: number;
     changeCount: number;
     errorCount: number;
+    ignoredCount?: number;
```
justin-thurman, line 30: This value was added here, but not actually read in the return. Should we read it, or just remove?

node-src/types.ts
```diff
@@ -448,6 +448,7 @@ export interface Context {
     errorCount: number;
     accessibilityChangeCount: number;
     interactionTestFailuresCount: number;
+    ignoredCount?: number;
```
justin-thurman, line 451: Should this be required, like the other `*Count` variables above?

node-src/ui/messages/errors/buildHasChanges.ts
justin-thurman, line 14: `build.ChangeCount` includes all changed tests, regardless of whether they're ignored. So we might end up with, e.g., one visual change that needs review and two ignored changes, but the CLI output would be "3 visual changes must be accepted as baselines" and "2 tests were ignored", but only one test actually needs to be accepted. Do we need to handle that case? If you need to look at the monorepo, the relevant code is `Test.ts`, lines 909, 965, and 1001, and `testChangeCounts.ts`.

justin-thurman (changes requested):
Looks good overall, but a few areas I think could be cleaned up. Also, a couple questions.
1. Ignored-only builds go through `buildPassed`, not `buildHasChanges`, which doesn't include messaging around ignored count. Do we need to add it there? `--exit-zero-on-changes` also flows through `buildPassed` and has the same concern.
2. Some concerns about the messaging about visual changes including ignored tests (see inline comment).

PR #1484, opened 2026-09-10
Strip webpack escape characters in file paths

codykaup:
Closes CAP-5067

Looks like Webpack can add NUL byte escape characters based on the file paths. We need to remove those in order to get the correct file path on disk to hash the file.

For example:

`path/#/to/file.js` is `path/\u0000#/to/file.js` in the Webpack output of `preview-stats.json`.

That output is then converted to `path/\0#/to/file.js` when we do a `JSON.parse` so we can simply strip those out and we're left with the real file path: `path/#/to/file.js`.

<details>
  <summary>📦 Published PR as canary version: <code>18.8.2--canary.1484.34614158315.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.8.2--canary.1484.34614158315.0
  # or 
  yarn add chromatic@18.8.2--canary.1484.34614158315.0
  ```
</details>

### DIFF
--- a/node-src/lib/turbosnap/v2/paths.test.ts
+++ b/node-src/lib/turbosnap/v2/paths.test.ts
@@ -227,6 +227,12 @@ describe('normalizeStatsPath', () => {
     expect(normalizeStatsPath('./src/a.ts + 1 module', projectRoot)).toBe('./src/a.ts');
   });
 
+  it('keys a NUL-escaped filename by the name it really has on disk', () => {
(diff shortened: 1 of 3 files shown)


jmhobbs (approved):
That's a weird one, but the change makes sense.

codykaup:
> That's a weird one, but the change makes sense.

Builders are weird, man. 🤷‍♂️

PR #1489, opened 2026-09-14
Determine globals by their imports

codykaup:
> [!IMPORTANT]
> This is still running in a monitoring state and won't impact customer bills!

It appears we have a bug in TurboSnap 2.0 where dependencies that should be `storybookGlobals` are getting linked to stories when they shouldn't be. This means that if you import something like `React` to a single story, then upgrade it, we only detect a change to that single story. Instead, we need to recapture the entire Storybook again. 

This fixes it by walking to `storybookGlobals` from two kinds of roots. First, the builder's composition roots (`storybook-config-entry.js` on webpack/rsbuild, `vite-app.js` and `project-annotations.js` on Vite): everything they reach loads for every story, so it is global even when a story also imports it. Second, as a fallback for stats that omit those edges, every hashed file in no story subtree and no preview subtree, plus what it imports. Both walks stop at story files. A file can now sit in both a story subtree and `storybookGlobals`; the story hash names the affected story and the globals hash says the change can affect every story.

> [!IMPORTANT]
> On Vite, the composition-root walk needs the stats from storybookjs/storybook#36345, which links `project-annotations.js` to the annotations. Storybook 10.3.0 through the release before that patch falls back to the disconnected-file rule. Webpack and Rsbuild need no Storybook change, so they are the easiest place to QA this.

# Example

> [!NOTE]
> This example is used for demonstration and not an exact example we've seen in production.

Let's say the story import `react`:
```
...
import { useEffect } from 'react';
...
```

## Before

Manifest from React `17.0.1`:
```
{
  "storybookHash": "storybookHash1",
  "storybookConfigHashes": {
    ...
    "storybookGlobals": "storybookGlobalHash1",
    ...
  },
  ...
  "storyFiles": {
    "./src/stories/Button.stories.jsx": "storyHash1"
  },
  ...
```

Manifest from React `17.0.2`:
```
{
  "storybookHash": "storybookHash2", ✅ this changed
  "storybookConfigHashes": {
    ...
    "storybookGlobals": "storybookGlobalHash1", ❌ this should've changed
    ...
  },
  ...
  "storyFiles": {
    "./src/stories/Button.stories.jsx": "storyHash2" ✅ only this changed so we'd only recapture `./src/stories/Button.stories.jsx`
  },
  ...
```

## After

Manifest from React `17.0.1`:
```
{
  "storybookHash": "storybookHash1",
  "storybookConfigHashes": {
    ...
    "storybookGlobals": "storybookGlobalHash1",
    ...
  },
  ...
  "storyFiles": {
    "./src/stories/Button.stories.jsx": "storyHash1"
  },
  ...
```

Manifest from React `17.0.2`:
```
{
  "storybookHash": "storybookHash2", ✅ this changed
  "storybookConfigHashes": {
    ...
    "storybookGlobals": "storybookGlobalHash2", ✅ this changed
    ...
  },
  ...
  "storyFiles": {
    "./src/stories/Button.stories.jsx": "storyHash2" ✅ this also changed which is fine because we're recapturing anyway
  },
  ...
```

<details>
  <summary>📦 Published PR as canary version: <code>18.9.4--canary.1489.35256360426.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.9.4--canary.1489.35256360426.0
  # or 
  yarn add chromatic@18.9.4--canary.1489.35256360426.0
  ```
</details>


node-src/lib/turbosnap/v2/storybookFiles.ts
```diff
@@ -31,17 +31,34 @@ function isPreviewConfig(filePath: FilePath, configDirectory: FilePath): boolean
...
+ * What the caller already learned about the stories while hashing them, passed in rather than walked
+ * again here.
+ */
+export interface Stories {
+  /** The story files themselves, where the globals walk stops. */
+  storyFiles: Set<FilePath>;
+  /** The union of every story's transitive subtree. */
+  reachable: Set<FilePath>;
+}
+
```
justin-thurman, line 44: I'm not sure this interface earns its keep. It looks like `collectStorybookFiles` expects this interface to be consistent with the import graph, but nothing enforces that (the tests take advantage of this by passing inconsistent interfaces). That's some pretty heavy coupling between `collectStorybookFiles` and the caller. Avoiding one more walk of the graph doesn't seem like that big of a gain, especially given that we're already doing per-story walks. I think passing in the reachable set made sense when it was just one parameter, but now it's two coupled parameters. What do you think about removing `reachable`, just passing in story files, and recomputing reachable inside `collectStoryFiles`?

justin-thurman (changes requested):
This looks good for fixing one class of this bug, but I think there's a gap here. So here's my understanding of how things work before and after this PR (visually, because reasoning about this in text form is melting my brain).

Before the PR:
Image

After this PR:
Image

Runtime library update is detected as a global, even though the story imports it. 👍 

Missed case:
Image

If the story imports the add-on, nothing walks from the add-on forward to the runtime -> globals hash doesn't change.

Am I missing something here about why this can't happen or isn't a problem?

Reproducing test case, run from `lib/turbosnap/v2`:
```ts
import { expect, it } from 'vitest';

import { Stats } from '../../../types';
import { createFixture, projectRoot, syntheticAbsent } from './__fixtures__/manifestFixtures';
import { buildManifest } from './manifest';

it.each([false, true])('keeps a runtime upgrade global when a story imports its entry: %s', async (importsEntry) => {
  const story = `${projectRoot}/src/Button.stories.tsx`;
  const otherStory = `${projectRoot}/src/Header.stories.tsx`;
  const addon = `${projectRoot}/node_modules/example-addon/preview.js`;
  const runtime = `${projectRoot}/node_modules/example-runtime/index.js`;
  const unrelated = `${projectRoot}/node_modules/other-addon/preview.js`;
  const configEntry = './storybook-config-entry.js';
  const stats: Stats = {
    modules: [
      { id: 1, name: story, reasons: [{ moduleName: './storybook-stories.js' }] },
      { id: 2, name: otherStory, reasons: [{ moduleName: './storybook-stories.js' }] },
      { id: 3, name: addon, reasons: [
        { moduleName: configEntry },
        ...(importsEntry ? [{ moduleName: story }] : []),
      ] },
      { id: 4, name: runtime, reasons: [{ moduleName: addon }, { moduleName: story }] },
      { id: 5, name: unrelated, reasons: [{ moduleName: configEntry }] },
    ],
  };
  const { disk, input } = createFixture({
    isAbsent: syntheticAbsent,
    fileHashes: { [story]: 'S', [otherStory]: 'H', [addon]: 'A', [runtime]: 'R1', [unrelated]: 'U' },
  });
  const before = await buildManifest(stats, input);
  disk.fileHashes![runtime] = 'R2';
  const after = await buildManifest(stats, input);

  expect(after.storyFileHashes.get('./src/Button.stories.tsx')).not.toBe(before.storyFileHashes.get('./src/Button.stories.tsx'));
  expect(after.storyFileHashes.get('./src/Header.stories.tsx')).toBe(before.storyFileHashes.get('./src/Header.stories.tsx'));
  expect(after.storybookConfigHashes.get('storybookGlobals')).not.toBe(before.storybookConfigHashes.get('storybookGlobals'));
});
```

justin-thurman:
Also, this one is entirely AI generated, so grain of salt, but I don't know enough about builders to validate whether it's accurate. But it seems potentially big enough that it's worth passing along. Does this only work for webpack but not for vite or other builder shapes? I know you said in the PR body that we usually see this in webpack, but is it impossible in vite? This test demonstrates the issue: globals membership seems dependent on builder shape, namely whether it records the `config-entry -> preview.ts` import edge or not.
```ts
describe('collectStorybookFiles globals membership by builder shape', () => {
  // Same project, same files, same preview subtree. The only difference is whether the bundler
  // records the config-entry -> preview.ts import edge (webpack/rsbuild do, vite does not).
  const story = './src/a.stories.tsx';
  const preview = './.storybook/preview.ts';
  const theme = './.storybook/theme.ts';
  const stories = { reachable: new Set([story]), storyFiles: new Set([story]) };

  function globalsFor(files: Map<FilePath, TurboSnapFile>): FilePath[] {
    const { attribution } = collectStorybookFiles(
      files,
      makeHashes([...files.keys()].filter((filePath) => !filePath.startsWith('glob:'))),
      stories,
      DEFAULT_CONFIG_DIR,
      identity
    );
    return [...attribution.storybookGlobals].sort();
  }

  it('webpack shape: the config entry imports preview, so the preview subtree lands in globals', () => {
    const files = makeFiles({
      './storybook-config-entry.js': [preview, 'glob:stories'],
      'glob:stories': [story],
      [preview]: [theme],
      [theme]: [],
      [story]: [],
    });

    expect(globalsFor(files)).toEqual([preview, theme, './storybook-config-entry.js']);
  });

  it('vite shape: preview is a detached root, so the preview subtree stays out of globals', () => {
    const files = makeFiles({
      './node_modules/@storybook/core/entry.js': ['glob:stories'],
      'glob:stories': [story],
      [preview]: [theme],
      [theme]: [],
      [story]: [],
    });

    expect(globalsFor(files)).toEqual(['./node_modules/@storybook/core/entry.js']);
  });
});
```

The second test passing is (supposedly) the bug.

codykaup:
> If the story imports the add-on, nothing walks from the add-on forward to the runtime -> globals hash doesn't change.
>
> Am I missing something here about why this can't happen or isn't a problem?

Nope, that's totally a valid case. I thought add-ons would be swept into globals since it'd be imported by one of the runtime files but, apparently, it's not. 🫠 

Nice find — I'm going to move this into draft until I can figure this one out.

node-src/lib/turbosnap/v2/storybookFiles.ts
codykaup: Made that adjustment!

codykaup:
> Also, this one is entirely AI generated, so grain of salt, but I don't know enough about builders to validate whether it's accurate. But it seems potentially big enough that it's worth passing along. Does this only work for webpack but not for vite or other builder shapes? I know you said in the PR body that we usually see this in webpack, but is it impossible in vite? This test demonstrates the issue: globals membership seems dependent on builder shape, namely whether it records the `config-entry -> preview.ts` import edge or not.

@justin-thurman thanks for passing it along! I don't _think_ it's valid after the changes I made.

codykaup:
@justin-thurman thanks for calling out the addon case. It's fixed now so we're in a better spot. However, we're more-or-less stuck on the Vite side of things (Storybook 10.3+) until we get a complete `preview-stats.json` graph from [this PR](https://github.com/storybookjs/storybook/pull/36345).

We already talked about the change but basically we're walking forward from the builder entry point and treating everything it reaches as global, stopping at story files so we don't include story code as global. A file that both the entry point and a story reach lands in both the story's hash and the globals hash. Unfortunately, our Vite builder has a bug where it's filtering out the path between `builder entry -> globals`, which is what that Storybook PR addresses. 🤞

node-src/lib/turbosnap/v2/storybookFileKeys.ts
```diff
@@ -11,7 +11,10 @@
  */
 export const STORYBOOK_PREVIEW_KEY = 'preview';
 
-/** Every orphan global, rolled up; see {@link collectStorybookFiles}. */
+/**
+ * Every file that can affect every story, rolled up: the files in no story and no preview
+ * subtree, plus everything they import. See {@link collectStorybookFiles}.
+ */
```
justin-thurman, line 25: This describes only the fallback rule, but the root walk is now the primary mechanism. Should we update this to reflect that?

node-src/lib/turbosnap/v2/statsGraph.ts
```diff
@@ -92,6 +113,29 @@ export function countNodeModulesFiles(stats: Stats): number {
...
+ * absent from the content hashes, but they are retained in this graph.
+ * @param roots The filesystem roots used to resolve builder-specific module names; see
+ * {@link StatsRoots}.
+ *
+ * @returns The canonical paths of the global composition roots present in the graph.
+ */
+function detectGlobalRoots(files: Map<FilePath, TurboSnapFile>, roots: StatsRoots): Set<FilePath> {
+  // Normalize the known builder-specific names the same way as every other stats path before
+  // comparing them with the graph's canonical keys.
+  const known = [...CONFIG_ENTRY_FILES, ...VITE_COMPOSITION_ROOTS].map((name) =>
```
justin-thurman, line 137: It seems like a bit of a code smell that this function needs to read `CONFIG_ENTRY_FILES`, which comes from another module, and `VITE_COMPOSITION_ROOTS`, which is in this file. In terms of how we're using these values, are they kind of the same thing? Should these files be consolidated somehow?

justin-thurman (approved):
Looks good! A couple non-blocking thoughts. Is the plan to wait to merge this until we get a yes/no on the Storybook upstream PR? If that PR falls through for some reason, it seems like this PR will need to be reworked.

node-src/lib/turbosnap/v2/storybookFileKeys.ts
codykaup: Ope, missed that one!

codykaup: Addressed in 46270ac1fac85bba214421d8f0af358493fff8b1

node-src/lib/turbosnap/v2/statsGraph.ts
codykaup: Based on this new bug discovery, the whole parsing flow feels like a code smell now. 🫠

`CONFIG_ENTRY_FILES` and `VITE_COMPOSITION_ROOTS` are pretty confusing, so there's a refactoring opportunity here.

- `CONFIG_ENTRY_FILES`: on Webpack/Rspack, the config entry loads the preview annotations _and_ imports the story-importers. We use this to determine which stories are included in the Storybook.
- `VITE_COMPOSITION_ROOTS`: on Vite, these only load the Storybook internal loaders (addons, preview config, etc)

Merging them would make Vite builds mark things like `.storybook/preview.js` as a story because we're expecting `CONFIG_ENTRY_FILES` to lead us to story files that live in the file system.

This basically goes back to what we discussed yesterday: Pulling out all the builder-specific logic into separate packages. Since they're all jumbled together, it's _really_ hard to reason about. Isolating all that logic should help a ton.

codykaup:
> Is the plan to wait to merge this until we get a yes/no on the Storybook upstream PR? If that PR falls through for some reason, it seems like this PR will need to be reworked.

@justin-thurman I was planning to merge this one regardless since this makes globals more accurate for Webpack and Rspack. Vite is a grey area because 10.3+ is detached but at least we're including the dependencies of the globals we find in our calculation. Then if the Storybook PR merges, it just becomes more accurate with a Storybook version bump.

node-src/lib/turbosnap/v2/statsGraph.ts
codykaup: I'm fine with merging this as-is so we can see if it fixes our issue in the metrics. I think we're going to have to pull each builder out in a refactor because it's _already_ hard to figure out what's going on. Isolating that logic is going to be a must moving forward.

WDYT?

justin-thurman: Yeah, I'm fine to merge this as is. And I didn't mean merging these two arrays into one, btw. I meant merging the two files that they come from into one file, or a directory module if you want to keep separate files. They seem to serve very similar purposes, and so the fact that they come from two different modules struck me as odd. Or another way this manifests: I feel like the `detectGlobalRoots` function living in `statsGraph.ts` currently is arbitrary, and you could easily make the argument that it live in `storyDetection.ts` instead. That suggests to me that this is actually one module incorrectly split into two.

But it's very possible I'm misreading this or don't have a full understanding of the legitimate split here! Definitely non-blocking.

codykaup: Completely valid feelings! I think we'll need to come back and refactor since concepts are kinda merging a little bit here.

PR #1490, opened 2026-09-14
Always upload TurboSnap v2 hashes

codykaup:
Because TurboSnap v2 uses a different mechanism, we need to be collecting source file hashes way before the customer requests TurboSnap to run. Since v2 is fast enough and the manifest isn't too expensive to store, we decided to run TurboSnap v2 by default but without impacting what stories we capture. It simply uploads the manifest and moves on.

It could also be pretty confusing to run a normal build but see TurboSnap logs so we're flipped the log level of the TurboSnap v2 logger based on if TurboSnap was requested or not.

<details>
  <summary>📦 Published PR as canary version: <code>18.10.0--canary.1490.35002347581.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.10.0--canary.1490.35002347581.0
  # or 
  yarn add chromatic@18.10.0--canary.1490.35002347581.0
  ```
</details>


node-src/lib/getEnvironment.ts
```diff
@@ -11,6 +11,7 @@ export interface Environment {
   CHROMATIC_SHARE_PROGRESS_INTERVAL: number;
   CHROMATIC_STORYBOOK_VERSION?: string;
   CHROMATIC_TIMEOUT: number;
+  CHROMATIC_TURBOSNAP_DISABLE_HASHES: boolean;
```
justin-thurman, line 14: What's the point of this?

justin-thurman (approved):
👍

node-src/lib/getEnvironment.ts
codykaup: Just in case something funky happens with the TurboSnap v2 algorithm, it gives them an escape hatch to not run it until we have a chance to patch the issue.

PR #1491, opened 2026-09-14
Read Storybook version from config directory then fallback

codykaup:
TurboSnap v2 was failing to determine the Storybook version which prevents the algorithm from running. Right now, it appears to be because we aren't looking for an installed Storybook from the config directory. So if you run `chromatic` in a parent directory, we read the parent directory's `node_modules` instead of the config directory's (which is normally a subdirectory). This change bases it's `require` from the config directory then walks up the directory structure (which then includes the parent directory it was called from) looking for the dependency we need.

This also falls back to the previously determined Storybook version based on the view layer from `ctx.storybook.version`. I didn't just lean on that version to keep the scope of the change down to _just_ TurboSnap v2 rather than the entire CLI. Ideally we'd grab the Storybook version then use it everywhere.

<details>
  <summary>📦 Published PR as canary version: <code>18.9.2--canary.1491.35112949780.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.9.2--canary.1491.35112949780.0
  # or 
  yarn add chromatic@18.9.2--canary.1491.35112949780.0
  ```
</details>


node-src/lib/turbosnap/v2/storybookVersion.ts
```diff
@@ -11,36 +12,47 @@ const STORYBOOK_CORE_PACKAGES = [
...
- * a hashable file in `preview-stats.json`. The version gate covers exactly that blind spot.
+ * This backs the `storybookVersion` manifest entry, which gates recapture on a Storybook upgrade.
+ *
+ * We resolve from the config directory, because that is the package directory that should've
+ * been used to build the Storybook. It then walks up every ancestor `node_modules` to find the
+ * first package it can resolve.
  *
- * We therefore do not reuse `ctx.storybook.version`: that reports the view-layer package. It also
- * resolves relative to the working directory without walking up to a hoisted install, and can be a
- * bare semver range rather than a concrete version.
```
justin-thurman, line 23: As far as I can tell, this reasoning for not using the fallback is still valid and isn't addressed by the fallback at all.

justin-thurman (changes requested):
Swapping the directory to the config dir looks good. But I _really_ think the fallback logic should be removed. As far as I can tell, it's entirely speculative (is that right?) and really muddles the design. There's a ton of code added here that basically just exists to plumb the fallback storybook version around, all with the most likely result that we throw it away anyway. And even if it's used, it seems likely to be wrong, based on the deleted comment (I added an inline comment pointing it out). Falling back also looks like there's a potential bug, in that the core version can move without the framework version moving alongside it.

If we really need the fallback, then I think we should rework this to make `findStorybookVersion` in `getStorybookMetadata.ts` figure out the package from `configDir` and then we just always consume `ctx.storybook.version` directly rather than resolve it here. I feel like having multiple sources of truth for storybook version is just going to bite us if we don't consolidate in one spot or another.

codykaup:
> I think we should rework this to make `findStorybookVersion` in `getStorybookMetadata.ts` figure out the package from `configDir` and then we just always consume `ctx.storybook.version` directly rather than resolve it here.

💯 — I called that out in the PR description too. I would agree with removing the fallback for now and coming back to make `ctx.storybook.version` more robust. Having multiple places that derive that value is confusing.

codykaup:
@justin-thurman that cleaned things up substantially, thanks for the suggestion!

justin-thurman (approved):
Looks good!

PR #1496, opened 2026-09-17
Add TurboSnap v2 context to Sentry and log Storybook resolution paths

codykaup:
Related to CAP-5065

Unfortunately, TurboSnap v2 is still failing to resolve a Storybook version after #1491. This PR adds the context to hopefully narrow down the root cause of the problem.

- Tags the TurboSnap v2 Sentry scope with `storybook_build_dir` so we can measure how much of the volume is prebuilt Storybooks in aggregate.
- Adds a `turbosnap_v2` Sentry context with the cwd, project root, config directory, and `storybookBuildDir` value for reading individual events.
- Logs the resolution error and every directory Node searched, marking the ones that don't exist, when a Storybook package can't be resolved. This lands in the debug log uploaded with the build so we can debug after the fact.

<details>
  <summary>📦 Published PR as canary version: <code>18.9.5--canary.1496.35258113804.0</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install chromatic@18.9.5--canary.1496.35258113804.0
  # or 
  yarn add chromatic@18.9.5--canary.1496.35258113804.0
  ```
</details>

PR #1498, opened 2026-09-17
Advise users when TurboSnap v2 runs without dependencies

codykaup:
Relates to CAP-5065

After looking at some of the metrics from #1496, it appears the directories are correct. The Storybooks themselves are leaning on TurboSnap v1 which allows them to run TurboSnap without installing their dependencies. So they build the Storybook and upload the artifact later. Since TurboSnap v2 uses source file hashes, we need the dependencies installed or we're going to miss real changes.

This adds a log line to detect this case to nudge the user in the right direction while continuing to fail the v2 run to fallback to v1.

> [!NOTE]
> This does not change Sentry behavior so we'll continue to see these pop up for a bit. We can look to filter them later!

<details>
  <summary>📦 Published PR as canary version: <code>18.9.5--canary.1498.35371619688.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.9.5--canary.1498.35371619688.0
  # or 
  yarn add chromatic@18.9.5--canary.1498.35371619688.0
  ```
</details>


node-src/lib/turbosnap/v2/index.ts
```diff
@@ -96,6 +99,54 @@ export async function traceChangedFiles(
...
+ * This is a different question from whether a package resolved. An install can exist and still
+ * lack the package we asked for, and the manifest can fail for reasons other than resolution. Only
+ * a missing install makes "install your dependencies" the right advice.
+ *
+ * @param directory The absolute directory to resolve from.
+ * @param projectFiles How to read the disk.
+ *
+ * @returns Whether node_modules exist anywhere `require` would look.
+ */
+function hasNodeModulesAbove(directory: AbsolutePath, projectFiles: ProjectFiles): boolean {
```
codykaup: I thought about moving the `node_modules` functions into their own thing and the refactor got _really_ messy. Figured the changes are fairly well isolated so I left it here.

node-src/lib/turbosnap/v2/index.ts
```diff
@@ -96,12 +111,59 @@ export async function traceChangedFiles(
...
+ * @param input.projectFiles How to read the disk.
+ */
+function adviseOnMissingInstall(
+  input: Pick<TraceChangedFilesInput, 'log' | 'turboSnapRequested' | 'configDir' | 'projectFiles'>
+) {
+  if (hasNodeModulesAbove(input.configDir, input.projectFiles)) {
+    return;
+  }
+
+  const advice = input.turboSnapRequested
```
codykaup: Commit 6d3f418014d1b80713260b8e57b00695e508c393 is to support this exact ternary. It didn't make any sense that it was using `failureLogLevel` to figure out if TurboSnap was enabled or not. I decided to pull that concept into this package instead.

node-src/lib/turbosnap/v2/index.ts
```diff
@@ -96,12 +111,59 @@ export async function traceChangedFiles(
...
+ * @param projectFiles How to read the disk.
+ *
+ * @returns Whether node_modules exist anywhere `require` would look.
+ */
+function hasNodeModulesAbove(directory: AbsolutePath, projectFiles: ProjectFiles): boolean {
+  const searchPaths = createRequire(path.join(directory, 'package.json')).resolve.paths(
+    'storybook'
+  );
+  return searchPaths?.some((searchPath) => projectFiles.isDirectory(searchPath)) ?? false;
+}
```
justin-thurman: Something about this function being here is rubbing me the wrong way. We now have a function that answers the question "did this turbosnap run fail because node modules doesn't exist, or because it exists but doesn't have the package we want?", but this function is only called in a logging helper. That just screams drift risk to me.

This also has the (admittedly rare) edge case, where any random failure in a project with no install prints this error message, which is valid, but a little odd, imo.

Would it make more sense to move this check and the advise log message into `resolveStorybookVersion`? This is the function that actually knows resolution failed. You could fold the user-facing advice into the thrown error message, and then all this stuff goes away.

justin-thurman (approved):
Approving, though I have a non-blocking suggestion that I think could clean this up a bit and prevent some drift risk.

node-src/lib/turbosnap/v2/index.ts
codykaup: I originally opt-ed for clarity for the user but I can see how it makes our side more confusing. I adjusted it so the change is simpler while still helping the customer out.

Thanks for calling this out! Sometimes you see stuff while make the change but accept the drawbacks. It's helpful when someone pushes back on that so thank you!

codykaup:
@justin-thurman enough changed that I poked you again just to be safe.

node-src/lib/turbosnap/v2/storybookVersion.ts
```diff
@@ -42,8 +42,9 @@ export function resolveStorybookVersion(
   }
 
   // Without a version there is no gate on a Storybook upgrade, so refuse to build a manifest that
-  // would silently under-capture.
+  // would silently under-capture. Not finding a Storybook dependency usually means nothing is
+  // installed so we'll inform the customer accordingly.
   throw new Error(
-    `Could not resolve a Storybook version from ${input.configDir}: none of ${STORYBOOK_CORE_PACKAGES.join(', ')} could be resolved with a version.`
+    `Could not resolve a Storybook version from ${input.configDir}: none of ${STORYBOOK_CORE_PACKAGES.join(', ')} could be resolved with a version. TurboSnap hashes installed dependencies and need to be installed before running Chromatic, even for a prebuilt Storybook.`
```
justin-thurman: This is a little confusing and kind of reads like Turbosnap needs to be installed before running Chromatic. What about this?

```suggestion
    `Could not resolve a Storybook version from ${input.configDir}: none of ${STORYBOOK_CORE_PACKAGES.join(', ')} could be resolved with a version. TurboSnap hashes installed dependencies, so they must be installed before running Chromatic, even for a prebuilt Storybook.`
```

justin-thurman (approved):
Yes, I like this much better, thank you!

PR #1499, opened 2026-09-17
Advise users when the Storybook config directory is invalid

codykaup:
Fixes CAP-5096

## Problem

TurboSnap v2's hash upload was rejected on every build for some projects:

```
The backend API rejected the hash upload: storybookConfigHashes was formatted incorrectly:
✖ Invalid input: expected string, received undefined
  → at storybookConfigFiles
```

The CLI resolves the Storybook config directory from `--storybook-config-dir`, the build script's `-c`, or `<projectRoot>/.storybook`, and never checks that it exists. TurboSnap v2 needs a valid config directory or we won't be able to hash important files which make up the entire Storybook foundation.

## Fix

TurboSnap v2 now verifies that the config directory it was given holds a `main.*` file. Storybook requires that file, so a directory without one is not a Storybook config directory. When it is missing, v2 logs a warning to the user, informing them to set `--storybook-config-dir` so TurboSnap v2 can work appropriately.

> [!NOTE]
> The rest of the CLI works just fine with an invalid config dir option so I kept this behavior isolated to where it's required (TurboSnap v2).

### Additional bug fix

If you point your static directory to the same one as the Storybook config, all of those files would be put into the `staticFiles` bucket. This change puts them in both places.

> [!IMPORTANT]
> With this change, the golden manifest test changes. This is fine since no active customers are impacted by manifest layout changes like this.

<details>
  <summary>📦 Published PR as canary version: <code>18.9.5--canary.1499.35401208677.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.9.5--canary.1499.35401208677.0
  # or 
  yarn add chromatic@18.9.5--canary.1499.35401208677.0
  ```
</details>


justin-thurman (approved):
Looks good, though I have one non-blocking thought on whether or not `storybookConfigFiles` is required. We seem to assume it is in one place and is not in another place.

The docstring for `BuildUploadHashesInput` in `uploadHashes.ts` says `The Index
 requires `storybookVersion` and `storybookConfigFiles``, but the docstring for `rollUpOutOfGraphFiles` in `outOfGraphFiles.ts` says `A section with no files contributes no entry at all, matching how the `storybookGlobals` roll-up is omitted when empty.`. I think these are two different rules that can't both hold. The roll up says we just leave out the `storybookConfigFiles` key if it's missing or every file got filtered out (in `hashOutOfGraphFiles`). But the upload hash input says that key is required. So we're building a payload we know is invalid, and we get that `expected string, received undefined at storybookConfigFiles.` error.

The guard you added fixes the case where the config dir doesn't have the `MAIN_CONFIG_PATTERN` file, which is probably the most common failure mode here, but other failure modes seem possible and would hit the same error. Example: setting `staticDirs: ['.']` in `main.ts` marks every config file as static and empties out the `storybookConfigFiles` section of the rollup.

What if, instead, we treat `storybookConfigFiles` as always present? I think you could remove it from the empty section filter, so an empty config dir sends the hash of an empty array. Then the index contract is always satisfied, and the guard you added here becomes pure advice, rather than a necessary guard.

It's possible that this breaks something else I haven't thought of and isn't a good idea. Just brainstorming.

codykaup:
> What if, instead, we treat `storybookConfigFiles` as always present? I think you could remove it from the empty section filter, so an empty config dir sends the hash of an empty array. Then the index contract is always satisfied, and the guard you added here becomes pure advice, rather than a necessary guard.

We're using the Index contract to tell us if we've built something incorrectly. We should **always** have `storybookConfigFiles` at the end because it's a representation of the required `.storybook/main.js` file and it's neighbors. If we're getting an error on that call, there's a bug inside of the CLI that should be **loud** if it's hit. Passing an invalid value to satisfy the contract would make it more difficult to debug later on, imo.

> The guard you added fixes the case where the config dir doesn't have the `MAIN_CONFIG_PATTERN` file, which is probably the most common failure mode here, but other failure modes seem possible and would hit the same error. Example: setting `staticDirs: ['.']` in `main.ts` marks every config file as static and empties out the `storybookConfigFiles` section of the rollup.

This is a very real bug so great catch! I opted to remove the static file filtering when we determine `storybookConfigFiles` so if you set a directory inside the Storybook config dir, those files will end up in both buckets. I don't know that I _love_ that solution because you can end up with files in both if you nest the static directory inside the config dirctory. Something like `staticDirs: ['./static']` would yield:

```
"storybookConfigFiles": {
  ".storybook/preview.js": "hash1",
  ".storybook/main.js": "hash2",
  ".storybook/static/image.png": "hash3"
},
"staticFiles": {
  ".storybook/preview.js": "hash1",
  ".storybook/main.js": "hash2",
  ".storybook/static/image.png": "hash3"
},
```

They both service the same purpose: If either changes, recapture everything. Personally, I think this is fine since it matches the behavior of TurboSnap v1 and is likely the safest solution. WDYT?

justin-thurman:
Yeah, that makes sense. I think double-dipping the hashes in this weird edge case is fine. Not likely to bite for normal users. Though the fact that we should always have `storybookConfigFiles` makes me wonder if we should change the hash rollup logic in some way. If an empty `storybookConfigFiles` key is invalid, I feel like the roll up logic should encode the fact that this is an invalid state. Is there a clean way to do that? Either by throwing some kind of error if we encounter that situation, or even better, typing the input so that a null/empty/undefined/whatever `storybookConfigFiles` just never reaches the rollup logic at all?

Edit: another way to put this is that I'm not at all convinced the `staticDirs: ['.']` case is the only weird edge case that could result in an empty `storybookConfigFiles` key, and it'd be nice if we could be sure, from the type system, that we just won't reach that state.

codykaup:
My hesitation with moving any of the contract logic into the CLI is duplicate ownership since we'd have to keep the CLI and the Index aligned in their schemas. I'll chew on the idea of having the type system help us out more though. That could be pretty useful as we iterate on this package. 🤔

codykaup:
Saw some funky behavior in our generated story renderer so I fixed those up while I was in there!

justin-thurman (commented):
> My hesitation with moving any of the contract logic into the CLI is duplicate ownership since we'd have to keep the CLI and the Index aligned in their schemas. I'll chew on the idea of having the type system help us out more though. That could be pretty useful as we iterate on this package. 🤔

I think we might be talking about two different contracts here. "The payload must have key X" is the index contract, but "A config dir must have a `main` file" is our contract here in the CLI. Related to the index contract, obviously, since we're building a payload for the index, but still a separate rule. We're already enforcing that contract in this PR. My suggestion is just that we move that enforcement to where that section is built, in `hashOutOfGraphFiles`, operating on the same `configPaths` array it already works with. It could throw a dedicated error type for the missing `main` file, and then`traceChangedFiles` catches that type to print the advice. That gives us one place that walks the disk, knows what shape a config dir should be, and the `storybookConfigFiles` section of the rollup now can't be empty because the code that builds it is the code that enforces, instead of enforcing separately in `index.ts` where it can drift.

If that's too big a lift right now, or if it doesn't work for some reason I'm missing, then I think at the very least the comment on `outOfGraphFiles.ts` needs to change. Right now, it says `a section with no files contributes no entry, and the Index rejects missing required fields`, which just documents the fact that we might build an invalid payload. I think, if this PR stays as-is, that comment should be rephrased to say that the config section is never empty and point to the `hasValidStorybookConfigDirectory` guard in `index.ts`. This will at least prompt anyone touching this stuff in the future to be aware of the cross-file dependency and keep them in sync.

What do you think?

codykaup:
Ooooooooohhhhhh I see what you're saying now! Yes, we should uphold our flavor of the contract. Since this isn't an urgent change (it doesn't improve the analytics for TS2.0 at all), I'd say we go ahead and make that change. I'll toss this back into draft to get that added.

codykaup:
@justin-thurman added in https://github.com/chromaui/chromatic-cli/pull/1499/commits/f68ade900a6706b7082029552d44400f8a795f91. Let me know what you think!

PR #1501, opened 2026-09-21
Follow committed symlinks when checking out baseline lockfiles

codykaup:
<details>
  <summary>📦 Published PR as canary version: <code>18.9.5--canary.1501.35632161953.0</code></summary>
  <br />

  ```bash
  npm install chromatic@18.9.5--canary.1501.35632161953.0
  # or 
  yarn add chromatic@18.9.5--canary.1501.35632161953.0
  ```
</details>


node-src/git/git.ts
```diff
@@ -466,6 +472,35 @@ export async function checkoutFile(
...
+ *
+ * @returns The repository-relative path of the regular file the symlink chain points to.
+ */
+async function resolveSymlinkAtReference(deps: GitDeps, reference: string, fileName: string) {
+  const visited = new Set<string>();
+  let current = fileName;
+
+  while (!visited.has(current)) {
+    visited.add(current);
+    const entry = await execGitCommand(deps, `git ls-tree ${reference} -- "${current}"`);
```
justin-thurman: I think this will fail from a subdirectory. `git ls-tree <ref> -- <path>` treats the path as a pathspec relative to the current working directory. `git show <ref>:<path>` treats it as relative to the repo root. `--full-tree` should fix.

```suggestion
    const entry = await execGitCommand(deps, `git ls-tree --full-tree ${reference} -- "${current}"`);
```

justin-thurman (changes requested):
Looks good! One bug with subdirectories needs a quick fix though.

node-src/git/git.ts
codykaup: I tested it against the repo root but, sure enough, testing against a subdirectory breaks the call. Great catch!

justin-thurman (approved):
👍

